PR #9, opened 2025-01-22
Differentiate docker vs local setup

### LINKED ISSUE #8: Differentiate docker vs local setup commands
The README contains instructions for local setup and running via Docker.

These instructions should be separate.

radishmouse:
Closes #8 

* adds management command for generating Fernet key
* adds management command for making an existing user a superuser
* README divides setup instructions into three sections:
  * shared/common steps
  * docker-specific steps
  * steps for running the server locally
* where applicable, README shows usage of new management commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,32 +10,53 @@ This project uses [uv](https://github.com/astral-sh/uv).
 
 Thunderbird Accounts is a Django 5.1 application. Please read up on Django before diving in.
 
-## Setup / Running
+Docker is the recommended way to run the server, though you have the option to run it locally.
 
-### Installing
 
-Ensure that uv is installed, and then run:
+## Shared setup
+
+Whether you're running the server via Docker or locally, you'll need to set up your `.env` file and copy the example Stalwart config.
+
+### Create the .env file
+
+Start from the `.env.example` file to create your dev and test `.env` files:
+
 ```shell
-uv sync
+cp .env.example .env
+cp .env.example .env.test
 ```
 
-### Using the example config for Stalwart
+In `.env.test`, set `APP_ENV=test`
+
+#### Get the `FXA_CLIENT_ID` and `FXA_SECRET`
+
+These can be found in the Services 1password vault in the item named "FXA stage application keys for TB Accounts".
+
+#### Generate `SECRET` keys
+
+Generate a different, random secrets for `SECRET_KEY` and `LOGIN_CODE_SECRET` using this command:
+
+```shell
+openssl rand -hex 16
+```
+
+### Use the example config for Stalwart
 
 Note: Do not use this for production environments.
 
 Copy config.toml.example to `mail/etc/config.toml` before first run.
 
-The default admin password should be `accounts`. 
-
-### Creating a local SuperUser
-
-Run the following and follow the terminal prompts:
-
 ```shell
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -10,46 +10,93 @@ This project uses [uv](https://github.com/astral-sh/uv).
...
+Start from the `.env.example` file to create your dev and test `.env` files:
+
 ```shell
-uv sync
+cp .env.example .env
+cp .env.example .env.test
 ```
 
-### Using the example config for Stalwart
+In `.env.test`, set `APP_ENV=test`
```
MelissaAutumn, line 29: .env.test is safe to commit to the repo as long as we don't have any live keys in it. So I can get that up later today.

README.md
```diff
@@ -10,46 +10,93 @@ This project uses [uv](https://github.com/astral-sh/uv).
...
 
-### Running local dev server
+### Generate an FXA Encryption Secret
+
+
+Generate one with this command:
 
 ```shell
-./manage.py runserver 0.0.0.0:8087
+docker compose exec backend uv run manage.py generate_key
```
MelissaAutumn, line 73: I think we can probably use this value for the other secret values too. To keep it simple.

src/thunderbird_accounts/authentication/management/commands/make_superuser.py
```diff
@@ -0,0 +1,24 @@
+from django.core.management.base import BaseCommand, CommandError
+from thunderbird_accounts.authentication.models import User
+
+
+class Command(BaseCommand):
```
MelissaAutumn, line 5: I think we can just use https://docs.djangoproject.com/en/5.1/ref/django-admin/#createsuperuser with the arguments. Unless it errors out on existing user.

MelissaAutumn (commented):
Thanks! A few comments.

src/thunderbird_accounts/authentication/management/commands/make_superuser.py
radishmouse: Just to make sure I'm clear, you're suggesting the following, yes?
* delete the custom command
* change the text to show using the `manage.py` command

README.md
radishmouse: Gotcha - I can:
* nix the `openssl` commands
* have them use `generate_key` as part of the "Shared setup"
* use that key for all three `*SECRET*` env vars

src/thunderbird_accounts/authentication/management/commands/make_superuser.py
MelissaAutumn: Yep!

README.md
MelissaAutumn: Sounds good to me!

src/thunderbird_accounts/authentication/management/commands/make_superuser.py
radishmouse: Ah - I just remembered why I created the custom command: it's if they already tried to log in via FXA but got `401 Unauthorized`. 

I already have instructions to use the `manage.py createsuperuser`, but you make a good point about adding the `--email <fxa_email_address>`

I'm going to keep the custom command and the instructions so someone can quickly fix the `401 Unauthorized` situation.

PR #11, opened 2025-01-22
Add initial vite/vuejs/services-ui integration (Fixes #10)

### LINKED ISSUE #10: Add Vite/Services-UI

MelissaAutumn:
Fixes #10 

This runs another container that runs the vite dev server. django_vite (via the tags) will insert the hmr (in dev mode), and the correct js. 

Works for me locally! Added some basic documentation, we can look into typescript a bit later just wanted to get this running. I still need to see if using createApp per SFC is a good idea, but I've used this methodology with React in a past life and it worked quite well for small reactive pages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ db.sqlite3
 
 .env
 build
+node_modules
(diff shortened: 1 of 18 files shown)


radishmouse (approved):
Looks great!

PR #13, opened 2025-01-28
Add `/waitlist` page

### LINKED ISSUE #12: Create a waiting list page for mail
Actual design pending, but we can get started. 

Edit:

On further talk with folks, creating a mailchimp in-take form. Less work, and we can import the list for the actual invites.

radishmouse:
closes #12 

* Add URL route
* Add view that renders barebones template
* Includes dummy form (that does not go anywhere yet)


MelissaAutumn:
Minor correction needed: This should be the waiting list. My bad.

src/thunderbird_accounts/urls.py
```diff
@@ -24,16 +24,14 @@
     path('auth/<str:login_code>/?redirect_to=<str:redirect_to>', auth_views.fxa_start, name='fxa_login'),
     # This will be auth/callback in the future.
     re_path(r'^favicon\.ico$', favicon_view),
-
     # Test
     path('', mail_views.home),
     path('self-serve', mail_views.self_serve),
-
+    path('waitlist', mail_views.waitlist),
```
MelissaAutumn, line 30:
```suggestion
    path('wait-list', mail_views.wait_list),
```

makes it a little easier to read

src/thunderbird_accounts/mail/views.py
```diff
@@ -5,9 +5,14 @@
...
-    return TemplateResponse(request,'mail/index.html', {})
+    return TemplateResponse(request, 'mail/index.html', {})
 
 
 def self_serve(request: HttpRequest):
-    return TemplateResponse(request,'mail/self-serve.html', {})
+    return TemplateResponse(request, 'mail/self-serve.html', {})
+
+
+def waitlist(request: HttpRequest):
```
MelissaAutumn, line 17:
```suggestion
def wait_list(request: HttpRequest):
```

src/thunderbird_accounts/mail/views.py
```diff
@@ -5,9 +5,14 @@
...
+    return TemplateResponse(request, 'mail/waitlist.html', {})
```
MelissaAutumn, line 18:
```suggestion
    return TemplateResponse(request, 'mail/wait-list.html', {})
```
(and do a rename on that template)

templates/mail/waitlist.html
```diff
@@ -0,0 +1,45 @@
...
+<h1>Join the waitlist</h1>
+<div id="mcWaitlistForm">
+  <!-- Begin Mailchimp Waitlist Form -->
+  <form
+    action="#"
+    class="waitlist-form"
+    method="post"
+    id="mc-embedded-waitlist-form"
+    name="mc-embedded-waitlist-form"
+  >
```
MelissaAutumn, line 11: I think this is suppose to have a valid action? If so can you add it to an env val, pipe it through to settings and add it as a template context? 

I do something similar with connection info at https://github.com/thunderbird/thunderbird-accounts/pull/14/files#diff-f9a2b1bfd2b55e873305640e2c285d3e3ff850edece24ab036e94001a2c61312R39

radishmouse: I just copied the `<form>` for the newsletter signup from thunderbird.net  -- we haven't created anything on the MC side, so I removed the `action`. Also, I'm unsure whether the actual form code from MC will be structured differently.

That said, I can put a value in an env val (using a dummy for now) and use that via template context

MelissaAutumn: Oh the mailchimp account should be in 1password, feel free and connect with @inattee to create a list.

.env.example
```diff
@@ -28,4 +28,7 @@ DATABASE_HOST=postgres
 REDIS_URL=redis://redis:6379
 
 # Auth scheme
-AUTH_SCHEME=fxa
\ No newline at end of file
+AUTH_SCHEME=fxa
+
+# MailChimp form URL for Wait List
+WAIT_LIST_FORM_ACTION=http://http://localhost:8087
```
MelissaAutumn, line 34: 🤔 

You can probably just leave this blank by default.

templates/mail/wait-list.html
```diff
@@ -0,0 +1,70 @@
+{% extends 'base.html' %} {% block body %}
```
MelissaAutumn, line 1:
```suggestion
{% extends 'base.html' %}
{% block body %}
```

templates/mail/wait-list.html
```diff
@@ -0,0 +1,70 @@
...
+              class="button"
+              value="Subscribe"
+            />
+          </div>
+        </div>
+      </form>
+    </div>
+  </div>
+  <!--End mc_embed_waitlist-->
+</div>
```
MelissaAutumn, line 69: The rest of this form is from their "copy this html" blerb right?

MelissaAutumn (commented):
Just a few last things sorry. 😄

templates/mail/wait-list.html
radishmouse: That's correct - it's embed code that Natalie sent over

MelissaAutumn (approved):

PR #14, opened 2025-01-29
Mail - User Management (Fixes #2)

### LINKED ISSUE #2: Add a basic account management page
For now we can have this as backend generated html via django. 
https://docs.djangoproject.com/en/5.1/topics/templates/

MelissaAutumn:
Fixes #2 

Still work-in-progress as I was wrangling multiple app password support for a bit there. Still need to add remove app password (which should be easy, but I'm out of time for today.)

Programmer art here, also there's a mix of styles from the Services-UI library (as seen on app passwords), and a lack of styles on the connection info page. That'll get corrected, but low priority as it's all temp until we get designs for the pages. 

Most of the PR size is from assets. 

### Screenshots
Screen Shot 2025-01-28 at 16 03 19-fullpage
Screen Shot 2025-01-28 at 16 03 23-fullpage
Screen Shot 2025-01-28 at 16 05 04-fullpage


.env.example
```diff
@@ -28,4 +28,16 @@ DATABASE_HOST=postgres
...
+IMAP_HOST=localhost
+IMAP_PORT=993
+IMAP_TLS=True
+
+SMTP_HOST=localhost
+SMTP_PORT=465
+SMTP_TLS=True
+
+JMAP_HOST=localhost
+JMAP_PORT=7768
```
MelissaAutumn, line 42: I just took the first number of the ord value for each letter of jmap. This is technically an http listener.

src/thunderbird_accounts/mail/views.py
```diff
@@ -1,13 +1,92 @@
...
+@require_http_methods(['POST'])
+@sensitive_post_parameters('password')
+def self_serve_app_password_add(request: HttpRequest):
+    label = request.POST['name']
+    password = make_password(request.POST['password'], hasher='argon2')
+
+    if not label or not password:
+        ValidationError('Label and password are required')
+
+    hash_algo = identify_hasher(password)
```
MelissaAutumn, line 76: I couldn't get other algos working, so I'm restricting password creation to argon2. If the hasher somehow fails, it'll error out instead of trying another hashing algo.

templates/mail/self-serve/connection-info.html
```diff
@@ -0,0 +1,38 @@
...
+        <h4 id="smtp">SMTP</h4>
+        <p><b>Server Name:</b> {{ SMTP.HOST }}</p>
+        <p><b>Server Port:</b> {{ SMTP.PORT }}</p>
+        <p><b>Security:</b> {% if SMTP.TLS %}SSL/TLS{% else %}None{% endif %}</p>
+        <p><b>Username:</b> {{ mail_username }}</p>
+        <p><b>Password:</b> Your App Password</p>
+      </div>
+    </div>
+  </div>
+{% endblock %}
```
MelissaAutumn, line 38: Thinking about splitting username/password from each block. I think this is all the information someone may need though.

MelissaAutumn:
We don't have any app password rules right now. That'll be good to get down soon.

templates/mail/self-serve/app-passwords.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'mail/self-serve/_base.html' %}
+{% block page_heading %}
+  <h2>Self Serve - App Passwords</h2>
+{% endblock %}
+{% block body %}
+  <div id="selfServeForm"></div>
+  <script>
+    window._page.appPasswords = {{ app_passwords|safe }};
```
radishmouse, line 8: alright - I think I see! this is how you're passing the labels to the vue app :+1:

src/thunderbird_accounts/mail/views.py
```diff
@@ -1,13 +1,101 @@
...
+        },
+    )
+
+
+@require_http_methods(['POST'])
+def self_serve_app_password_remove(request: HttpRequest):
+    if request.user.is_anonymous:
+        return JsonResponse({'success': False})
+    account = request.user.account_set.first()
+    account.secret = None
```
radishmouse, line 67: And they're currently limited to one account (aka, app password) per user, yes? I'm a little fuzzy on how this is set up

radishmouse (approved):
I had a couple questions/observations, but I think that it looks reasonable :+1:

PR #17, opened 2025-02-11
Add some more client tests

rwood-moz:
Add some more `create_client`, `create_client_contact`, and `create_client_environment` tests. Note that the following two tests will currently fail because there is no validation yet to prevent duplicate client contacts and duplicate client environments to be created for the same client:

`test_create_duplicate_client_environment`
`test_create_duplicate_client_contact`

### DIFF
--- a/src/thunderbird_accounts/client/tests.py
+++ b/src/thunderbird_accounts/client/tests.py
@@ -1,3 +1,5 @@
+import uuid
+
 from io import StringIO
 
 from django.core import management
@@ -10,7 +12,7 @@ fake = Faker()
 
 
 # Create your tests here.
-class CreateClientCommand(TestCase):
+class CreateClientCommands(TestCase):
     def setUp(self):
         super().setUp()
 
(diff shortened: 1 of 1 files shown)


src/thunderbird_accounts/client/tests.py
```diff
@@ -90,3 +94,142 @@ def test_create_client_when_client_already_exists(self):
 
         # Ensure that the query only equals the existing client
         self.assertQuerySetEqual(client_query, [existing_client])
+
+    def test_create_client_environment_multiple(self):
+        existing_client = models.Client.objects.create(name=self.client_name)
+        output = StringIO()
+        env_list = ['env_one', 'env_two', 'env_three', 'env_four', 'env_five']
+
+        # add eacb environment to existing client
```
MelissaAutumn, line 103:
```suggestion
        # add each environment to existing client
```

rwood-moz: Thanks Mel!

PR #19, opened 2025-02-12
Stage environment infra build

ryanjjung:
Get the accounts API backend online in AWS in some fashion.


Dockerfile
```diff
@@ -1,43 +1,40 @@
 FROM python:3.12-bookworm
 
-RUN mkdir app
+RUN mkdir -p /app/src
```
ryanjjung, line 3: In a Dockerfile, each step is a layer. Most of the changes here involve reducing the number of layers by compressing similar steps into one.

Dockerfile
```diff
@@ -1,43 +1,40 @@
...
-COPY scripts/dev-entry.sh scripts/dev-entry.sh
-
-# Dev only
-RUN echo '!! If the next step fails, copy .env.example to .env in the backend folder !!'
-
-COPY .env* .
-RUN rm .env.example
 
-# Add this hack to line it up with our dev environment.
-# I'll buy whoever fixes this a coffee.
```
ryanjjung, line 33: Did I fix it?

ryanjjung, line 33: I took a stab at fixing this, but I'm not sure of the original intention. This ends up with a working image, and dev environments shouldn't be affected. The docker-compose.yml is still set up to mount in the code to /app, but I think that's not really necessary as long as you're using the `--build` flag. Still, I'm not sure I've wrapped my head around what this was doing before I went and mucked with it.

docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
     volumes:
       - ./.env:/app/.env
       - ./vite.config.mjs:/app/vite.config.mjs
-      - ./src/thunderbird_accounts:/app/thunderbird_accounts
+      - ./src/thunderbird_accounts:/app/src/thunderbird_accounts
```
MelissaAutumn, line 28: Thanks! I'm not sure why i'm voluming mount at /app/thunderbird_accounts and not just /app/src/thunderbird_accounts. I'll look at this locally, but I may owe you a coffee 😄

ryanjjung:
Think this one is ready for review. This will be iterative, of course, but this gets us to some basic infra with a stable container.

src/thunderbird_accounts/infra/views.py
```diff
@@ -0,0 +1,7 @@
+from django.http import HttpRequest, HttpResponse
+
+
+def health_check(request: HttpRequest):
+    """Do nothing, only prove the service is responsive."""
+
+    return HttpResponse('Aliveness demonstrated!')
```
MelissaAutumn, line 7: On appointment we demonstrate connectivity and aliveness. Did we want to do a db query here?

MelissaAutumn: (Although I think the middleware technically does touch the db, so I think we may be fine.)

MelissaAutumn (approved):
Looks good, thanks!

PR #28, opened 2025-02-26
Fix broken local tests

rwood-moz:
This [previous PR](https://github.com/thunderbird/thunderbird-accounts/pull/19/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557) made some Dockerfile changes which inadvertently broke running the tests against your local dev env. The `docker compose exec backend uv run python manage.py test` cmd fails with settings.SECRET_KEY is not set error.

Add back some of the Dockerfile commands to fix this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,10 @@ COPY scripts/dev-entry.sh scripts/dev-entry.sh
 COPY static ./static/
 COPY templates ./templates/
 
+# Required in order to run tests against local dev
+COPY .env.test .
+RUN ln -s /app/thunderbird_accounts src/thunderbird_accounts
+
 # Add our source code
 ADD src/thunderbird_accounts ./src/thunderbird_accounts
 


rwood-moz:
Now with this PR the tests run locally again:

```
rwood@Roberts-MacBook-Pro thunderbird-accounts % docker compose exec backend uv run python manage.py test
[DEBUG] 2025-02-26 17:45:00,204: Not in REPL -> leaving logger event level as is.
..
----------------------------------------------------------------------
Ran 23 tests in 0.036s

```

MelissaAutumn (approved):
Thanks! I think we should explicitly copy .env.test and .env though. 

This prevents a case where someone might have a .env.something that contains testing secrets. While not a problem in itself, if that container were to accidentally be uploaded somewhere or pulled it would contain that .env file.

rwood-moz:
> > Thanks! I think we should explicitly copy .env.test and .env though.
> > This prevents a case where someone might have a .env.something that contains testing secrets. While not a problem in itself, if that container were to accidentally be uploaded somewhere or pulled it would contain that .env file.
> 
> Good call, updated, thanks

Turns out we only need to cp `.env.test` anyway, updating again haha

MelissaAutumn:
we'll need .env for normal dev work as well. Can just push that up on main?

rwood-moz:
> we'll need .env for normal dev work as well. Can just push that up on main?

Ahhhh true, it's ok I'll make another PR but use the same issue. Thanks!

PR #31, opened 2025-02-27
Bring managed Redis online; fix health checks

ryanjjung:
**DO NOT MERGE,** as this PR depends on the release of a new tb_pulumi version.

This PR accomplishes a bunch of things:

1. It brings Redis infrastructure online for the stage environment using a new tb_pulumi module.
2. It updates the task definition to use a temporary image containing the code changes in this PR.
3. It updates our authentication middleware to allow for requests aimed at the container's IP address, such as the health check requests that come from the ALB in front of these containers.
4. It builds a DNS record in CloudFlare to route traffic to the service.

Some more commentary inline...


.env.test
```diff
@@ -20,10 +23,15 @@ FXA_ALLOW_LIST=
...
+DATABASE_NAME=accounts
 DATABASE_USER=accounts
 DATABASE_PASSWORD=accounts
-DATABASE_HOST=sqlite
+DATABASE_HOST=postgres
+
+# Postgres image (https://hub.docker.com/_/postgres/)
+POSTGRES_USER=accounts
+POSTGRES_DB=accounts
+POSTGRES_PASSWORD=accounts
```
ryanjjung, line 34: These `POSTGRES_*` changes ensure that when the Postgres image spins up, it changes these values to what the client wants them to be (set with the `DATABASE_*` variables). This solved a local dev problem for me where the service could not talk to the local DB instance.

docker-compose.yml
```diff
@@ -29,6 +29,8 @@ services:
       dockerfile: ./Dockerfile
     ports:
       - "8087:8087"
+    environment:
+      <<: *env
```
ryanjjung, line 33: The backend should read in the .env file as well so we can manipulate Django options through it in local dev.

pulumi/config.stage.yaml
```diff
@@ -104,7 +127,7 @@ resources:
           - FARGATE
         container_definitions:
           accounts:
-            image: 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:jungtest9
+            image: 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:jungtest15
```
ryanjjung, line 130: Temporary image containing the code proposed in this PR. This is what is currently live at https://accounts-stage.tb.pro/

pulumi/requirements.txt
```diff
@@ -1 +1,2 @@
-tb_pulumi @ git+https://github.com/thunderbird/pulumi.git@accounts
+tb_pulumi @ git+https://github.com/thunderbird/pulumi.git@v0.0.12
```
ryanjjung, line 1: This is the as yet unreleased version of tb_pulumi.

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -53,6 +54,9 @@ def __call__(self, request: HttpRequest):
         allowed_hosts = cache.get(settings.ALLOWED_HOSTS_CACHE_KEY)
         if not allowed_hosts:
             allowed_hosts = ClientEnvironment.cache_hostnames()
+        allowed_hosts.append(gethostbyname(gethostname()))
```
ryanjjung, line 57: This gets the container's current IP address and adds it as an allowed host.

.env.test
MelissaAutumn: This is the test env file, which runs sqlite so tests can be run outside of docker. We can change this to postgres, but did you mean to make these changes in .env instead?

pulumi/config.stage.yaml
```diff
@@ -134,27 +157,28 @@ resources:
                 valueFrom: arn:aws:secretsmanager:eu-central-1:768512802988:secret:accounts/stage/secret-key-ZBBV7V
             environment:
               - name: ADMIN_WEBSITE
-                value: https://www.thunderbirid.net
+                value: https://www.thunderbird.net
```
MelissaAutumn, line 160: image

Hmmmmmm

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: Can you throw a comment on that line?

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -53,6 +54,9 @@ def __call__(self, request: HttpRequest):
...
+        if settings.DEBUG:
+            allowed_hosts.append('localhost')
```
MelissaAutumn, line 59: I feel like this conditional should be in settings.py since it's fairly static.

MelissaAutumn (changes requested):
Thanks! Just a few things.

.env.test
ryanjjung: I don't think there is an `.env` file here. There's an "example" and a "test" env file. Seems the docker-compose reads them all in (`<<: *env`), so they apply. These changes solved connectivity problems in local dev for me, but maybe they should be moved to the .example?

ryanjjung: On my machine, I have `.env` as a symlink to `.env.test`.

pulumi/config.stage.yaml
ryanjjung: There's literally no reason not to :)

src/thunderbird_accounts/authentication/middleware.py
ryanjjung: 👍🏻

src/thunderbird_accounts/authentication/middleware.py
ryanjjung: I think it is, actually. I'l tidy this up. This was probably leftovers from some testing.

.env.test
MelissaAutumn: Yeah that's not correct, you should copy over .env.example as .env. It's for personal use and isn't tracked by git so you can put all of the secrets you could ever hope for in there without the risk of leaking them. (well via git anyways.)

 Apologies for the confusion, I'll update the documentation!

ryanjjung:
I think I've resolved the issues with the test environment setup. Requesting re-review.

.env.example
```diff
@@ -21,12 +24,17 @@ FXA_CALLBACK=http://localhost:8087/api/v1/auth/fxa/callback
...
+# Local database container setup
 DATABASE_NAME=accounts
 DATABASE_USER=accounts
 DATABASE_PASSWORD=accounts
 DATABASE_HOST=postgres
 
+# Point Django at the local database container (image docs: https://hub.docker.com/_/postgres/)
+POSTGRES_USER=accounts
+POSTGRES_DB=accounts
+POSTGRES_PASSWORD=accounts
```
MelissaAutumn, line 36: This should be covered by the env section in docker-compose, you can remove it here.

MelissaAutumn (approved):
One thing and you're good to go!

.env.example
ryanjjung: They're not currently covered by the docker-compose, to my knowledge, because the docker-compose just [volume-mounts the env file](https://github.com/thunderbird/thunderbird-accounts/blob/main/docker-compose.yml#L26-L40) and doesn't specify any other variables.

ryanjjung: Sorry for the flurry of commentary/deletions there, I realized I was wrong about something and went to take a better look. I think there are a number of small issues here that I'm going to tidy up with one more commit.

This file contains variables for many different services, which I have rather poorly documented in the comments. In docker-compose, the postgres image takes its environment from the env files with a `<<: *env` line. The Django services (backend/frontend) get the `.env` file volume-mounted to `/app/.env`, which would cause an app-level override of anything we set in the `environment` section of the service. That's fine, especially as this .env file is mostly targeted at the Django application.

My proposal: Stop reading in the env files for the postgres service, opting to specify these values directly in docker-compose (the only context in which these settings matter, I think?). Keep the volume mapping for the Django application since that's the most direct way of influencing the local services.

ryanjjung:
I think that takes care of that last round of issues. One more look, please?

MelissaAutumn (approved):
Thanks for the changes!

PR #33, opened 2025-02-28
Fix serving assets on dev, and split dev and stage docker/entry files (Fixes #32)

### LINKED ISSUE #32: Fix dev not serving assets as static assets

MelissaAutumn:
When I added ServeStatic I accidentally broke instant reloading with dev. This PR fixes it, and because dev doesn't need collectstatic I removed that line from dev-entry.

...and since stage seems to use dev-entry, I've mirrored how it was working before in entry.sh and dockerfile.stage. 
(cc @ryanjjung to be aware of changes.)

### DIFF
--- a/src/thunderbird_accounts/asgi.py
+++ b/src/thunderbird_accounts/asgi.py
@@ -24,6 +24,7 @@ def immutable_file_test(path, url):
 
 
 application = get_asgi_application()
-application = ServeStaticASGI(
-    application, root=settings.STATIC_ROOT, prefix='static', immutable_file_test=immutable_file_test
-)
+if not settings.DEBUG:
+    application = ServeStaticASGI(
+        application, root=settings.STATIC_ROOT, prefix='static', immutable_file_test=immutable_file_test
+    )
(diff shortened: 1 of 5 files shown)


src/thunderbird_accounts/urls.py
```diff
@@ -58,4 +58,4 @@
 
 # Needed with uvicorn dev server
 if settings.DEBUG:
-    urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
+    urlpatterns += static(settings.STATIC_URL, document_root=settings.ASSETS_ROOT)
```
MelissaAutumn, line 61: For local dev we serve the assets directory at /static/

src/thunderbird_accounts/asgi.py
```diff
@@ -24,6 +24,7 @@ def immutable_file_test(path, url):
 
 
 application = get_asgi_application()
-application = ServeStaticASGI(
-    application, root=settings.STATIC_ROOT, prefix='static', immutable_file_test=immutable_file_test
-)
+if not settings.DEBUG:
```
MelissaAutumn, line 27: For non-local dev envs we let staticserve serve the files that were collectstatic'd on.

radishmouse (changes requested):
It looks good, but when I run it locally editing a css file does not change the css files in the container.

(I've pruned all my volumes, images, etc. and started the containers from scratch)

MelissaAutumn:
I've pushed a hotfix up and that should fix the issue.

PR #56, opened 2025-03-24
Add is in allow list endpoint (Fixes #55)

### LINKED ISSUE #55: Authentication Support
Changes related to connecting one of our services to TB Account for session auth.

MelissaAutumn:
Issue #55 

Like Appointment we first check to see if the user exists, otherwise we check the allow list. We add some caching into the mix (1 day by default), and clear if a user is deleted, or if they change their email.

This is a single allow list which will cover all services right now. We can add additional checks as we move services into and out of public/closed access.

Additionally this adjusts the create client management command to make on-boarding a service easier for dev purposes. (You'll see in the appointment-side PR soon lol.)

For existing services that have their own allow list we'll need to set our client environment as public, and handle it on the service-side.


MelissaAutumn:
Thanks all fixed now. I need to get the CI tests online.

src/thunderbird_accounts/client/management/commands/create_client_contact.py
```diff
@@ -26,17 +34,22 @@ def add_arguments(self, parser):
...
+        verbosity = options.get('verbosity', 1)
 
         try:
             client = models.Client.objects.get(uuid=client_uuid)
         except models.Client.DoesNotExist:
-            self.stdout.write(self.style.ERROR(f'Client: {client_uuid} does not exist'))
-            return
+            if verbosity > 0:
+                self.stdout.write(self.style.ERROR(f'Client: {client_uuid} does not exist'))
+            return self.ReturnCodes.CLIENT_DOESNT_EXIST.value
```
rwood-moz, line 44: Returning a value here is breaking some tests in `src/thunderbird_accounts/client/tests.py`. I looked into it more and the tests are capturing stdout and checking it for expected errors output by the client command; however the django stdout cannot handle this integer return value (expects a string only). Here's an example:

```
======================================================================
ERROR: test_create_client_contact_when_client_not_exist (thunderbird_accounts.client.tests.CreateClientCommands.test_create_client_contact_when_client_not_exist)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/thunderbird_accounts/client/tests.py", line 199, in test_create_client_contact_when_client_not_exist
    management.call_command(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        'create_client_contact', fake_client_uuid, self.contact_name, self.contact_email, self.contact_url, stdout=output
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/.venv/lib/python3.13/site-packages/django/core/management/__init__.py", line 194, in call_command
    return command.execute(*args, **defaults)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/django/core/management/base.py", line 468, in execute
    self.stdout.write(output)
    ~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/django/core/management/base.py", line 178, in write
    if ending and not msg.endswith(ending):
                      ^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'endswith'
```
Same goes for the other places with `return self.ReturnCodes...`. Not sure how to fix that except by returning a string instead of an enum value.

rwood-moz (changes requested):

src/thunderbird_accounts/client/management/commands/create_client_contact.py
MelissaAutumn: Hmm its mostly to simplify command output for first time setup on appointment.

I will have to think about this. hmm

src/thunderbird_accounts/authentication/signals.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.db.models.signals import post_delete, pre_save
+from django.dispatch.dispatcher import receiver
+
+from thunderbird_accounts.authentication.models import User
+from thunderbird_accounts.authentication.utils import delete_cache_allow_list_entry
+
+
+@receiver(post_delete, sender=User)
+def user_post_delete(sender, instance, **kwargs):
```
rwood-moz, line 11: I don't see `sender` being used

src/thunderbird_accounts/authentication/tests.py
```diff
@@ -334,3 +337,140 @@ def test_fxa_process_delete_user(self):
...
+
+        # Okay the allow list should return false now
+        response = self.api_client.post(
+            'http://testserver/api/v1/auth/is-in-allow-list/',
+            data={'email': original_email, 'secret': self.client_env.auth_token},
+        )
+        self.assertEqual(response.status_code, 200, response.content)
+        result = response.json().get('result')
+        self.assertIsNotNone(result)
+        self.assertFalse(result)
```
rwood-moz, line 476: These tests are awesome! Suggestion for a couple more:
- test entry when no allow list exists (expect True)
- test entry but don't provide email in the request (expect False)

src/thunderbird_accounts/authentication/tests.py
```diff
@@ -334,3 +337,140 @@ def test_fxa_process_delete_user(self):
...
+        )
+        self.assertEqual(response.status_code, 200, response.content)
+        result = response.json().get('result')
+
+        self.assertIsNotNone(result)
+        self.assertTrue(result)
+
+        self.assertTrue(get_cache_allow_list_entry(self.user.email))
+
+        # okay lets delete the user, this should remove the cache...
```
rwood-moz, line 461: nit, comment should be something like `okay lets change the user's email`

aaspinwall (approved):
Looking good! 
I added a non-blocker comment.

rwood-moz (changes requested):
Very cool! Just a couple of small suggestions

src/thunderbird_accounts/authentication/signals.py
MelissaAutumn: Looks like it's working without sender. So I'll pop that off.

src/thunderbird_accounts/authentication/tests.py
MelissaAutumn: Copy-pasta error, thanks for the catch!

src/thunderbird_accounts/authentication/tests.py
```diff
@@ -474,3 +474,61 @@ def test_entry_with_updated_email_is_uncached(self):
...
+
+        response = self.api_client.post(
+            'http://testserver/api/v1/auth/is-in-allow-list/',
+            data={},
+        )
+        result = response.json()
+
+        self.assertEqual(response.status_code, 401)
+        self.assertIn('detail', result)
+        self.assertEqual(result['detail'], 'Authentication credentials were not provided.')
```
rwood-moz, line 534: Awesome (and some other great additions) thanks! All tests passed for me locally.

rwood-moz (approved):
Great!

PR #65, opened 2025-03-26
Use the enums instead of int literals for "are we done" check on the client creation command

MelissaAutumn:
Should fix the issue where the client id and secret don't show up in the results.

You'll still get some OKs though 😅


rwood-moz (approved):
Thanks! I pulled locally and confirmed it fixes the issue with the create_client command. Much appreciated!

PR #68, opened 2025-03-31
Update TB Pro Sign up with new form id, the slim appeal footer, and updated styling.

MelissaAutumn:
Screen Shot 2025-03-31 at 12 12 30-fullpage

I've checked in the edits that were missing from this repo from https://thundermail.com/. 

The form id is actually an env file so it's not visible here. I've also added deployment instructions.

### DIFF
--- a/docs/deploying-thundermail-waiting-list.rst
+++ b/docs/deploying-thundermail-waiting-list.rst
@@ -1,0 +1,24 @@
+==============================================
+Deploying Thundermail.com Waiting List
+==============================================
+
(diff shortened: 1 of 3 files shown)


docs/deploying-thundermail-waiting-list.rst
```diff
@@ -0,0 +1,24 @@
+==============================================
+Deploying Thundermail.com Waiting List
```
malini, line 2: thank you!

Issue #71, opened 2025-04-02
Implement Celery


MelissaAutumn:
We need:
- [x] Add celery
- [x] Add tests to ensure celery is working correctly
- [x] Docker container to run celery worker
- [x] Add extra docker container to deploy stage CI runner

Issue #51, opened 2025-03-18
Implement Client Webhooks


MelissaAutumn:
For services that need to store user data we'll need to pass down a webhook for delete and email change requests. Logouts can be triggered simply by deleting the cache session. 

We'll need some sort of queue to store requests to send until they receive a 200 from the client.

PR #73, opened 2025-04-02
Add celery + docker config and a stub task for client webhooks.

MelissaAutumn:
Part of #71 

Will add it to the ci runner once that is working.

This PR adds [celery](https://docs.celeryq.dev/en/stable/)! Basically it's a task broker that throws stuff into queue and runs it async. It can also store results if needed. I haven't tested the stage dockerfile, will add onto in a future PR.

There's also no tests currently, because mock patching doesn't seem to work nicely with Django Rest Framework. I'm tempted to add in pytest so we can actually patch stuff behind requests. My other alternative is just a standard function unit test. I'll expand on that for the #51 which this is a pre-req for.


docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - "6380:6379" # Redis exposed on port 6380
       - "8071:8001" # Insights exposed on port 8071
 
-  backend:
+  backend: &backend
```
radishmouse, line 28: TIL about anchors in YAML

radishmouse (approved):

PR #77, opened 2025-04-02
Add the celery worker to pulumi and the github deploy workflow

MelissaAutumn:
Part of #71 

This hopefully spins up a second task that just runs the celery worker. (Concurrency defaults to the number of cpus available, see https://github.com/thunderbird/thunderbird-accounts/blob/main/scripts/celery-entry.sh for more info.) We don't need any http listener, but I'm unsure how the health check will run otherwise. 🤔 

In the future we may need a second celery container task def to run a cron scheduler. So taking suggestions on optimizing this workflow/pulumi config.
https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#starting-the-scheduler

### DIFF
--- a/pulumi/config.stage.yaml
+++ b/pulumi/config.stage.yaml
@@ -83,7 +83,7 @@ resources:
         - paddle-price-id-lo
         - paddle-price-id-md
         - paddle-price-id-hi
-  
+
   # Config to build a jumphost
   # tb:ec2:SshableInstance:
   #   jumphost:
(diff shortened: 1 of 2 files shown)


pulumi/config.stage.yaml
```diff
@@ -193,6 +196,23 @@ resources:
...
+                value: '10'
+              - name: CELERY_BROKER
+                value: rediss://master.accounts-stage-redis.7syyoy.euc1.cache.amazonaws.com:6379
+              - name: CELERY_BACKEND
+                value: rediss://master.accounts-stage-redis.7syyoy.euc1.cache.amazonaws.com:6379
+              - name: CELERY_EAGER
+                value: 'False'
+          celery_worker:
+            <<: *accounts_def
+            portMappings: null
```
MelissaAutumn, line 215: Image is shoved in during the github workflow.

ryanjjung (approved):

PR #82, opened 2025-04-07
Add unit test attempting to sign up with non-unique email address

### LINKED ISSUE #83: User is able to sign up for an email address that is already taken by another user
Able to sign up for an email address that is already taken by a different TB accounts user. If one user signs up for an email address, then another user signs up for the same email address, the `sign_up_submit` view doesn't return an error and proceeds to the `connection_info` view.

rwood-moz:
Add a unit test that attempts to sign up for email using an email address that is already taken by another user. Also add an update to the `sign_up_submit` view to fix #83 as there was no check to ensure the requested email wasn't already taken.

### DIFF
--- a/src/thunderbird_accounts/mail/tests.py
+++ b/src/thunderbird_accounts/mail/tests.py
@@ -8,20 +8,21 @@ class TestMail(TestCase):
     def setUp(self):
       cache.clear()
       self.c = Client(enforce_csrf_checks=False)
-      UserModel = get_user_model()
-      self.user = UserModel.objects.create(fxa_id='abc123', email='user@test.com')
-
-    def tearDown(self):
-      cache.clear()
-  
-    def test_sign_up_successful(self):
-      self.c.force_login(self.user)
-
-      resp = self.c.post(reverse('sign_up_submit'), data={
+      self.UserModel = get_user_model()
+      self.user = self.UserModel.objects.create(fxa_id='abc123', email='user@test.com')
+      self.sign_up_data = {
             'email_address': 'new-thundermail-address',
             'email_domain': 'tb.pro',
             'app_password': 'password',
-        })
+        }
+
+    def tearDown(self):
+      cache.clear()
+
+    def test_sign_up_successful(self):
+      self.c.force_login(self.user)
+
+      resp = self.c.post(reverse('sign_up_submit'), data=self.sign_up_data)
 
       # expect redirect to connection-info after email created successfully
       self.assertEqual(resp.status_code, 302)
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/mail/tests.py
```diff
@@ -82,29 +71,47 @@ def test_sign_up_invalid_domain(self):
...
+        username='Second User'
+      )
+      second_client.force_login(self.second_user)
+
+      # then have the 2nd user attempt to sign up for email using the same email address
+      resp = second_client.post(reverse('sign_up_submit'), data=self.sign_up_data)
+
+      # expect redirect back to /sign-up since sign up should fail b/c requested email address is not unique
+      self.assertEqual(resp.status_code, 302)
+      self.assertEqual(resp.url, '/sign-up/')
```
MelissaAutumn, line 117: Can you add some assertions to ensure the email account, and address aren't created in the db side too?

MelissaAutumn (commented):
Some suggestions, looks good otherwise!

src/thunderbird_accounts/mail/tests.py
rwood-moz: Oh great idea, will do!

src/thunderbird_accounts/mail/views.py
```diff
@@ -62,6 +62,12 @@ def sign_up_submit(request: HttpRequest):
     )
     account.save_app_password('Mail Clients', request.POST['app_password'])
 
+    try:
+        Email.objects.get(address=email_address)
+        return raise_form_error(request, reverse('sign_up'),  _('Requested email is already taken'))
+    except Email.DoesNotExist:
+        pass
```
MelissaAutumn, line 69: I think I forgot to add a comment, but can you move this up with the other validation? So we're not creating anything if we don't have a unique email.

src/thunderbird_accounts/mail/views.py
```diff
@@ -53,6 +53,12 @@ def sign_up_submit(request: HttpRequest):
 
     email_address = f'{request.POST['email_address']}@{request.POST['email_domain']}'
```
MelissaAutumn, line 54: Hmm I guess we may want to make a ticket to further sanitize this email address input. (No @'s, no spaces, etc..)

MelissaAutumn (approved):
Thank you!!

src/thunderbird_accounts/mail/views.py
rwood-moz: Good idea, I created issue #85.

Thanks for the review!

Issue #53, opened 2025-03-20
Change email domains


MelissaAutumn:
We should only display thundermail.com right now.

- [x] Update stage (stage-thundermail.com)
- [x] Update prod (thundermail.com)

MelissaAutumn:
Awaiting stage's domain

PR #94, opened 2025-04-09
Tie email domains to env, and update stage's variable. (#53)

MelissaAutumn:
Part of #53 

Ties email domains to env variables so we can vary them by environment...

image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,9 @@ APP_ENV=dev
 APP_DEBUG=True
 PUBLIC_BASE_URL=http://localhost:8087
 
+# Comma separated list of tlds (e.g. ALLOWED_EMAIL_DOMAINS=example.com,example.org)
+ALLOWED_EMAIL_DOMAINS=example.org
+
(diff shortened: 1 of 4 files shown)


test/e2e/const/constants.ts
```diff
@@ -29,7 +29,7 @@ export const APP_PASSWORD_NONE = 'Your App Password';
 
 // email sign-up
 export const EMAIL_SIGN_UP_EMAIL_ADDRESS = 'my-new-email-address';
-export const EMAIL_SIGN_UP_DOMAIN = 'tb.pro';
+export const EMAIL_SIGN_UP_DOMAIN = 'example.org';
```
rwood-moz, line 32: Awesome thank you 😃

rwood-moz (approved):
LGTM. Pulled the branch and ran locally and worked great, was able to add some more ALLOWED_EMAIL_DOMAINS and use them in the email sign-up. 🎉

PR #102, opened 2025-04-16
Display the user's primary email address, and fix smtp settings. (Fixes #98)

### LINKED ISSUE #98: Display the user's thundermail email address
Under connection info we show them the various information they need to login to an external email client...except the actual thundermail address.

MelissaAutumn:
Simply adds their email address above connection info. Since their login username is their fxa email I realized we didn't show this anywhere...

This also fixes the smtp connection info showing jmap connection info


templates/mail/self-serve/connection-info.html
```diff
@@ -4,6 +4,7 @@ <h2>Self Serve - Connection Information</h2>
 {% endblock %}
 {% block body %}
   <div class="connection-info-container">
+    <p data-testid="connection-info-primary-email-address"><b>Your email address:</b> {{ mail_address }}</p>
```
rwood-moz, line 7: Thanks for that! I'll add this field to the E2E tests.

rwood-moz (changes requested):
If you could please update this value in the E2E tests constants with the corrected SMTP port (456) and that will fix the broken E2E test, thanks!

https://github.com/thunderbird/thunderbird-accounts/blob/main/test/e2e/const/constants.ts#L25

rwood-moz:
I also notice there is a space in the email address after the `@` sign however in this case I'm guessing it's like that in the DB?

Screenshot 2025-04-16 at 1 25 53 PM

MelissaAutumn:
> I also notice there is a space in the email address after the `@` sign however in this case I'm guessing it's like that in the DB?
> Screenshot 2025-04-16 at 1 25 53 PM

Seems like you have a space in the env? I've corrected this in the next commit but you'll have to fix this in your db's mail_email table iirc.

rwood-moz:
> Seems like you have a space in the env? I've corrected this in the next commit but you'll have to fix this in your db's mail_email table iirc.

Ahh yes sorry about that, I had spaces after the commas, my bad

`ALLOWED_EMAIL_DOMAINS=example.org, rwood.ca, coffee.com, latte.ca, mocha.ca, somebiglongdomain.ca, anotherone.com`

rwood-moz (approved):

MelissaAutumn:
> > Seems like you have a space in the env? I've corrected this in the next commit but you'll have to fix this in your db's mail_email table iirc.
> 
> Ahh yes sorry about that, I had spaces after the commas, my bad
> 
> `ALLOWED_EMAIL_DOMAINS=example.org, rwood.ca, coffee.com, latte.ca, mocha.ca, somebiglongdomain.ca, anotherone.com`

No worries, thanks for finding that bug! Love the domains.

Issue #105, opened 2025-04-17
E2E test failure when run on stage


rwood-moz:
The E2E tests are failing when running on stage after stage deployments. The tests are failing to run for some reason, here's an example [github action link](https://github.com/thunderbird/thunderbird-accounts/actions/runs/14516958388/job/40728419013#annotation:6:41), where the E2E tests fail with:

```
 > npx browserstack-node-sdk playwright test --grep e2e-suite

2025-04-17 13:39:54 - info: Running SDK with args /opt/hostedtoolcache/node/20.19.0/x64/bin/node /home/runner/work/thunderbird-accounts/thunderbird-accounts/test/e2e/node_modules/.bin/browserstack-node-sdk playwright test --grep e2e-suite
2025-04-17 13:39:54 - info: Reading configs from /home/runner/work/thunderbird-accounts/thunderbird-accounts/test/e2e/browserstack.yml
2025-04-17 13:39:54 - info: SDK run started with id: d22937ca-746d-4a77-952c-75d4daacf924
2025-04-17 13:39:54 - info: Skipping buildIdentifier as BROWSERSTACK_BUILD_NAME is set
2025-04-17 13:39:54 - info: Project and debug logs are captured by default. To disable, set disableAutoCaptureLogs: true in config.
2025-04-17 13:39:55 - info: Testhub started with id: 2sxkkomh1mpp1pkw3mczptukxg9bqrjlvtmpnnho

Running 8 tests using 2 workers

Timed out waiting 300s for the test suite to run

Timed out waiting 300s for the teardown for test suite to run

  8 did not run
```

When I issue the same command to run the tests on BrowserStack but from my local machine they do run fine, so need to investigate what is happening in CI.

Note: The stage deploy step is passing fine so this issue is not stopping actual stage deployments.

PR #106, opened 2025-04-17
Fix stage E2E test BrowserStack job

rwood-moz:
The E2E tests BrowserStack job on stage is failing with:

```
Running 8 tests using 2 workers

Timed out waiting 300s for the test suite to run
```

See #105  for more info.

The only thing I can see is that the playwright browsers are being installed, but we don't need them as the tests are running in BrowserStack. Not sure if that is the exact issue but perhaps the installation of the playwright browsers is causing the tests to fail to connnect / start in the BrowserStack browsers. The playwright browsers aren't required either way so good to remove and hopefully this will fix the issue. 🤞

### DIFF
--- a/.github/workflows/deploy-stage.yml
+++ b/.github/workflows/deploy-stage.yml
@@ -170,7 +170,6 @@ jobs:
         run: |
           cd ./test/e2e
           npm install
-          npx playwright install
 
       - name: BrowserStack Env Setup
         uses: browserstack/github-actions/setup-env@master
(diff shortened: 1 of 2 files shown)


MelissaAutumn (approved):
Thanks!

rwood-moz:
Yayy that did fix the issue, tests running fine now in BrowserStack after stage deployments 🥳 

https://github.com/thunderbird/thunderbird-accounts/actions/runs/14520500594/job/40740213235

PR #110, opened 2025-04-25
Celery containers

MelissaAutumn:
The only addition I added was moving common keys to a single key and then we merge it where we need it. I also override allowed_hosts for celery to null it out.

I haven't ran it yet, and this is too big for a PR so included is the out.log from the diff. 

[out.log](https://github.com/user-attachments/files/19915311/out.log)


ryanjjung (approved):

MelissaAutumn:
There's some problems with the celery worker not looking at redis which i still need to sort out.

PR #122, opened 2025-04-30
Use login_required decorator instead of duplicate code. (Fixes #121)

### LINKED ISSUE #121: Re-work authentication re-direct
Currently we have to put
```
    if request.user.is_anonymous:
        return HttpResponseRedirect(
            reverse('fxa_login', kwargs={'login_code': get_admin_login_code(), 'redirect_to': quote_plus(request.path)})
        )
```

in every view. That's not ideal.

MelissaAutumn:
Fixes #121 

We now use the [login_required decorator](https://docs.djangoproject.com/en/5.2/topics/auth/default/#the-login-required-decorator). This adds a generic /login route which that decorator is pointed to. I basically just moved the copy-pasted code to there, and set a default redirect to if someone goes to /login directly.

I've added some auth tests to help ensure it doesn't break.


src/thunderbird_accounts/authentication/tests.py
```diff
@@ -545,3 +549,40 @@ def test_with_no_secret_or_email(self):
...
+        client = RequestClient()
+        login_url = reverse(settings.LOGIN_URL)
+
+        for key, _ in self.login_required_keys:
+            url = reverse(key)
+            self.assertTrue(url)
+
+            response = client.get(url, follow=False)
+            self.assertEqual(response.status_code, 302)
+            self.assertTrue(response.url.startswith(login_url))
```
MelissaAutumn, line 575: This returns a redirect response, which has url

src/thunderbird_accounts/authentication/tests.py
```diff
@@ -545,3 +549,40 @@ def test_with_no_secret_or_email(self):
...
+        user = User.objects.get_or_create(username='test', fxa_id='1234')[0]
+        client.force_login(user)
+
+        for key, status in self.login_required_keys:
+            url = reverse(key)
+            self.assertTrue(url)
+
+            response = client.get(url, follow=False)
+            self.assertEqual(response.status_code, status)
+            self.assertTrue(response.request['PATH_INFO'].startswith(url))
```
MelissaAutumn, line 588: This usually returns a templateresponse, but varies based on view. We retrieve the request data to ensure it's not redirecting to login.

MelissaAutumn:
Okie tests are fixed.

PR #124, opened 2025-05-01
Add an ENV for containers and use a different base path if within a container vs outside a container. (Fixes #123)

MelissaAutumn:
This removes the weird path logic I came up with that attempted to solve both issues but didn't really.

Basically I run tests outside the container because it's easier to do. However the CI runs tests within a container, and honestly it shouldn't matter where we run tests. It should just work(tm).

So this should solve it, we build the container and set a ENV that only exists if you're in a container.


.github/workflows/validate.yml
```diff
@@ -23,9 +23,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Apply env
-        run: cp .env.test.docker .env.test
-
```
MelissaAutumn, line 28: This is a work-around I added right before I merged the last PR to fix tests.

MelissaAutumn:
The work around actually breaks local dev, so merging this ahead of schedule.

PR #128, opened 2025-05-01
Fix E2E tests on stage

rwood-moz:
Fix the broken E2E tests on stage:

- Include the `EMAIL_SIGN_UP_ADDRESS` in the GHA workflow env
- Update the incorrect `ACCTS_HOST` value in the `test/e2e/.env.stage.example`


MelissaAutumn (approved):
LGTM!

PR #138, opened 2025-05-13
Paddle model + webhook changes (Fixes #93)

### LINKED ISSUE #93: Update paddle models with recommended fields
We should store the following properties from webhooks:

At minimum:

* notification.occured_at
* subscription.id
* subscription.status
* subscription.items[].price.id, subscription.items[].quantity
* subscription.items[].price.product_id
* subscription.collection_mode
* subscription.scheduled_change

Recommended:

* subscription.status
* subscription.next_billed_at
* subscription.current_billing_period
* transaction.details.totals
* transaction.payments[].method_details
* subscription.billing_details

And ensure per webhook we should check `occurred_at` as there's no order guarentee.

MelissaAutumn:
This PR re-works the paddle models and adds support for paddle webhooks. Additionally due to the soft-requirement for the paddle client I've added an 'extras' for subscription, so if someone doesn't need the subscription app they can just skip the requirements.

Changes:
* Added Price, Plan, SubscriptionItem (not used yet), and Transaction models to Subscription app.
* Added webhooks for capturing Paddle's transaction and subscription events
* Adjusted Paddle frontend checkout flow to include the logged in user's uuid (signed to prevent tampering) so we can tie webhooks to users. 
* Added a couple of webhook fixtures from Paddle's event documentation (this data is fake)
* Added PADDLE_WEBHOOK_KEY and PADDLE_API_KEY env variables.

There's some additional work like pulling in SubscriptionItems, and tying the Plan model to access but I'd like to test webhooks so I need this on stage. 😅 

The pulumi changes for the new secret will be pushed up separately.


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "thunderbird_accounts"
-version = "0.3.3"
+version = "0.3.99"
```
MelissaAutumn, line 7: Paddle will make it 0.4.

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -65,7 +64,7 @@ def __call__(self, request: HttpRequest):
 
         end = time.perf_counter_ns()
 
-        logging.debug(f'> total time {(end - start) / 1000000} ms')
+        sentry_sdk.set_measurement('cached_allowed_host_get_time', (end - start) / 1000000, 'millisecond')
```
MelissaAutumn, line 67: Got annoyed at some logging, switched it to sentry metrics. We probably don't need this anymore but easy to ship it to sentry instead.

src/thunderbird_accounts/subscription/decorators.py
```diff
@@ -0,0 +1,24 @@
...
+    def _inject_paddle(*args, **kwargs):
+        if Options:
+            options = Options(Environment.SANDBOX if settings.PADDLE_ENV == 'sandbox' else Environment.PRODUCTION)
+        else:
+            options = None
+
+        kwargs['paddle'] = Client(settings.PADDLE_API_KEY, options=options) if Client else None
+        return func(*args, **kwargs)
+
+    return _inject_paddle
```
MelissaAutumn, line 24: I ended up not needing this, but figured we may need it in the future.

src/thunderbird_accounts/authentication/permissions.py
```diff
@@ -122,3 +128,16 @@ def authenticate(self, request):
...
+    def authenticate(self, request):
+        if not Verifier or not Secret:
+            logging.error('Paddle package is not installed. This webhook has been rejected.')
+            return None
+
+        integrity_check = Verifier().verify(request, Secret(settings.PADDLE_WEBHOOK_KEY))
+
+        # We need to return a user, but we don't need the user for these requests
+        # So return an empty user object
+        return User(), None if integrity_check else None
```
rwood-moz, line 143: `None if integrity_check else None`

Is the first `None` a typo?

rwood-moz: I take this as it will always return `User()` but only return a second argument on condition (but always returns `None`)

MelissaAutumn: Good catch, that should be the first argument!

rwood-moz (approved):
So cool! LGTM.

src/thunderbird_accounts/subscription/models.py
```diff
@@ -1,57 +1,229 @@
...
+    """Holds a paddle id with some sensible defaults, but is otherwise just a CharField"""
 
-    :param paddle_id: The customer id field in paddle
-    :param name: The name of the customer
-    :param email: The email of the customer
-    """
-    paddle_id = models.CharField(max_length=256)
+    def __init__(self, *args, db_collation=None, **kwargs):
+        kwargs = {'max_length': 256, 'db_index': True, 'null': True, 'default': None, **kwargs}
+        super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
MelissaAutumn: Yep, just a standardized way to define the paddle model ids.

PR #151, opened 2025-05-28
Install Keycloak in stage

ryanjjung:
This PR introduces a new Fargate cluster to run KeyCloak on.

While introducing the new service, it became apparent that some rejankering with `__main__.py` was necessary to avoid writing more and more of the same code pattern to set security group rule sources. That has been reconfigured to loop over YAML mapping entries and to use the service names to line up all the rules.

Finally, this repo had been pinned to tb_pulumi's main branch to take advantage of a feature that had not been released. It has since been released, so I have pinned this to that version.

### DIFF
--- a/pulumi/Pulumi.stage.yaml
+++ b/pulumi/Pulumi.stage.yaml
@@ -40,3 +40,13 @@ config:
     secure: AAABALvrXmWRSnWv5rAY9ZOC0i+b1cjaHv3ON4xnrykNFC8AAl97TU3LW482PB1jAftoFRqgT0oxfkRle5gvxjfvz69qppZR6antftiGvwa3NXjf2yuPk16B95ephQBeszjr+lOR
   accounts:paddle-api-key:
     secure: AAABAGvRiB3sw8q8fjWKnbDYD+YWmIJ2FoTk911ZUtbSPvhgXSmSCdYfYdOqT4qqVn/btZm4RnqWb/ZKiHnTWz+uq5jcd3vHiIYHvNh/pAsVJTb/NPZwNiVBoMYQ6d4B7xGXoyY=
+  accounts:keycloak-database-host:
+    secure: AAABABvUj9oQDBM4eIxCdWfO2cucEL5Hi/pgIVJqnEyKzL43bFeIcjlv8qDRsh2746Cyh1He3FYMgI+o4QkzcNeVPbUlWdqsC9ZVax/qjb0rk8u1O9tf4Q==
+  accounts:keycloak-database-name:
+    secure: AAABAJmBTRWkmJ1XQ5Hwaar39+g/hcOnsVPLy/xovtgYKDjIav1qkQ==
+  accounts:keycloak-database-user:
+    secure: AAABANqTjC2te5v5jJpJF/hOplSIzzIIHaIAm2Otks7oHymg3lYUtvF9KbunqQ==
+  accounts:keycloak-database-password:
+    secure: AAABADZZBpSAN5LhftkPxRF5Kc0FhNgCU65V8Zp0VwxuhzQHuyySFMrleK5NDw/c
+  accounts:keycloak-admin-password:
+    secure: AAABAJickUKn6Zxpqow593gbQbotzoY+vNQBG0xXhOQrRsaCtCs6XpSN55UXSRpig5vBtFPO4vPopG+/rQ==
--- a/pulumi/__main__.py
+++ b/pulumi/__main__.py
@@ -10,6 +10,11 @@ import tb_pulumi.fargate
 import tb_pulumi.network
 import tb_pulumi.secrets
 
+
+MSG_LB_MATCHING_CONTAINER='In this stack, container security groups must have matching load balancer security groups.'
+MSG_LB_MATCHING_CLUSTER='In this stack, Fargate clusters must have matching load balancer security groups.'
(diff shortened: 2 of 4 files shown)


pulumi/Pulumi.stage.yaml
```diff
@@ -40,3 +40,13 @@ config:
...
     secure: AAABAGvRiB3sw8q8fjWKnbDYD+YWmIJ2FoTk911ZUtbSPvhgXSmSCdYfYdOqT4qqVn/btZm4RnqWb/ZKiHnTWz+uq5jcd3vHiIYHvNh/pAsVJTb/NPZwNiVBoMYQ6d4B7xGXoyY=
+  accounts:keycloak-database-host:
+    secure: AAABABvUj9oQDBM4eIxCdWfO2cucEL5Hi/pgIVJqnEyKzL43bFeIcjlv8qDRsh2746Cyh1He3FYMgI+o4QkzcNeVPbUlWdqsC9ZVax/qjb0rk8u1O9tf4Q==
+  accounts:keycloak-database-name:
+    secure: AAABAJmBTRWkmJ1XQ5Hwaar39+g/hcOnsVPLy/xovtgYKDjIav1qkQ==
+  accounts:keycloak-database-user:
+    secure: AAABANqTjC2te5v5jJpJF/hOplSIzzIIHaIAm2Otks7oHymg3lYUtvF9KbunqQ==
+  accounts:keycloak-database-password:
+    secure: AAABADZZBpSAN5LhftkPxRF5Kc0FhNgCU65V8Zp0VwxuhzQHuyySFMrleK5NDw/c
+  accounts:keycloak-admin-password:
```
ryanjjung, line 51: This is the password for the `temp-admin` account that Mailstrom creates at bootstrapping time. I have already set up a real admin user and disabled this in stage.

pulumi/config.stage.yaml
```diff
@@ -20,49 +20,101 @@ resources:
...
+              from_port: 443
+              to_port: 443
+          egress:
+            - description: Allow outbound traffic to anywhere
+              protocol: tcp
+              from_port: 0
+              to_port: 65535
+              cidr_blocks:
+                - 0.0.0.0/0
+      keycloak-mgmt: null
```
ryanjjung, line 62: We do not expose the management port to anything external.

pulumi/__main__.py
```diff
@@ -58,36 +65,36 @@
...
     **redis_opts,
 )
 
 # Build Fargate clusters to run our containers
-fargate_clusters = {
-    cluster: tb_pulumi.fargate.FargateClusterWithLogging(
-        name=f'{project.name_prefix}-fargate-{cluster}',
+fargate_clusters = {}
+for service, opts in resources['tb:fargate:FargateClusterWithLogging'].items():
+    lb_sg_ids = [lb_sgs[service].resources['sg'].id] if lb_sgs[service] else []
```
MelissaAutumn, line 82: Do we need `(service in lb_sgs and lb_sgs[service])` or will service always be in lb_sgs and will never ever through a key error? 😄

MelissaAutumn: We could also wrap this in a try/catch and yell the developer to fix their stuff if service is in fargate but not sg.

pulumi/__main__.py
```diff
@@ -58,36 +65,36 @@
...
+    depends_on = [
+        container_sgs[service].resources['sg'],
+        *vpc.resources['subnets'],
+    ]
+    if lb_sgs[service]:
```
MelissaAutumn, line 87: again here, check if service is in lb_sgs first, otherwise this may throw a key error.

pulumi/config.stage.yaml
```diff
@@ -20,49 +20,101 @@ resources:
...
+          ingress:
+            - description: Allow traffic from the load balancer to the service port
+              protocol: tcp
+              from_port: 8080
+              to_port: 8080
+            # KeyCloak exposes health checks on the management port. Don't route any real traffic here.
+            - description: Allow traffic from the load balancer to the management port
+              protocol: tcp
+              from_port: 9000
+              to_port: 9000
```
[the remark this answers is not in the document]
MelissaAutumn: cool and I'm guessing the keycloak management api? (I haven't looked where this is kept)

pulumi/config.stage.yaml
```diff
@@ -103,6 +149,77 @@ resources:
...
+            port: 9000
+      task_definition:
+        network_mode: awsvpc
+        cpu: 512
+        memory: 2048
+        requires_compatibilities:
+          - FARGATE
+        container_definitions:
+          keycloak:
+            image: quay.io/keycloak/keycloak:26.2
```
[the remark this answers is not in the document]
MelissaAutumn: That's an odd choice, I wonder if they follow semver properly or if minor versions can potentially break installs.

MelissaAutumn (approved):
looks good, one small fix and you should be good to merge!

pulumi/__main__.py
ryanjjung: Hrm, I'm not sure. A KeyError seems likely. I know I threw a few when I was initially working through this. It's how I knew things weren't right. I left a comment in the config file, which is where I imagine most changes will happen, but I can implement a more robust input validation here.

pulumi/__main__.py
ryanjjung: I [pushed a commit](https://github.com/thunderbird/thunderbird-accounts/pull/151/commits/6aa29bf830645e80c10659b66b98b277f4da1cd4) to validate all of these inputs.

PR #158, opened 2025-05-29
Update subscribe page to use values from DB

### LINKED ISSUE #154: Update subscribe page to use database values instead of hardcoded values
Right now we shove in 3 price object ids into our env and use that to retrieve price info from paddle. However we store the price, and products now. We should retrieve the approriate information from our plans.

MelissaAutumn:
Fixes #154 

We now pull out price ids from our existing plan models. We then push it down to the frontend. I made an mistake when initially making products / pricing. We actually want 1 product per plan, and then multiple price schemes on that product. 

This obviously doesn't mean we'll use them, but it's now structured so it will work however we want it. This means that I had to restructure how the prices are displayed, so I might need to tweak some e2e tests. 

Additionally I noticed archived products and prices weren't being pulled, so I fixed that here too.


MelissaAutumn:
That's most likely due to you not setting up a plan object in the admin panel. I'll fix that spot to hide it if there's no plan object. 

Also I'll fix the e2e tests.

MelissaAutumn:
Phew, I'm surprised on how easy of a fix that was. 

Will get on those docs after lunch. But to run e2e tests you'll need to setup 3 plan objects with the product ids of the new products (https://sandbox-vendors.paddle.com/products-v2 , login in 1pass)

Feel free to dm me if you run into any issues with that.

test/e2e/README.md
```diff
@@ -34,10 +34,8 @@ In the top level `.env` file, add the values from the "Paddle Sandbox env vars"
 
 ```dotenv
 PADDLE_TOKEN=<token>
+PADDLE_API=<token>
```
rwood-moz, line 37: Should be `PADDLE_API_KEY`

rwood-moz: Ahhh makes sense, ok thanks

MelissaAutumn:
Ooh just reread what I wrote in the readme. The actual product ids are in 1password, the one in the readme is a fake one.

Can you confirm that the 1pass ones work?

rwood-moz:
> Ooh just reread what I wrote in the readme. The actual product ids are in 1password, the one in the readme is a fake one.
> 
> Can you confirm that the 1pass ones work?

Yep I've been using the ones in 1password. If you want to land this anyway I'm fine with that since it works fine in CI, just something I must have wrong in my local setup.

rwood-moz (approved):

MelissaAutumn:
I'll move this forward, and will look into further improvements for the next PR!

Issue #173, opened 2025-06-23
Host a zendesk form for all our products


malini:
Right now, we provide a link to [this zendesk form](https://tbpro.zendesk.com/hc/en-us/requests/new) for any support requests, which is a contact form hosted by Zendesk. Zendesk no longer supports embeddable forms, so we'll have to use Zendesk's API to host the form on our domains.

**Describe the solution you'd like**

We should create an accounts.tb.pro/contact page which uses the zendesk API to create and submit a Zendesk support issue. Use the same fields as our current form, but keep in mind that we may want to have conditional fields per service in the future.

Ex: If a customer selects "Thunderbird Appointment" as the "Product", then we may want to add additional optional fields like "Calendar provider".

rtanglao:
> For the [TB Pro Zendesk form](https://tbpro.zendesk.com/hc/en-us/requests/new) we are using some custom_fields for "Type of Request" (field_key: 38214346931987) and "Product" (field_key: 38815672253843) and I hit a validation error for "Type of Request" specifically.
> May I ask someone to confirm the `field_key` IDs and what underlying values are accepted for both "Product" and "Type of Request" exactly? Are there other custom values that I should be capturing?

Confirming:
* `Product`: `38815672253843`
![Image](https://github.com/user-attachments/assets/bc8cccfe-d58f-4cda-ba8c-b1ad4611165a)

* `Type of request`: `38214346931987`
![Image](https://github.com/user-attachments/assets/347337da-d31f-49cb-b0d6-382924ffbeba)

> Perhaps also worth noting that the API Key that was shared with me somehow doesn't authenticate when trying to make requests to the endpoint that would grant me the info (`GET` to `/api/v2/ticket_fields`). Unsure how this works from the Zendesk dashboard perspective but the API Key + user email combo needs to have at least permissions to:
> 
>     * Create attachments
> 
>     * Create tickets (either from the Requests API or the Tickets API)
> 
> 
> cc [@rtanglao](https://github.com/rtanglao)

rtanglao:
@laurelterlesky @inattee @malini :  What is the official name for Thundermail? We need it for the Zendesk support form that Davi is implementing in this github issue.
* e.g. `Thunderbird Email` to be consistent with `Thunderbird Send` and `Thunderbird Appointment`

davinotdavid:
[...]

PR #175, opened 2025-06-24
Host a Zendesk form for all our products

davinotdavid:
## Description of the Change

[**Backend changes**]
- Adds a new `POST /contact` route for accounts that hosts a contact form (the form itself is a Vue component) that uses the [Zendesk Requests API](https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#tickets-and-requests) to replace the [current Zendesk form](https://tbpro.zendesk.com/hc/en-us/requests/new).
- Adds a new `POST /contact/attach_file` route to upload a file to Zendesk and generate a token to be later attached to the ticket creation. [[Zendesk Docs for Attaching Files](https://developer.zendesk.com/documentation/ticketing/managing-tickets/adding-ticket-attachments-with-the-api/)]

[**Frontend changes**]
- Added a new `ContactSupportVue` that uses the `services-ui` form components for visual consistency (see image below). Please note that the UI itself will probably evolve / change very soon so please focus on the functionality itself for the time being.
 
image

> [!IMPORTANT]  
> The original Zendesk form is configured with `custom_fields` that require specific field IDs to work. The Vue form values (the select values) and the Python env vars related to those need to be in sync with whatever is configured in Zendesk. In other words, changing Zendesk's `custom_fields` will break the form if they don't get updated in the code. More info [here](https://github.com/thunderbird/thunderbird-accounts/issues/173#issuecomment-3008760622).

## Benefits

- Be able to host the form in our domains instead of redirecting the user to Zendesk's while maintaining similar functionality (potentially even improving on current functionality as we might be a little more flexible on fields).

## Applicable Issues

https://github.com/thunderbird/thunderbird-accounts/issues/173


templates/mail/contact.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'base.html' %}
+{% block body %}
+  <h2>Submit a request</h2>
+  <span>Fields marked with an asterisk (*) are required.</span>
+
+  <div id="contactSupportForm"></div>
+{% endblock %}
```
MelissaAutumn, line 7: Thanks for dealing with my wacky vue integration. It's an experiment so we can discuss how we like this workflow later.

assets/app/vue/components/ContactSupportForm.vue
```diff
@@ -0,0 +1,312 @@
...
+})
+const fileInput = ref(null)
+
+// Product options for SelectInput
+// These values are important to match the Zendesk ticket custom field type
+const productOptions = [
+  { label: 'Thunderbird Assist', value: 'thunderbird_assist' },
+  { label: 'Thunderbird Appointment', value: 'thunderbird_appointment' },
+  { label: 'Thunderbird Send', value: 'thunderbird_send' }
+];
```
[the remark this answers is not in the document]
davinotdavid: Hmm I thought these need to be the exact values as expected from the Zendesk form configs so I got those from here: https://github.com/thunderbird/thunderbird-accounts/issues/173#issuecomment-3008760622

assets/app/vue/components/ContactSupportForm.vue
```diff
@@ -0,0 +1,312 @@
...
+
+    <ul v-if="form.attachments.length" class="attachment-list">
+      <li v-for="(attachment, index) in form.attachments" :key="index">
+        <span class="attachment-filename">{{ attachment.filename }}</span>
+        <button @click="removeAttachment(index)" class="remove-attachment-btn">Remove</button>
+      </li>
+    </ul>
+
+    <notice-bar type="error" v-if="errorText" class="notice">{{ errorText }}</notice-bar>
+    <notice-bar type="success" v-if="successText" class="notice">{{ successText }}</notice-bar>
```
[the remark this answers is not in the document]
davinotdavid: Sure! Just confirming, do you mean here?

image

src/thunderbird_accounts/mail/views.py
```diff
@@ -100,6 +101,86 @@ def self_serve_common_options(is_account_settings: bool, user: User, account: Ac
...
+    if zendesk_api_response.ok:
+        return JsonResponse({'success': True})
+
+    return JsonResponse({'success': False}, status=500)
+
+
+@login_required
+@require_http_methods(['POST'])
+def contact_attach_file(request: HttpRequest):
+    """ Uses Zendesk's API to generate a token to be used for contact form attachments
```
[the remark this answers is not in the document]
davinotdavid: For sure, that makes a lot of sense, thanks!

src/thunderbird_accounts/mail/views.py
```diff
@@ -100,6 +101,86 @@ def self_serve_common_options(is_account_settings: bool, user: User, account: Ac
...
+@login_required
+@require_http_methods(['POST'])
+def contact_submit(request: HttpRequest):
+    """ Uses Zendesk's Requests API to create a ticket
+        Ref https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#tickets-and-requests"""
+
+    if request.user.is_anonymous:
+        return JsonResponse({'success': False})
+
+    data = json.loads(request.body)
```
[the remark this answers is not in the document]
davinotdavid: oh interesting! I tried to find `request.json` but the closest I've found is request.POST which returns a dict if the submission is a form data instead of JSON so I've changed the implementation to form data instead so we can leverage it

assets/app/vue/components/ContactSupportForm.vue
```diff
@@ -0,0 +1,312 @@
...
+      name="description"
+      type="textarea"
+      v-model="form.description"
+      required="required"
+      data-testid="contact-description-input"
+    >
+      Description*
+    </text-input>
+
+    <div
```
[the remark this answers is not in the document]
davinotdavid: Oh yeah for sure, added! Thanks!

.env.example
```diff
@@ -74,3 +74,11 @@ CELERY_BROKER=redis://redis:6379
...
 CELERY_EAGER=False
 
+# Zendesk (contact support form)
+ZENDESK_SUBDOMAIN=
+ZENDESK_USER_EMAIL=
+ZENDESK_API_TOKEN=
+
+# Zendesk (contact support form) custom fields
+ZENDESK_PRODUCT_FIELD_ID=
+ZENDESK_TYPE_OF_REQUEST_FIELD_ID=
```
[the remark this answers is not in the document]
davinotdavid: Now getting the fields dynamically! Currently, there is only a specific user that have access to that endpoint so I will DM you the credentials to be used for testing this. Thanks! 🙇

src/thunderbird_accounts/mail/views.py
```diff
@@ -100,6 +101,123 @@ def self_serve_common_options(is_account_settings: bool, user: User, account: Ac
...
+                    }
+                    for option in field['custom_field_options']
+                ]
+            }
+            fields_with_options.append(field_data)
+
+    return JsonResponse({
+        'success': True,
+        'ticket_fields': fields_with_options
+    })
```
MelissaAutumn, line 150: Since we control the returb structure here, we should return the data in a way that the frontend can just use without any post processing.

I think making field_with_options a dict keyed by title would work

assets/app/vue/components/ContactSupportForm.vue
```diff
@@ -0,0 +1,354 @@
...
+const fetchTicketFields = async () => {
+  try {
+    const response = await fetch('/contact/fields', {
+      method: 'GET',
+      credentials: 'include',
+    });
+
+    const data = await response.json();
+
+    if (data.success && data.ticket_fields) {
```
MelissaAutumn, line 74: We should be able to just pick the field type out of the return value and set it without any looping.

assets/app/vue/components/ContactSupportForm.vue
```diff
@@ -0,0 +1,354 @@
...
+      body: formData,
+      headers: {
+        'X-CSRFToken': csrfToken.value
+      },
+    });
+
+    const data = await response.json();
+
+    if (!data.success) {
+      errorText.value = 'Failed to submit contact form';
```
MelissaAutumn, line 154: Do we get an error from the server? if so please show the error on the frontend

src/thunderbird_accounts/mail/views.py
```diff
@@ -100,6 +101,123 @@ def self_serve_common_options(is_account_settings: bool, user: User, account: Ac
...
+@require_http_methods(['GET'])
+def contact_fields(request: HttpRequest):
+    """Get ticket fields from Zendesk API and filter for fields with custom options."""
+    zendesk_client = ZendeskClient()
+    result = zendesk_client.get_ticket_fields()
+
+    if not result['success']:
+        return JsonResponse({
+            'success': False,
+            'error': result.get('error', 'Failed to fetch ticket fields')
```
MelissaAutumn, line 122: On the backend can you wrap text with _( ... ). Eventually we'll have localization, and _() is just an easy way to extract and replace strings.

MelissaAutumn (changes requested):
Some minor things, looking good!

src/thunderbird_accounts/mail/views.py
davinotdavid: Ah that makes sense, much better indeed! Changed [in this commit](https://github.com/thunderbird/thunderbird-accounts/pull/175/commits/da9b0b68f64152b282ee472a396a1d2ff7feba50).

assets/app/vue/components/ContactSupportForm.vue
davinotdavid: Yep, that makes sense! Thanks! Changed [in this commit](https://github.com/thunderbird/thunderbird-accounts/pull/175/commits/da9b0b68f64152b282ee472a396a1d2ff7feba50).

src/thunderbird_accounts/mail/views.py
davinotdavid: TIL! Thanks, changed here and will keep that in mind for the next tasks on accounts 🙇

PR #179, opened 2025-07-02
Update E2E test to check new paddle consent checkbox

### LINKED ISSUE #178: Update E2E tests for new paddle confirm consent check box
Paddle has added a new `confirm your consent to continue` check box that must be checked on in order to continue with the subscription checkout process. This breaks the current E2E tests. Update the [E2E checkout test](https://github.com/thunderbird/thunderbird-accounts/blob/main/test/e2e/tests/checkout.spec.ts) accordingly.

Thank you @davinotdavid for making us aware of this 😸

rwood-moz:
Fixes #178.

### DIFF
--- a/test/e2e/pages/paddle-frame.ts
+++ b/test/e2e/pages/paddle-frame.ts
@@ -8,6 +8,7 @@ export class PaddleFrame {
   readonly countryField: Locator;
   readonly postalCodeField: Locator;
   readonly continueButton: Locator;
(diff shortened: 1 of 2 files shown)


rwood-moz:
Note that this will fix the E2E tests when they run against PRs or your local stack. However when running on stage (ie. after stage deployments) the E2E tests fail because of a separate issue #180.

PR #184, opened 2025-07-15
Fix issue with non-raw jq output producing double-doublequotes

ryanjjung:
The `-r` argument to `jq` means it produces the "raw" output as opposed to the thing you queried with double-quotes around it. Leaving it off here results in newimage.json having some bad output with double double-quotes. A mockup of how it looks:

`image: ""host/image:tag""`

Restoring the `-r` fixes this.


MelissaAutumn (approved):
lgtm

Issue #185, opened 2025-07-16
Make Zendesk contact form loginless


davinotdavid:
Currently the Zendesk form on the path `/contact` is `@login_required` but users may have problems regarding login / signup so it would be better for it not to require login

cc @rtanglao

PR #186, opened 2025-07-16
Make Zendesk form loginless

davinotdavid:
Remove the login requirements for the Zendesk related endpoints to enable a loginless form submission + tests

PS: Currently all the e2e test suites are tied up with the `auth.setup.ts` as a prerequisite but perhaps in the future it would be nice IMHO to decouple that for test suites that rely on non-logged in behavior. I am using `clearCookies` for now as doing a full logout interfere with other test suites.

Solves https://github.com/thunderbird/thunderbird-accounts/issues/185


devmount (approved):
I love how those decorators make implementing sth like this a lot easier! Thanks for making those public.

Issue #200, opened 2025-07-18
Zendesk tickets are being filed as the Agent – not the end user


davinotdavid:
The following was reported by the support team: 

"The current tickets are coming through as a proactive ticket, e.g. a ticket by an agent on behalf of a user, rather than submitted by the end user. This will have an impact on metrics and reporting, so we do want to get this to be submitted by the end user."

My guess is that fetching the custom fields need to be done using the `zendesk-user-email` env var but the actual submitting of tickets needs to be done with the end users' email paired with our API key.

PR #201, opened 2025-07-18
Make Zendesk form produce tickets with correct triggers

davinotdavid:
## Description of change
If we send a request to the Requests API to create a ticket using the `ZENDESK_USER_EMAIL`, it will produce a `proactive ticket` which means a ticket created by an agent on the users behalf.

However, this endpoint should be called using the end users' email + API key combo for auth.

Note that the `GET` requests to the dynamically get the `custom_fields` still needs to be done with the `ZENDESK_USER_EMAIL` + API key combo otherwise it fails auth.

## Solves
https://github.com/thunderbird/thunderbird-accounts/issues/200


devmount (approved):
Thanks!

PR #213, opened 2025-07-22
Add OIDC secrets and flip AUTH_SCHEME to oidc

MelissaAutumn:
Forgot we removed FXA out of the PR, so I have to setup the oidc vars now. 

I think there will still be some keycloak-side client configuration but I'll work through it as it happens.


davinotdavid (approved):
LGTM 🚀

PR #222, opened 2025-07-23
Add the env `OIDC_FALLBACK_MATCH_BY_EMAIL` that when true matches incoming oidc claims by email if oidc_id fails.

### LINKED ISSUE #221: Match old users with the new oidc system on stage

MelissaAutumn:
Fixes #221 

I wasn't expecting us to not nuke stage with the oidc move. But with this option it doesn't matter. Once we're good we can disable it so incoming users from the oidc login flow will be matched only by oidc_id (or created if not matched.)

### DIFF
--- a/src/thunderbird_accounts/settings.py
+++ b/src/thunderbird_accounts/settings.py
@@ -328,6 +328,10 @@ if AUTH_SCHEME == 'oidc':
 
     OIDC_OP_LOGOUT_URL_METHOD = 'thunderbird_accounts.settings.oidc_logout'
 
+    # this should only be used to transition users from one oidc to another
+    # ideally you should turn this off when you're done.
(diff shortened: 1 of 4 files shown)


src/thunderbird_accounts/settings.py
```diff
@@ -328,6 +328,10 @@ def oidc_logout(request):
 
     OIDC_OP_LOGOUT_URL_METHOD = 'thunderbird_accounts.settings.oidc_logout'
 
+    # this should only be used to transition users from one oidc to another
+    # ideally you should turn this off when you're done.
+    OIDC_FALLBACK_MATCH_BY_EMAIL = os.getenv('OIDC_FALLBACK_MATCH_BY_EMAIL', '').lower() == 'True'
```
davinotdavid, line 333: I think `'True'.lower() == 'True` would be false here, perhaps we should drop the lower()?

MelissaAutumn: Whoops, I'll keep the lower, but fix the comparison

PR #226, opened 2025-07-29
Revamp release process

ryanjjung:
Adding some process to do the following:

- Deploy using the infra code that existed at the time the merge happened, not what's in main at the time of release.
- Create draft releases when changes are merged. This contains the compressed infra code and a JSON file from which the git-hashed tag and the semantic version declared in the code at the time of the build can be extracted.
- When such a release is published, the git-hashed image is retagged to be a semver-tagged image. That image is pushed and then the Pulumi configs are updated. The services are deployed with the semver-tagged image name.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -81,9 +81,9 @@ STALWART_API_AUTH_METHOD=basic
 # Keycloak's (or another oidc provider) client settings
 OIDC_CLIENT_ID=tb-accounts
 OIDC_CLIENT_SECRET=KoPz6llGgyFJPJvn21i3FTtBiqLj2Zbd
-OIDC_URL_AUTH=http://keycloak:8999/realms/tbpro/protocol/openid-connect/auth
-OIDC_URL_TOKEN=http://keycloak:8999/realms/tbpro/protocol/openid-connect/token
-OIDC_URL_USER=http://keycloak:8999/realms/tbpro/protocol/openid-connect/userinfo
-OIDC_URL_JWKS=http://keycloak:8999/realms/tbpro/protocol/openid-connect/certs
-OIDC_URL_LOGOUT=http://keycloak:8999/realms/tbpro/protocol/openid-connect/logout
+OIDC_URL_AUTH=http://localhost:8999/realms/tbpro/protocol/openid-connect/auth
+OIDC_URL_TOKEN=http://localhost:8999/realms/tbpro/protocol/openid-connect/token
+OIDC_URL_USER=http://localhost:8999/realms/tbpro/protocol/openid-connect/userinfo
+OIDC_URL_JWKS=http://localhost:8999/realms/tbpro/protocol/openid-connect/certs
+OIDC_URL_LOGOUT=http://localhost:8999/realms/tbpro/protocol/openid-connect/logout
 OIDC_SIGN_ALGO=RS256
--- a/.env.example
+++ b/.env.example
@@ -81,9 +81,9 @@ STALWART_API_AUTH_METHOD=basic
 # Keycloak's (or another oidc provider) client settings
 OIDC_CLIENT_ID=tb-accounts
 OIDC_CLIENT_SECRET=KoPz6llGgyFJPJvn21i3FTtBiqLj2Zbd
-OIDC_URL_AUTH=http://localhost:8999/realms/tbpro/protocol/openid-connect/auth
-OIDC_URL_TOKEN=http://localhost:8999/realms/tbpro/protocol/openid-connect/token
-OIDC_URL_USER=http://localhost:8999/realms/tbpro/protocol/openid-connect/userinfo
-OIDC_URL_JWKS=http://localhost:8999/realms/tbpro/protocol/openid-connect/certs
-OIDC_URL_LOGOUT=http://localhost:8999/realms/tbpro/protocol/openid-connect/logout
+OIDC_URL_AUTH=http://keycloak:8999/realms/tbpro/protocol/openid-connect/auth
+OIDC_URL_TOKEN=http://keycloak:8999/realms/tbpro/protocol/openid-connect/token
+OIDC_URL_USER=http://keycloak:8999/realms/tbpro/protocol/openid-connect/userinfo
+OIDC_URL_JWKS=http://keycloak:8999/realms/tbpro/protocol/openid-connect/certs
+OIDC_URL_LOGOUT=http://keycloak:8999/realms/tbpro/protocol/openid-connect/logout
 OIDC_SIGN_ALGO=RS256
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -137,7 +137,7 @@ jobs:
         id: iac-archive
         uses: actions/upload-artifact@v4
         with:
-          name: pulmi
+          name: pulumi
           path: pulumi.tbz
       
       # Deploy to stage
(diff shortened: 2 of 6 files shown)


.github/workflows/merge.yml
```diff
@@ -66,7 +67,7 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
-          aws-region: ${{ vars.AWS_DEFAULT_REGION }}
+          aws-region: eu-central-1
```
ryanjjung, line 70: Hardcoding this for now because (a) it's true, and (b) I think I can't change this variable without breaking other things.

.github/workflows/merge.yml
```diff
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  contents: read
+  contents: write
```
ryanjjung, line 15: This is needed to create a release.

.github/workflows/release.yml
```diff
@@ -16,80 +16,76 @@ permissions:
...
         with:
           python-version: ${{ vars.PYTHON_VERSION}}
 
-      - name: Set up virtual environment
-        if: contains(github.event.release.assets.*.name, 'deployment.zip')
-        shell: bash
-        run: |
-          python -m pip install virtualenv
-          cd pulumi
-          virtualenv ./venv
```
ryanjjung, line 79: Pulumi manages its own virtual environment, even installing from requirements.txt, without us instructing it to do so. We can snip out all of that kind of code.

pulumi/config.stage.yaml
```diff
@@ -255,7 +255,7 @@ resources:
           - FARGATE
         container_definitions:
           accounts:
-            image: 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:cf7fc9c8b2cf37f6b1f0ae443779cc187f21cbb7
+            image: 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:cd16087ea05b1a375f6017330b7a4a4eae0d9d0a
```
ryanjjung, line 258: Currently deployed image

.github/workflows/merge.yml
```diff
@@ -126,7 +127,19 @@ jobs:
         with:
           name: deployment
           path: deployment.json
-
+      
+      - name: Package Pulumi code
+        shell: bash
+        run:
+          tar -cvjf pulumi.tbz pulumi/
```
MelissaAutumn, line 134: nice keyboard smash

MelissaAutumn (approved):
lgtm!

aatchison (approved):
LGTM, nice!

.env.example
```diff
@@ -81,9 +81,9 @@ STALWART_API_AUTH_METHOD=basic
...
-OIDC_URL_AUTH=http://keycloak:8999/realms/tbpro/protocol/openid-connect/auth
-OIDC_URL_TOKEN=http://keycloak:8999/realms/tbpro/protocol/openid-connect/token
-OIDC_URL_USER=http://keycloak:8999/realms/tbpro/protocol/openid-connect/userinfo
-OIDC_URL_JWKS=http://keycloak:8999/realms/tbpro/protocol/openid-connect/certs
-OIDC_URL_LOGOUT=http://keycloak:8999/realms/tbpro/protocol/openid-connect/logout
+OIDC_URL_AUTH=http://localhost:8999/realms/tbpro/protocol/openid-connect/auth
+OIDC_URL_TOKEN=http://localhost:8999/realms/tbpro/protocol/openid-connect/token
+OIDC_URL_USER=http://localhost:8999/realms/tbpro/protocol/openid-connect/userinfo
+OIDC_URL_JWKS=http://localhost:8999/realms/tbpro/protocol/openid-connect/certs
+OIDC_URL_LOGOUT=http://localhost:8999/realms/tbpro/protocol/openid-connect/logout
```
MelissaAutumn, line 88: The only problem here is that for the backend container these routes aren't accessible via localhost. These routes are used for both client and server, which is why I'm using the docker dns name here, and also I have a note in the readme to add it to your /etc/hosts. 

If you have a good solution here lmk!

.env.example
```diff
@@ -81,9 +81,9 @@ STALWART_API_AUTH_METHOD=basic
 # Keycloak's (or another oidc provider) client settings
 OIDC_CLIENT_ID=tb-accounts
 OIDC_CLIENT_SECRET=KoPz6llGgyFJPJvn21i3FTtBiqLj2Zbd
-OIDC_URL_AUTH=http://keycloak:8999/realms/tbpro/protocol/openid-connect/auth
-OIDC_URL_TOKEN=http://keycloak:8999/realms/tbpro/protocol/openid-connect/token
-OIDC_URL_USER=http://keycloak:8999/realms/tbpro/protocol/openid-connect/userinfo
-OIDC_URL_JWKS=http://keycloak:8999/realms/tbpro/protocol/openid-connect/certs
-OIDC_URL_LOGOUT=http://keycloak:8999/realms/tbpro/protocol/openid-connect/logout
+OIDC_URL_AUTH=http://localhost:8999/realms/tbpro/protocol/openid-connect/auth
```
ryanjjung, line 84: These hostname changes are necessary because the `keycloak` hostname only makes any sense from within the docker context. These tests are run against a docker-compose-based setup, but the calls are still made from "outside the house".

.github/workflows/validate.yml
```diff
@@ -130,7 +131,7 @@ jobs:
           echo "PADDLE_API_KEY=$PADDLE_API_KEY" >> .env
 
       - name: Start appointment stack via docker
-        run: docker compose up -d --build -V postgres redis backend frontend
+        run: docker compose up -d --build -V
```
ryanjjung, line 134: This list doesn't contain `keycloak`, leading to the service not being available at test time. Rather than trying to remember to maintain this list, I've chosen to delete it, causing all services in the compose file to come online.

.env.example
ryanjjung: Ok, back to the drawing board with this one, then.

test/e2e/.env.dev.example
```diff
@@ -25,4 +25,11 @@ THUNDERMAIL_USERNAME=admin@example.org
...
-EMAIL_SIGN_UP_ADDRESS=melissa+test@thunderbird.net
\ No newline at end of file
+EMAIL_SIGN_UP_ADDRESS=melissa+test@thunderbird.net
+
+# Custom OIDC paths since this runs from outside the Docker context
+OIDC_URL_AUTH=http://localhost:8999/realms/tbpro/protocol/openid-connect/auth
+OIDC_URL_TOKEN=http://localhost:8999/realms/tbpro/protocol/openid-connect/token
+OIDC_URL_USER=http://localhost:8999/realms/tbpro/protocol/openid-connect/userinfo
+OIDC_URL_JWKS=http://localhost:8999/realms/tbpro/protocol/openid-connect/certs
+OIDC_URL_LOGOUT=http://localhost:8999/realms/tbpro/protocol/openid-connect/logout
```
ryanjjung, line 35: We can override the OIDC routes here where they only apply to the e2e tests, leaving the settings in the "root" .env file.

ryanjjung: This file is what the [automation copies](https://github.com/thunderbird/thunderbird-accounts/blob/main/.github/workflows/validate.yml#L157) over before tacking on secrets. This works for me in my local dev, and it ought to work on the runner as well.

MelissaAutumn (approved):

Issue #196, opened 2025-07-17
Create sign-in page


MelissaAutumn:
## Details

Leverage services-ui to create a fairly basic sign in page. Forgot password should be pretty similar. 

## Zeplin
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/6866f4b0e5fa5c6178005f04

## Screenshots

<img width="420" height="326" alt="Image" src="https://github.com/user-attachments/assets/7915d859-ad66-4078-9d2c-1ab522930667" />

Issue #195, opened 2025-07-17
Add a new bundle for the keycloak theme


MelissaAutumn:
## Details

This might be tricky, we might need to create a different vite project but I'm hopeful that we can just use the existing project and create multiple bundles. 

Multi-page might work? https://vite.dev/guide/build.html#multi-page-app

Currently we have a docker-compose container setup that just acts as the dev server to allow hot-reloading. On container build we build the assets to html/js/css and serve them statically via django without the dev server. 

This bundle should only include the services-ui, and any keycloak js / css and the static files should be included on the template like we do on django. (Using the dev server in dev mode, and the static files on build.)

PR #227, opened 2025-07-30
[WIP] Keycloak Theme Wireframes

### LINKED ISSUE #194: Create a keycloak theme
## Details

We should create a slim theme for keycloak overriding the basic theme and pulling and adjusting where possible. Keycloak designs are a guideline due to the unknowns it may bring. 

Designs indicate a modal, but these pages live in keycloak so they should not be a modal. Give the content a div with a max-width of the modal width, and center that div. (As to basically appear in the same position but without it being a modal.)

MelissaAutumn:
## Details

This currently resolves
#194, #195, #196, and #197.

#198 is still pending...

There's an error page I need to re-trigger, and the final info page needs doing as well.

I've created a new vite config file, and an npm command. Right now you need to manually run it but I plan on bundling it with the dev server before this PR is up for review.

The setup is a bit weird, we're using vue-router and matching on route names instead of paths. The route names are from the template's pageId value (the value in the top left of the screenshots) and are set in App.vue on setup. 

I originally based it off of route name but routes and templates aren't matched uniquely, meaning a "Password link expired" would show up on the login page and other such nonsense.

Additionally I'd like to explore not having to use the window._page system, but I don't think there's a good alternative here besides inlining the SFC or App definition into the individual templates, and that seems a little annoying. 

## Screenshots

Screen Shot 2025-07-30 at 16 30 19-fullpage
Screen Shot 2025-07-30 at 16 30 23-fullpage
Screen Shot 2025-07-30 at 16 30 27-fullpage
Screen Shot 2025-07-30 at 16 30 35-fullpage
Screen Shot 2025-07-30 at 16 30 43-fullpage
Screen Shot 2025-07-30 at 16 30 47-fullpage
Screen Shot 2025-07-30 at 16 30 54-fullpage


MelissaAutumn:
...i also need to add some test ids.

.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{vue,html,css,js,json,yml,ts,ftl}]
+charset = utf-8
+indent_style = space
+indent_size = 2
```
MelissaAutumn, line 10: had to setup a new dev env, so copied over the editorconfig from services-ui, and added the ftl templates too.

keycloak/data/import/tbpro-realm.json
```diff
@@ -1536,7 +1536,7 @@
...
       }
```
MelissaAutumn, line 1540: Made first/last name not required.

keycloak/themes/tbpro/assets/router.js
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    name: 'route-not-implemented',
+    component: RouteNotImplementedView,
+  }
+];
+
+// create router object to export
+const router = createRouter({
+  history: createMemoryHistory(),
```
MelissaAutumn, line 78: We don't want vue-router to actually change routes automatically. I just wanted a sane switch-case for displaying templates. This seems like a nice way to do it within the constraints of keycloak.

keycloak/themes/tbpro/assets/vue/views/ConfigToptView/index.vue
```diff
@@ -0,0 +1,261 @@
...
+// Manual mode settings
+const realmTitle = ref(window._page.realmTitle);
+const username = ref(window._page.currentView?.loginTotp?.username);
+const secretEncoded = ref(window._page.currentView?.loginTotp?.secretEncoded);
+const secret = ref(window._page.currentView?.loginTotp?.secret);
+const qrUrl = ref(window._page.currentView?.loginTotp?.qrUrl);
+const algorithmKey = ref(window._page.currentView?.loginTotp?.algorithmKey);
+const digits = ref(window._page.currentView?.loginTotp?.digits);
+const period = ref(window._page.currentView?.loginTotp?.period);
+const initialCounter = ref(window._page.currentView?.loginTotp?.initialCounter);
```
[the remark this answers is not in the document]
MelissaAutumn: Ah thanks, meant to clean this up. Before I added the otpauth link for the manual route we just displayed a few ugly variables.

keycloak/themes/tbpro/assets/vue/views/DeleteCredentialView/index.vue
```diff
@@ -0,0 +1,61 @@
...
+
+const isCancelSubmit = ref(false);
+
+const onSubmit = () => {
+  loginForm?.value?.submit();
+};
+const onCancel = () => {
+  isCancelSubmit.value = true;
+  onSubmit();
+};
```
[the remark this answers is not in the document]
davinotdavid: Interesting! TIL, thanks for the explanation!

From the `Email Verification` page at least, it seems that I am hitting a `m.submit is not a function` error. The code is not this exact line for this review but it uses a very similar pattern.

image

Not against the work around, just wanted to make sure that I am following / it works!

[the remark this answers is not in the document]
MelissaAutumn: I've added a CancelForm component which we can simply call .cancel on.

MelissaAutumn:
Added a build step to keycloak's dockerfile which builds the theme and copies it over. Also merged in dev-theme into vite-dev so it runs both the accounts dev server and dev-theme. 

And finally fixed the e2e tests. Just need to rebase but good for re-review.

docker-compose.yml
```diff
@@ -62,18 +62,19 @@ services:
...
 
   # Only used when APP_DEBUG=True
-  frontend:
+  vite-dev:
     build:
       context: .
       dockerfile: ./Dockerfile
-    entrypoint: npm run dev
+    # Run both vite devserver for accounts frontend, and the keycloak theme builder
+    entrypoint: bash -c 'npm run dev & npm run dev-theme'
```
davinotdavid, line 77: Nice! Easier to test now and we can skip some steps in the PR description, thanks!

PR #243, opened 2025-08-20
Setup accounts stalwart access

### LINKED ISSUE #241: Setup stalwart oidc for stage / prod

MelissaAutumn:
Fixes #241 

This is mostly just config values for prod and stage, but I've also added a new manual test route for jmap. 

A logged in user can go to `/test/jmap/` and see if `"connected": True`. If so, then our oauth system is working correctly for them. Otherwise we'll get a log error to investigate. Usually though if there's an error it's due to the access token no longer being valid. They just need to re-login. (Our login session probably lasts longer than the access token's life. Which is ok in this case.)


.env.test
```diff
@@ -69,7 +69,8 @@ ZENDESK_API_TOKEN=
 
 
 # This URL should not be called without a mock
-STALWART_API_URL=http://localhost:8080/api
+STALWART_HTTP_URL=http://stalwart:8081
+STALWART_API_URL=http://stalwart:8080
```
MelissaAutumn, line 73: Note: This value has changed!!! Pls update your envs.

davinotdavid: Nit but why have we removed the `/api` from the API_URL and now adding them here https://github.com/thunderbird/thunderbird-accounts/pull/243/files#diff-c13cbce7c93a56c376638de7ea97f0337e1edaae5aed61b603470e997248e6d8R30 ?

MelissaAutumn: keep it consistent. I can rename this stalwart_base_api_url and stalwart_base_jmap_url to be clearer

PR #250, opened 2025-09-02
Create stalwart account on login under certain conditions

MelissaAutumn:
Accounts stage and prod are a bit of mess, so if we don't have an accounts object then either create or partially create one. 

Since this is cross-service it's not easily testable, still thinking about how we can test this but for now this will do.


MelissaAutumn:
Punting those suggestions to https://github.com/thunderbird/thunderbird-accounts/issues/251

Thanks!

PR #252, opened 2025-09-02
Peer accounts-prod with mailstrom-prod

ryanjjung:
Also fix one last URL

### DIFF
--- a/pulumi/config.prod.yaml
+++ b/pulumi/config.prod.yaml
@@ -20,6 +20,16 @@ resources:
         - ecr.dkr
         - logs
         - secretsmanager
+      peering_connections:
+        mailstrom-prod:
+          peered_cidrs:
+            - 10.0.0.0/16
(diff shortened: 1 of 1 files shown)


MelissaAutumn (approved):

pulumi/config.prod.yaml
```diff
@@ -389,7 +399,7 @@ resources:
               - name: STALWART_BASE_JMAP_URL
                 value: 'https://mail.thundermail.com'
               - name: STALWART_BASE_API_URL
-                value: 'https://mail-i.thundermail.com'
+                value: 'https://mailstrom-management-i.thundermail.com'
```
MelissaAutumn, line 402: oops we need the 8080 port on these

MelissaAutumn (changes requested):

pulumi/config.prod.yaml
ryanjjung: Oh, we sure do. One sec...

PR #257, opened 2025-09-03
Fix browserstack jobs by updating playwright version

rwood-moz:
BrowserStack test jobs starting failing out of the blue (see #256) however updating the Playwright version locally and in BrowserStack appears to fix the issue


rwood-moz:
Tested running E2E tests on this branch against local stack as well as running a test in BrowserStack against stage (BrowserStack [link is here](https://automate.browserstack.com/dashboard/v2/builds/d8700741e1bf05dd409347eea9486fd63f32307c); note stage has an issue at the moment so test auth itself failed, however it did run).

PR #258, opened 2025-09-03
Lookup stalwart accounts by OIDC ID

### LINKED ISSUE #254: Lookup Stalwart accounts by oidc_id
Instead of looking up Stalwart accounts by the primary thundermail address we should look up the account by oidc_id (keycloak's uuid.)

Pairs with https://github.com/thunderbird/mailstrom/issues/88

MelissaAutumn:
Fixes #254 

A large refactor of how we call and reference stalwart accounts. 

## Confusing terms hopefully explained

Email and Username here refer to OIDC field names from the spec. Due to certain platform restrictions we have to be a little creative with what we put in which field.

`Email`:
* Stalwart config: `fields.email` is the primary email address used in Stalwart. Used in account look up for auth.
* Stalwart: The primary email address used in Stalwart.
* Keycloak: The backup email used for reset password and other system 
* Accounts: The backup email as defined in Keycloak. We don't really use this field right now, so you can ignore this.
 
`Username`: 
* Stalwart config: `fields.username` is the unique inbox (technically: principal object with type=individual) name and look up field. 
* Stalwart: Stalwart doesn't have the concept of a username afaik.
* Keycloak: The primary email address used for logging in.
* Accounts: The primary email address as defined in Keycloak, this isn't really used. Instead `User.Account.Emails.filter(type=primary)` is the way we match things (see below).

In accounts we additionally have an Account and Email object that, when a Stalwart inbox is created these objects are created and assigned to the user as well. Account references the Stalwart inbox (I'll reconcile these names in another ticket), and Email is an email address (which could be primary, or alias.) 

## So how does this all fit together?

During a successful login process with Keycloak the mail client receives and stores an access token. When the mail client wants to talk to Stalwart's IMAP, SMTP, or JMAP servers they give this access token to Stalwart. Stalwart then gives this token to Keycloak (via the token introspection endpoint) which if valid they'll receive some user information. 

Basically looks like this (slimmed down to just the fields we care about):
```
{
 'active': True,
 'client_id': '<oauth client used to login>',
 'email': 'my-backup-email-address@example.com',
 'sub': '<keycloak user uuid>',
 'username': 'my-cool-email-address@stage-thundermail.com',
  
 ... some other data too ...
}
```

Stalwart then checks the config.toml for the `fields.email` and `fields.username` options. (It also checks `fields.full-name`, but we don't need to set that.) These values need to match the keys in the json response from the token introspection call (see above.) 

Given the available options here, we're switching the config to `fields.email=username` and `fields.username=sub`. Which means Stalwart will look at Keycloak's `sub` field (the keycloak uuid) to set and uniquely look up the Stalwart inbox. It will also then use Keycloak's `username` field (our primary thundermail address) for the primary email look up.

## But why is email so weird?

Keycloak can only send service emails to the email property. If a user forgets their thundermail password, well a reset password email going to their thundermail address would be pretty pointless.

This will hopefully be fixed in the future, additionally I'm looking where I can remove unused references to email and username to help remove confusion in the system. (e.g. accounts.email and accounts.username.)


src/thunderbird_accounts/mail/client.py
```diff
@@ -199,23 +222,23 @@ def create_account(self, email, username, uuid='', app_password=None):
         # Return the pkid
         return data.get('data')
 
-    def get_account(self, username):
-        response = self._get_principal(username)
+    def get_account(self, oidc_id: str):
```
MelissaAutumn, line 225: Previously the inbox name (unique lookup field) was the primary email address (keycloak's username field.) That's no longer true as we're looking up keycloak's uuid. 

To keep things generic in case we switch away from keycloak it's called oidc_id.

src/thunderbird_accounts/mail/templates/admin_stalwart_view.html
```diff
@@ -29,5 +29,43 @@ <h1>Stalwart Principals</h1>
...
+    {% csrf_token %}
+    <div class="submit-row">
+      <a class="deletelink" id="purge" href="#">Purge individuals</a>
+    </div>
+    <pre id="output"></pre>
+  </div>
 </section>
+<script>
+  // Allows up to clean up the test data before the proper oidc schema has been finallized.
+  const btn = document.getElementById('purge');
```
MelissaAutumn, line 46: I'll need to clean up users on stage, and maybe prod if prod has any test data lying around.

MelissaAutumn: we'll then remove this purge call (and the purge function)

src/thunderbird_accounts/mail/views.py
```diff
@@ -83,26 +83,21 @@ def sign_up_submit(request: HttpRequest):
         pass
 
     user = request.user
+
+    # Set the user's username to the new primary email address and save the user
+    user.username = email_address
+    user.save()
```
MelissaAutumn, line 89: This route will probably be deprecated soon, but for the sake of completion I'm ensuring we're saving the brand new thundermail address as the user's username.

src/thunderbird_accounts/mail/views.py
```diff
@@ -448,6 +443,33 @@ def jmap_test_page(request: HttpRequest):
     )
 
 
+@require_http_methods(['POST'])
+@staff_member_required()
+def purge_stalwart_accounts(request: HttpRequest):
```
MelissaAutumn, line 448: We'll remove this at a later date once prod is up and running.

davinotdavid (approved):
LGTM!

Not 100% familiar with the Keycloak / OIDC / Stalwart flows so review is from the diff only!

Two tiny comments / suggestions

src/thunderbird_accounts/authentication/models.py
```diff
@@ -59,6 +59,10 @@ def has_active_subscription(self):
...
 
     @cached_property
-    def stalwart_username(self) -> str | None:
+    def stalwart_primary_email(self) -> str | None:
+        """Returns the primary email address used for Stalwart."""
         account = self.account_set.first()
-        return account.name if account else None
+        if account:
+            email = account.email_set.filter(type=account.email_set.model.EmailType.PRIMARY).first()
+            return email.address
```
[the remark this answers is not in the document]
MelissaAutumn: they're created in pairs, but i should really check for none anyways.

Will fix!

src/thunderbird_accounts/authentication/admin.py
```diff
@@ -17,15 +17,15 @@ def admin_create_stalwart_account(modeladmin, request, queryset):
     errors = 0
 
     for user in queryset:
-        account_not_found = not user.stalwart_username
+        account_not_found = user.account_set.length == 0
 
-        if user.stalwart_username:
+        if user.account_set.length == 0:
```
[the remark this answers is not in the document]
MelissaAutumn: thanks!

PR #267, opened 2025-09-10
Fix Stalwart's OIDC Lookup

### LINKED ISSUE #266: Revert Stalwart Account lookups to Thundermail Address
Turns out we cannot have the oidc_id as Stalwart's Account/Mail.Account's name field as that's the username field for app password logins.

So I need to revert most of #258

MelissaAutumn:
Fixes #266 

I'll have to run the new action on everyone in the admin panel, but we've reverted to using username (thundermail address) for both email and name. Which is confusing...but they can now use their thundermail address to login with app passwords, and our keycloak authentication also matches on primary thundermail address.

I also ruff formatted and cleaned up some code.


davinotdavid (approved):
LGTM! 🚀

PR #271, opened 2025-09-11
Zendesk Dynamic Form v2

### LINKED ISSUE #247: Make Zendesk custom fields dynamic / freely updatable through Zendesk admin
The current implementation of the Zendesk contact form is non-sufficient as we are currently hard-coding two specific fields to show in the form: `Product` and `Type of request`.

Currently:
- Changing the name of the field or adding / removing fields need to be paired with a code change
- Changing the _options_ within `Product` and `Type of request` is dynamic and does not require a code change

The ask is that there is a need to make it so that Zendesk admins can freely update any custom fields and have it reflected immediately / dynamically in our hosted Zendesk form.

A few technical challenges for whoever picks this up:
- The Zendesk API endpoint that we are currently using is `/ticket_fields` [[docs here](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_fields/#list-ticket-fields)]
  - This endpoint does not distinct between `custom_fields` in _different forms_ though. Apparently, each form may have any number of custom fields and the ask is to be able to get `custom_fields` from a specific given form
  - The returned fields don't discriminate on whether or not they should be visible to the end user so we perhaps we might need to add some allow list / deny list

- There is another Custom Fields API that we could potentially use [[docs here](https://developer.zendesk.com/api-reference/sales-crm/resources/custom-fields/)] but from the response fields, it also doesn't discriminate based on which form it belongs to
 
- There is an option in the Zendesk admin which should theoretically prevent the field from being exposed in the form ("Customers Can Edit"). The actual JSON field that maps to this (and where to get that from) is yet to be discovered:

Image

- i18n should be considered (TBD as currently the form is not internationalized)

Paraphrasing the exact message on how this should behave:

```
You should get all of the fields that I mark as user-editable and that are assigned to that Form. That's what I'd expect as a non-coder:

- Grab the form by ID
- Recurse through the fields, display ones that meet the criteria of belonging to that form and editable by the end user

No field hardcoding at all
```

davinotdavid:
## Description of change

[**Before**]

We were fetching for `custom_fields` specifically and hard-coding the ones we care about so that we can show them on a form. This worked well but it is problematic because any change done by the support team would require engineering work to update the form + a new deployment.

Also, using the endpoint we were using before, we didn't have the context of which Zendesk form these fields were in. The suggested approach was to fetch the form only, traverse the form field IDs in order and make a new request _per field_ to a different API to get the form field attributes (e.g. type, required, etc) which, even with caching, it is definitely not ideal.

[**Now**]
- Added a new `ZENDESK_FORM_ID` env var so that we can load dynamic form fields based off of a specific form.
- Added `ZENDESK_FORM_BROWSER_FIELD_ID` and `ZENDESK_FORM_OS_FIELD_ID` env vars to represent the hidden custom field ids from Zendesk.
- Changed the endpoint to `subdomain.zendesk.com/api/v2/ticket_forms/{ZENDESK_FORM_ID}?include=ticket_fields` which retrieves both the form and its ticket fields **in order** (this is important so that the field reordering is also offset to Zendesk admins).
-  Added a `TICKET_FIELD_TYPE_TO_COMPONENT` in the "frontend" so that we map [Zendesk's Form Field Types](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_fields/#create-ticket-field) into renderable Vue Components.
- Added [ua-parser](https://github.com/ua-parser/uap-python) and `regex` packages to parse the user agent string and send it to Zendesk.
- Still sending a `multipart-formdata` due to sending attachment files but the form fields are stringfied and sent in a `data` field of the form data.

### Zendesk-specific workarounds / adaptations

1. Weirdly, there are two special form fields: `subject` and `description`. Even though these are dynamically rendered, they need to be handled differently as they have to be filled in specific keys for the [Requests API](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket-requests/#json-format):
- `subject` needs to be sent in the `subject` key
- `description` needs to be sent in the `comment['body']` key and not in the `description` key, per docs (heh)

2. The only way to hide a field from the end user form is to set it as `Agents can edit`. However, if the field has that set it won't be updateable through the Requests API. It is important for the support team for the ticket _creation_ to be done through the [Requests API](https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#tickets-and-requests) so that it can have some metadata related to that. So in order to actually update hidden custom fields, we need to use the [Tickets API](https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#tickets-and-requests) using the agent's credentials to make a `PUT` on the newly created ticket.

## How to test it

In 1Password, look for the `Zendesk Sandbox: Design` to get the credentials needed for testing and env vars for your `.env`.

Helpful links:
- [Zendesk Form for Testing](https://tbpro1755174120.zendesk.com/admin/objects-rules/tickets/ticket-forms/edit/43835995509779) for manipulating the form fields and seeing the changes in the contact form
- [Zendesk Agent Tickets](https://tbpro1755174120.zendesk.com/agent/filters/43835995869715) for seeing your ticket submissions through the contact form and checking if fields are filled correctly

1. Navigate to `/contact`
2. Check that the form loads correctly and you can see the expected form fields according to the [Zendesk Form for Testing](https://tbpro1755174120.zendesk.com/admin/objects-rules/tickets/ticket-forms/edit/43835995509779)
3. Fill in the form and submit it
4. Check that the submission was done successfully and see the created ticket on the [Zendesk Agent Tickets](https://tbpro1755174120.zendesk.com/agent/filters/43835995869715) page
5. Go to the [Zendesk Form for Testing](https://tbpro1755174120.zendesk.com/admin/objects-rules/tickets/ticket-forms/edit/43835995509779) and make some changes (e.g. re-order the form fields, change the visibility of a field, add a new field, etc)
6. Repeat the steps (except that you will need to refresh the `/contact` page if you are already there)

> [!NOTE]  
> When this PR gets merged, it will only be possible to test in staging after adding the new `ZENDESK_FORM_ID`, `ZENDESK_FORM_BROWSER_FIELD_ID` and `ZENDESK_FORM_OS_FIELD_ID` .env var to Pulumi with the desired non-sandbox values (needs sync with the Support team).

## Screenshots

image

- We are not currently covering _all_ of the possible [Zendesk Form Field Types](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_fields/#create-ticket-field) so while this covers the three most common field types (Text, TextArea, Single Selects), it is currently not covering every single possibility but it should be trivial to extend this if needed / defined. However, this has been synced with Lisa that it is fine since we are not considering usage of other components in the foreseeable future.

## Solves
Fixes #247 
Fixes https://github.com/thunderbird/thunderbird-accounts/issues/383

### DIFF
--- a/assets/app/vue/views/ContactView/components/NativeInputWrapper.vue
+++ b/assets/app/vue/views/ContactView/components/NativeInputWrapper.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { TextInput } from '@thunderbirdops/services-ui';
+
 /**
  * This component is a simple wrapper for native HTML input elements.
  * It is used alongside the <component :is="" /> Vue component
@@ -23,7 +25,7 @@ function onInput(e: Event) {
 </script>
 
 <template>
-  <input
+  <text-input
     v-bind="$attrs"
     :type="type"
     :name="name"


assets/app/vue/components/UserMenu.vue
```diff
@@ -57,6 +58,14 @@ onBeforeUnmount(() => {
     <user-avatar :username="username" class="avatar" @click="toggleMenu" />
 
     <div v-if="showMenu" class="dropdown">
+      <router-link
+        v-for="internalItem in internalMenuItems"
```
MelissaAutumn: Maybe a comment above internal's array and external's array explaining that ones holds vuejs routes and the other holds django routes.

src/thunderbird_accounts/utils/utils.py
```diff
@@ -25,3 +26,29 @@ def dumps(self, obj):
 
     def loads(self, obj):
         return json.loads(obj)
+
+
+def parse_user_agent_info(user_agent_string: str) -> tuple[str, str]:
+    """Parse user agent string and return separate browser and OS information."""
+
+    # Extract browser information
+    browser_info = parse_user_agent(user_agent_string)
```
MelissaAutumn: Hmm instead try:
```python
    browser_string = browser_info.family
    browser_version = '.'.join(filter(lambda x: x, [
        browser_info.major,
        browser_info.minor,
        browser_info.patch,
    ]))
    if browser_version.strip():
        browser_string = f'{browser_string} {browser_version}'
```

MelissaAutumn: (and also for os)

assets/app/vue/components/UserMenu.vue
davinotdavid: Added, thanks!

assets/app/vue/views/ContactView/components/ContactSupportForm.vue
```diff
@@ -197,86 +206,44 @@ onMounted(() => {
...
-    <text-input
-      ref="emailInput"
-      name="email"
-      type="email"
-      v-model="form.email"
-      :required="true"
-      data-testid="contact-email-input"
-    >
-      Your email address
+    <!-- Email (always present) -->
```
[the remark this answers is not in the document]
davinotdavid: Asked design in [this Matrix question](https://matrix.to/#/!wniULDJJJnlCEQCKpK:mozilla.org/$Boloer76wCzTKwIfupAuMg9Q-0JHZanvX08FSCdphnE?via=mozilla.org&via=matrix.org&via=humanoids.be) and confirmed to show 5 `LoadingSkeletons` here. Added, thanks!

src/thunderbird_accounts/utils/utils.py
davinotdavid: Oh TIL, much better to read indeed. Changed, thanks!

PR #280, opened 2025-09-16
Enable scaling for keycloak

ryanjjung:
This is a self-referential rule. That is, we allow any entity the SG gets attached to to speak on port 7800 with any other entity the SG is attached to. This allows our Keycloak containers to speak their cluster protocol with each other.

### DIFF
--- a/pulumi/__main__.py
+++ b/pulumi/__main__.py
@@ -52,7 +52,8 @@ for service, sg in resources['tb:network:SecurityGroupWithRules']['containers'].
         raise ValueError(f'{MSG_LB_MATCHING_CONTAINER} Create a matching load_balancers entry for "{service}".')
     # Allow access from each load balancer to its respective container
     for rule in sg['rules']['ingress']:
-        rule['source_security_group_id'] = lb_sgs[service].resources['sg'].id
+        if 'self' not in rule or not rule['self']:
+            rule['source_security_group_id'] = lb_sgs[service].resources['sg'].id
     depends_on = [lb_sgs[service].resources['sg'], vpc] if lb_sgs[service] else []
 
     container_sgs[service] = tb_pulumi.network.SecurityGroupWithRules(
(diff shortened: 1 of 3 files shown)


pulumi/config.prod.yaml
```diff
@@ -109,6 +109,15 @@ resources:
...
               to_port: 9000
+            # KeyCloak clustered nodes speak to each other on port 7800
+            - description: Allow clustered nodes to communicate
+              protocol: tcp
+              from_port: 7800
+              to_port: 7800
+              # The following security group is self-referential. If you have to rebuild this env from scratch,
+              # you will have to uncomment this rule to do the initial build, then uncomment it and update this
+              # rule ID with the security group created by this config.
+              source_security_group_id: sg-0f703e22b05e6f0c3
```
MelissaAutumn, line 120: Oh as in this is the security group id for its own sg? 

Can we just have a custom property named self_reference_sg: true? 

We can punt that to a new ticket.

MelissaAutumn (approved):

pulumi/config.prod.yaml
ryanjjung: Hrm, lemme see, actually. There might be a special "self" option.

ryanjjung:
The `self` option works with a slight code tweak to detect it.

pulumi/__main__.py
```diff
@@ -52,7 +52,8 @@
         raise ValueError(f'{MSG_LB_MATCHING_CONTAINER} Create a matching load_balancers entry for "{service}".')
     # Allow access from each load balancer to its respective container
     for rule in sg['rules']['ingress']:
-        rule['source_security_group_id'] = lb_sgs[service].resources['sg'].id
+        if 'self' not in rule or not rule['self']:
+            rule['source_security_group_id'] = lb_sgs[service].resources['sg'].id
```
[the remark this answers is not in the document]
ryanjjung: Should be. All of these values are read in from a YAML file, so it should not be possible for this to get loaded into a class context.

PR #303, opened 2025-10-02
Enable keycloak autoscaling in prod

ryanjjung:
This is really a question for @MelissaAutumn to respond to:

Is there any reason we can't turn this on in prod? I didn't want to apply it without asking first because I wasn't sure if everything was fixed for this there.


MelissaAutumn (approved):
I don't think there's a problem if it works as well as it does in staging.

ryanjjung:
Yes, peering is on in prod. I was mostly worried about the usage of the built-in cache that gave us trouble before. But I think that's patched in the task def.

PR #324, opened 2025-10-20
Add monitoring to Accounts

ryanjjung:
This sets up the standard baseline metrics alerting and updates some images in stage to reflect reality and remove drift.


aatchison (approved):
Thanks for setting up this monitoring group!

PR #326, opened 2025-10-20
Add a route that will send a user to keycloak to update their password. (Fixes #193)

### LINKED ISSUE #193: Create a reset password route for password changes
## Details

Create a backend route that when called (via POST) will queue up a task that calls keycloak's reset password email api.

https://www.keycloak.org/docs-api/latest/rest-api/index.html#_put_adminrealmsrealmusersuser_idreset_password

This should obviously be logged in users only.

MelissaAutumn:
#193 

Since all credential actions need to take place in keycloak here's how we do it! We add a new endpoint that sends the user to keycloak via a app initiated action. The actual actions are defined as required actions in keycloak's admin panel, but I've copied them into an enum to freely use. 

Keycloak then handles the actual flow and sends the user to the redirect uri we send them. The actual flow on keycloak's end will need to be tweaked but it works well enough. I've added the login-username view (as a variant / prop of login view) since that oddly showed up in my testing. We'll sort that out later.

References:
https://www.keycloak.org/docs/latest/server_admin/index.html#con-aia_server_administration_guide
http://keycloak:8999/admin/master/console/#/master/providers (search for `required-action`)

### DIFF
--- a/assets/app/vue/views/DashboardView/components/AccountCard.vue
+++ b/assets/app/vue/views/DashboardView/components/AccountCard.vue
@@ -116,6 +116,7 @@ const { t } = useI18n();
 }
 
 .fake-button-link {
+  color: var(--colour-service-primary);
   display: flex;
   justify-content: center;
   align-items: center;
--- a/src/thunderbird_accounts/authentication/utils.py
+++ b/src/thunderbird_accounts/authentication/utils.py
@@ -55,5 +55,8 @@ def create_aia_url(action: KeycloakRequiredAction):
     redirect_uri = quote(get_absolute_url(reverse('login')))
     return urljoin(
         settings.KEYCLOAK_AIA_ENDPOINT,
-        f'?response_type=code&client_id={settings.OIDC_RP_CLIENT_ID}&kc_action={action.value}&redirect_uri={redirect_uri}',
+        f'?response_type=code'
+        f'&client_id={settings.OIDC_RP_CLIENT_ID}'
+        f'&kc_action={action.value}'
+        f'&redirect_uri={redirect_uri}',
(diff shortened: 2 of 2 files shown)


assets/app/vue/views/DashboardView/components/AccountCard.vue
```diff
@@ -113,4 +114,21 @@ const { t } = useI18n();
     min-width: 568px;
   }
 }
-</style>
\ No newline at end of file
+
+.fake-button-link {
```
MelissaAutumn: We should add link styling to services-ui so it mimics a link-button.

src/thunderbird_accounts/authentication/utils.py
```diff
@@ -7,3 +47,13 @@ def is_email_in_allow_list(email: str):
...
     return email.endswith(tuple(allow_list.split(',')))
+
+
+def create_aia_url(action: KeycloakRequiredAction):
+    """Create a url for a user to start a keycloak flow.
+    These flows are defined as actions in KeycloakRequiredAction."""
+    redirect_uri = quote(get_absolute_url(reverse('login')))
+    return urljoin(
+        settings.KEYCLOAK_AIA_ENDPOINT,
+        f'?response_type=code&client_id=tb-accounts&kc_action={action.value}&redirect_uri={redirect_uri}',
```
davinotdavid: As you've mentioned on Matrix, the client_id here shouldn't be hardcoded and changing to the following works:

```suggestion
        f'?response_type=code&client_id={settings.OIDC_RP_CLIENT_ID}&kc_action={action.value}&redirect_uri={redirect_uri}',
```

Issue #325, opened 2025-10-20
Navigating to OIDC / Keycloak's /logout sets Django's is_authenticated as False


davinotdavid:
## Describe your issue

As pointed by @rwood-moz [here](https://github.com/thunderbird/thunderbird-accounts/pull/318#issuecomment-3402567488), when the user navigates to `/logout` and lands on the Keycloak's logout confirmation page, it already sets the `user.is_authenticated` from Django auth middleware as `False`.

## How to reproduce

- Navigate to the new Dashboard / `http://localhost:8087`
- Click on the `Login` button on the top right
- Login
- Click on the `UserAvatar` on the top right and click on `Logout`
- DO NOT confirm – instead, click on `Back to application`
- Check that you are redirected to the new Dashboard, except that you are seemingly logged out
- Bonus: Clicking on Login auto-logs you in without prompting by user password, showing that the session still exists and it is valid

Issue #320, opened 2025-10-15
Sign-in page is extremely small on mobile browser


rwood-moz:
When signing into TB Accounts on a mobile browser the sign-in page and all of the elements are too small. PR #306 implemented hi-fi designs for login and there is a small login page however it is not being displayed when signing in on the mobile browser.

Here's a screenshot of signing into TB Accounts on a Google Pixel 10.

<img width="204" height="419" alt="Image" src="https://github.com/user-attachments/assets/01478554-2332-4176-8700-956488823ed2" />

davinotdavid:
@rwood-moz this should now be working on stage! Could you give it another try? Thanks!

PR #327, opened 2025-10-20
Fix / Create a 2-step OIDC logout flow + Misc fixes

davinotdavid:
## Description of changes

Turns out that the [mozilla-django-oidc](https://github.com/mozilla/mozilla-django-oidc)'s `OIDCLogoutView` that we were using when accessing the logout view actually logout the user from the Django perspective [[reference](https://github.com/mozilla/mozilla-django-oidc/blob/main/mozilla_django_oidc/views.py#L267-L268)].

This is not great as our current use with Keycloak has a confirmation screen for logout and the user might NOT confirm and go back to the application. So this PR introduces a "2-step" logout process where we instead use our own views to display the logout page and only actually logs out the user upon confirmation.

Some extra / tangent / slightly unrelated fixes:
- Updated subscription placeholder prices according to [Product docs](https://docs.google.com/document/d/1vQaED8C73RRvi8jHPOyqlqPc7h0okF9rZoOBh0hs5Z8/edit?tab=t.uk1i13nkgjgb#heading=h.adxexheztvfr)
- Fix missing locale for Login button label
- Make the user email real on the dashboard
- Adds a meta tag for mobile viewports to solve https://github.com/thunderbird/thunderbird-accounts/issues/320
- Miscellaneous lint fixes

Solves https://github.com/thunderbird/thunderbird-accounts/issues/325, solves https://github.com/thunderbird/thunderbird-accounts/issues/320

### DIFF
--- a/src/thunderbird_accounts/authentication/views.py
+++ b/src/thunderbird_accounts/authentication/views.py
@@ -38,6 +38,7 @@ def start_oidc_logout(request: HttpRequest):
     return HttpResponseRedirect(redirect_url)
 
 
+@login_required
 def oidc_logout_callback(request: HttpRequest):
     """Finalize logout locally after the user confirmed the logout."""
     django_logout(request)


src/thunderbird_accounts/authentication/views.py
```diff
@@ -1,11 +1,46 @@
...
+    logout_endpoint = f"{base_url}/protocol/openid-connect/logout"
+    redirect_url = (
+        f"{logout_endpoint}?client_id={settings.OIDC_RP_CLIENT_ID}"
+        f"&post_logout_redirect_uri={post_logout_redirect}"
+    )
+
+    return HttpResponseRedirect(redirect_url)
+
+
+def oidc_logout_callback(request: HttpRequest):
```
MelissaAutumn: Do we need a login_required on this route?

PR #328, opened 2025-10-21
Propagate Account edits to Stalwart and creating aliases by default. 

### LINKED ISSUE #323: Accounts doesn't create tb.pro aliases by default
New accounts should come with a tb.pro alias for the same username they have @ thundermail.com but this currently doesn't happen.

MelissaAutumn:
Fixes #323 

This change allows someone from the admin panel to create an alias and have it propagate to Stalwart. This change also automatically hooks up a new account with an aliases under all of the non-primary allowed domains.

This was a lot more work than I expected, oops.

## Screenshots
image

### DIFF
--- a/src/thunderbird_accounts/mail/admin/forms.py
+++ b/src/thunderbird_accounts/mail/admin/forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.validators import validate_email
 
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/mail/admin/__init__.py
```diff
@@ -0,0 +1,7 @@
+from django.contrib import admin
+
+from thunderbird_accounts.mail.admin.models import AccountAdmin
+from thunderbird_accounts.mail.models import Account
+
+# Register your models here.
+admin.site.register(Account, AccountAdmin)
```
MelissaAutumn: This file replaces admin.py, as you can't have a module (folder with __init__.py) and a python script with the same name.

src/thunderbird_accounts/mail/admin/forms.py
```diff
@@ -0,0 +1,70 @@
...
+
+class CustomAccountBaseForm(forms.ModelForm):
+    """Base class to use for custom email admin forms."""
+
+    class Meta:
+        model = Account
+        fields = '__all__'
+
+    def clean(self):
+        # TODO: Check for active stalwart / active subscription account here!
```
MelissaAutumn: Once we implement sign up flow...

MelissaAutumn:
Yeah that dropdown is entirely accounts only right now, changing something to a primary account is a bit more work as we'd need change the account name and re-sync a bunch of things. 

I'll add some verification on the clean method to prevent more than one primary. 

Thanks!

PR #337, opened 2025-10-23
SignUp Flow

### LINKED ISSUE #334: Create Sign-Up flow
This includes being invited to create an account (an account is setup in the admin panel and an email is sent to the end user to update their password), logging into accounts, going through the paddle payment flow, and landing on a dashboard.

MelissaAutumn:
Fixes #334 

But now you can enter Accounts and it will force you to subscribe. You can use a test card (instructions https://developer.paddle.com/concepts/payment-methods/credit-debit-card#test-payment-method) to create a subscription. 

Behind the scenes Paddle will send us a webhook, but that's not possible locally so I have to poke their wonderful webhook api to retrieve the transaction webhook and subscription webhook and run them manually. 

Atm this doesn't update Keycloak, and this doesn't prevent non-subscribed users from sending api requests lol. 

I had to cut this PR down as it would end up monstrous in size. Additional tickets + PRs will be coming within the week.


assets/app/vue/views/SubscribeView/components/CheckoutStep.vue
```diff
@@ -0,0 +1,373 @@
...
+            <p class="summary__cost-amount">
+              <loading-skeleton :is-loading="paddleLoading">
+                {{ order_summary.dueOnRenewal }}
+              </loading-skeleton>
+            </p>
+          </li>
+        </ul>
+      </card-container>
+      <card-container class="checkout-container">
+        <h2>Payment</h2>
```
MelissaAutumn: Paddle appends the payment form to this container.

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -48,7 +47,6 @@ def create_user(self, claims):
 
         # Refresh so we have access to the uuid
         user.refresh_from_db()
-        create_stalwart_account(user)
```
MelissaAutumn: Accounts are no longer created on login, instead they're moved to after the payment.

src/thunderbird_accounts/mail/clients.py
```diff
@@ -362,6 +369,20 @@ def update_individual(
...
+        ]
+        response = self._update_principal(principal_id, update_data)
+
+        # Returns data: null on success...
+        data = response.json()
+        error = data.get('error')
+        # I have no idea what the error is yet
+        if error:
+            logging.error(f'[update_individual] err: {data}')
+            raise RuntimeError(data)
```
MelissaAutumn: I'll cut a ticket to actually test these...This part is mostly copy paste from other update functions.

src/thunderbird_accounts/mail/migrations/0007_account_quota.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('mail', '0006_account_stalwart_created_at_account_stalwart_id_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='account',
+            name='quota',
+            field=models.BigIntegerField(help_text='Amount of mail storage this account has access to (in bytes).', null=True),
```
MelissaAutumn: Stalwart stores this in bytes, so we'll mirror it.

src/thunderbird_accounts/mail/models.py
```diff
@@ -60,6 +60,10 @@ class Account(BaseStalwartObject):
     name = SmallTextField(unique=True, help_text=_('The account name (this must be unique.)'))
     active = models.BooleanField(default=True)
     user = models.ForeignKey(User, on_delete=models.CASCADE, null=True)
+    quota = models.BigIntegerField(
+        null=True, help_text=_('Amount of mail storage this account has access to (in bytes).')
+    )
+    # TODO: Implement freeze_quota
```
MelissaAutumn: Incoming ticket for that

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -177,27 +206,58 @@ def create_stalwart_account(
         ],
     ]
 
+    _stalwart_check_or_create_domain_entry(stalwart, domain)
+    for alias in emails[1:]:
+        _domain = alias.split('@')[1]
+        print(f'checking domain {_domain}')
+        _stalwart_check_or_create_domain_entry(stalwart, _domain)
```
MelissaAutumn: Ran into an issue where alias domains weren't added. So this "fixes it" (It kinda sucks eh?)

src/thunderbird_accounts/subscription/utils.py
```diff
@@ -1,17 +1,30 @@
...
-    """Updates a user's mail accounts quota."""
-    accounts = user.account_set.all()
-    for account in accounts:
-        account.quota = calculate_quota_in_bytes(plan.mail_storage_gb)
-        account.save()
+def deactivate_subscription_features():
+    """
+    TODO: Implement me
+    """
+    pass
```
MelissaAutumn: Incoming ticket for this

src/thunderbird_accounts/subscription/views.py
```diff
@@ -23,6 +35,96 @@ def prefilter_paddle_webhook(event_type: str, event_data: dict) -> bool:
...
+                # The serialization trick is not perfect, we need to flatten custom_data
+                notification_data['custom_data'] = notification_data.get('custom_data', {}).get('data', {})
+
+                if notification_data.get('id') == subscription_id and notification.type == 'subscription.created':
+                    tasks.paddle_subscription_event.run(
+                        notification_data,
+                        notification.occurred_at,
+                        is_create_event=True,
+                    )
+                    break
```
MelissaAutumn: Notifications are just the webhook record / payload accessible via api. Very handy, because this would otherwise be super annoying to do.

vite.config.mts
```diff
@@ -35,6 +35,7 @@ export default defineConfig(({mode}) => {
       extensions: ['.ts', '.js', '.vue'],
     },
     server: {
+      hmr: false,
```
MelissaAutumn: hmr was freezing tabs while I was working on route guarding. Hope folks don't mind it being off until we have time to figure that out.

davinotdavid (commented):
Nice! That's a lot of work, thanks for doing this 🙏 

Before zooming in the code itself, there are a few things that I ran into when running this locally:

- The celery container doesn't load because there are two self-serve methods that were removed but the paths weren't. Removing these and deleting all containers and rebuilding everything loads the app well though!
  - `path('self-serve/app-passwords/set', mail_views.self_serve_app_password_set, name='app_password_set')`
  - `path('self-serve/display-name/set', mail_views.self_serve_display_name_set, name='display_name_set')`
 
-  I was able to go through the sign-up process _once_ successfully but I can't create new accounts anymore somehow as I run into this error:

image

This happened after confirming the email link and attempting to log in (before landing in the Order Summary page)

- This can be punt into another ticket probably but In the Order Summary page we are already logged in and showing all the navigation items. This is perhaps a UX issue where the user can try to navigate away to other pages without completing the Order Summary (also, not sure how one could get back here)

src/thunderbird_accounts/authentication/middleware.py
davinotdavid: That makes sense! Although the `mail/views.py` file on the `home` method seems to need a tweak (my bad on that one) so that it doesn't break the app like the review screenshot I shared

MelissaAutumn:
I think that was a bad merge, give me a sec!

src/thunderbird_accounts/authentication/models.py
```diff
@@ -53,6 +51,8 @@ class UserLanguageType(models.TextChoices):
         _('Test Account'), default=False, help_text=_('Whether this account is used for testing.')
     )
 
+    plan = models.OneToOneField('subscription.Plan', null=True, on_delete=models.SET_NULL)
```
davinotdavid: Unsure if I should mark it here or in the `0010_user_plan.py` migration but apparently the `plan_id` seems to be unique and trying to create a subscription for a new user with the same plan (we only have one now, right?) gives the `duplicate key value violates unique constraint "authentication_user_plan_id_key"`.

image

davinotdavid (commented):
For the first user, everything works well!

For the second user, right after the payment confirmation and before the redirect, it gives me a 500 and leaves the `authenticate_user` record in a weird state (I can use the Dashboard and I am logged in but there's no `plan_id` attached to it)

Issue #207, opened 2025-07-21
Create "Email Addresses" card


MelissaAutumn:
## Details

Create a card that shows the user's primary email address, and allows a user to add an email alias.

Note: We are not allowing the user to change their primary email at this time. 

## Zeplin
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/685f190983683b9d1478b4c7
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/685f1af444015d3c3d036aef

## Screenshot

<img width="675" height="424" alt="Image" src="https://github.com/user-attachments/assets/d0e280e5-af2a-4a88-a15e-69c575c75d85" />

<img width="730" height="700" alt="Image" src="https://github.com/user-attachments/assets/eca0bfdf-aea4-454b-84b5-925e5b109740" />

PR #341, opened 2025-10-24
Thundermail Email Aliases (UI only) + Small CSS fixes

davinotdavid:
## Description of changes
- Refactored / Broke down the `EmailSettingsSection` into separate components for better maintainability
- Added Email Aliases list + form related components
- Fixed HeaderBar and FooterBar links to `dashboard` since we always want to scroll to the top of the page instead of the section with the id="dashboard"
- Fixed the `DetailsSummary` component to have all borders with radius when closed and the chevron icon now rotates (without transitions yet).

Zeplin reference:
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/685f1af444015d3c3d036aef

> [!NOTE]  
> We don't have a hi-fi mockup for this part so this was based off of the wires in the link above + the pre-existing hi-fi behaviour of other pages. The UX from the video below been confirmed by the design team in this [Matrix message](https://matrix.to/#/!wniULDJJJnlCEQCKpK:mozilla.org/$8IXWg63tGvz0LnVKiJoGLRpTlnWmw-MdsDIO8P8nvCQ?via=mozilla.org&via=matrix.org&via=humanoids.be).

## Screenshots / Video

https://github.com/user-attachments/assets/5a50a4e2-99d9-4594-83f8-a5b78413b44c

- This is UI only so even though some behaviour seems real (make primary / delete / adding email aliases) the data shown is a hardcoded list of email aliases. The actual functionality will come in a separate PR!

## Related issues
https://github.com/thunderbird/thunderbird-accounts/issues/207

### DIFF
--- a/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
+++ b/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
@@ -37,6 +37,14 @@ const onMakePrimary = (alias: EmailAlias) => {
 
 const onDeleteAlias = (alias: EmailAlias) => {
   placeholderAliases.value = placeholderAliases.value.filter(item => item.id !== alias.id);
+
+  // Make the subscription alias the primary one if there is no primary alias
+  if (placeholderAliases.value.every(item => !item.isPrimary)) {
+    placeholderAliases.value = placeholderAliases.value.map(item => ({
+      ...item,
+      isPrimary: item.isSubscription
+    }));
+  }
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


rwood-moz (commented):
Tried it out locally, wow looks awesome.

I notice that you can make a different email your `primary` email, and then delete it even though it is set as your primary email - and then you don't have any primary email. Should that be allowed? Or is that something that will be added once the functionality is added to this new UI?

davinotdavid:
@rwood-moz thanks for the review and great points! Most of those can be solves when hooking up the real data but for completeness here's what I've changed in the latest commits:

- Deleting a `Primary` email moves the `Primary` to the `Subscription` (since the Subscription can't be deleted). I need to double check later if that's the design decision or if they prefer to not allow the deletion of Primary emails 🤔 
- Added custom validations to the email alias text input to follow what is written in the help text
- Hide the `Add email alias` when the alias limit is reached

Could you check again? 🙇

Issue #338, opened 2025-10-23
Tie a custom domain entry to a user


MelissaAutumn:
If a user creates and verifies a custom domain we should add a Domain model and attach it to the user (or account? I think user would be better.)

PR #343, opened 2025-10-27
Custom Domains functionality

davinotdavid:
## Description of changes
- Creates a Domain model with a User relationship + migration
- POST create custom domains in Stalwart
  - Verifies domain name existence
  - Checks if it passes `MAX_CUSTOM_DOMAINS` (currently in settings.py)
- GET for the custom domain DNS records and shows them in the specific step
  - This is being relayed as is from Stalwart but it seems that we will need to make changes to it soon (?)
- POST to verify the custom domain (_more in the TODO list_)
  - Calls the Troubleshooting API twice (once to [get a token](https://stalw.art/docs/api/management/endpoints#obtain-a-troubleshooting-token) and once to [actually verify](https://stalw.art/docs/api/management/endpoints#run-delivery-troubleshooting)), performs and collects [each verification step](https://github.com/stalwartlabs/webadmin/blob/main/src/pages/manage/troubleshoot.rs#L806-L1003) and returns it to the user for now (_more in the TODO list_)
  - Sets `verified_at` upon success
  - Sets `last_verification_attempt` regardless of the result
- DELETE to delete the custom domain

- DNS records are always verified successfully as long as the request to Stalwart is successful even though they are not actually verified successfully internally. There is still some questions around which DNS records to use / show but for now the "CRUD" for custom domains will unblock the "CRUD" for the email aliases. I am returning the verification.
- Also, we will need a job / task to verify the DNS records once the point above is done.
- Custom Domain DNS records are only shown _once_ as part of the creation process. I am currently asking Design if we can add an option somewhere to display them again _after_ the custom domain creation for reference.
- ~~Using POST to delete the custom domain is a little weird as we have the DELETE. However, this is also how the app passwords were handled in the `self-serve/`. Should we be more RESTful in our routes?~~
- Tests :(

## Solves
https://github.com/thunderbird/thunderbird-accounts/issues/338

### DIFF
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/utils.ts
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/utils.ts
@@ -73,7 +73,7 @@ export const generateDNSRecords = async (domainName: string) => {
     {
       type: 'TXT',
       name: domainName,
-      content: 'v=spf1 ra=postmaster -all',
+      content: `v=spf1 include:spf.${dnsHostname} -all`,
       priority: '-',
     },
     {


assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
```diff
@@ -20,16 +31,27 @@ const toggleMenu = () => {
...
-  // TODO: Implement delete functionality
-  console.log('Delete clicked', props.domain);
-  showMenu.value = false;
+const handleDelete = async () => {
+  const data = await removeCustomDomain(props.domain.name);
+  if (data.success) {
+    emit('custom-domain-removed', props.domain.name);
+    showMenu.value = false;
+  } else {
+    console.error(data.error);
```
MelissaAutumn: Can we show a page notice bar with the error?

assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const getDNSRecords = async (domainName: string) => {
+  const response = await fetch(`/custom-domains/dns-records?domain-name=${domainName}`);
+  return await response.json();
+};
+
+export const removeCustomDomain = async (domainName: string) => {
+  const response = await fetch(`/custom-domains/remove`, {
+    method: 'POST',
```
MelissaAutumn: Yeah we can swap this to DELETE.

MelissaAutumn (commented):
Neat, I'll pull this down and give it a good test in the am. Thank you!

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
davinotdavid: Ah yes, goes back to the patterns of showing error when there's not a super clear indication. I agree with adding a notice bar for now!

src/thunderbird_accounts/settings.py
```diff
@@ -410,6 +410,8 @@ def oidc_logout(request):
     ALLOWED_EMAIL_DOMAINS[0] if ALLOWED_EMAIL_DOMAINS and len(ALLOWED_EMAIL_DOMAINS) > 0 else 'example.org'
 )
 
+# Maximum number of custom domains a user can have
+MAX_CUSTOM_DOMAINS = 3
```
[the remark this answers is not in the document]
MelissaAutumn: For sign up flow pr, you'll want to create a single plan model in the admin panel and attach it to a paddle object. (They should be pulled in on container boot). 

Then you should be able to do user.plan.domains or whatever the attribute is.

MelissaAutumn:
@davinotdavid sorry, could you rebase main?

src/thunderbird_accounts/mail/views.py
```diff
@@ -64,10 +66,25 @@ def home(request: HttpRequest):
...
+        custom_domains = [
+            {
+                'name': domain.name,
+                'status': domain.status,
+            }
+            for domain in domains
+        ]
+
+        # Get user's plan info constraints
+        max_custom_domains = request.user.plan.mail_domain_count
```
MelissaAutumn: breaks for unauthed users 👀

assets/app/vue/views/MailView/sections/CustomDomainsSection/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+      type: 'SRV',
+      name: `_imaps._tcp.${domainName}.`,
+      content: `0 1 993 ${dnsHostname}`,
+      priority: '-',
+    },
+    {
+      type: 'SRV',
+      name: `_submission._tcp.${domainName}.`,
+      content: `0 1 587 ${dnsHostname}`,
+      priority: '-',
```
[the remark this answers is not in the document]
MelissaAutumn: ...etc 😄

MelissaAutumn (commented):
Sorry for the review snipe here.

davinotdavid:
@MelissaAutumn thank you SO much for the help with this + fast / snipe reviews, it is super helpful as there were (are?) still lots of unknowns for me in this task 😅 Applied the review points!

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
```diff
@@ -54,7 +83,7 @@ onBeforeUnmount(() => {
     </button>
 
     <div v-if="showMenu" class="dropdown">
-      <button @click="handleRetry">{{ t('views.mail.sections.customDomains.retry') }}</button>
+      <button @click="handleRetry" v-if="props.domain.status !== DOMAIN_STATUS.VERIFIED">{{ t('views.mail.sections.customDomains.retry') }}</button>
       <button @click="handleDelete">{{ t('views.mail.sections.customDomains.delete') }}</button>
```
MelissaAutumn: I'm not sure the best place to put this, but if you delete a domain while in the verifying step, you'll be stuck on that step until you refresh the page.

MelissaAutumn (commented):
Looks good, just two things.

.env.example
```diff
@@ -50,6 +50,8 @@ JMAP_HOST=localhost
 JMAP_PORT=443
 JMAP_TLS=True
 
+DNS_HOSTNAME=localhost
```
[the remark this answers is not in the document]
davinotdavid: Gotcha, makes sense. Removed `DNS_HOSTNAME` and using the connection info's smtp host for now. Thanks!

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
davinotdavid: Hmmm I am not sure about what to do there flow-wise. There's currently no way to see the DNS records again unless you are in that verify step so I've purposely added some logic to check if the domain you are removing is the one you are currently seeing the DNS records for (go back to the initial step if so or keep the verify step for the DNS records)

Maybe we could add a "back" / "cancel" button to the verify step?

MelissaAutumn: We should always have a way to access dns records for a given domain. That should be a big lift right?

MelissaAutumn:
Regarding dns steps I'm having issues getting the EHLO step working. But if you configure your stalwart (localhost:8080) hostname (http://stalwart:8080/settings/network/edit) to localhost, restart stalwart and setup a remote dns it partially works now (troubleshooting can be accessed http://stalwart:8080/manage/troubleshoot/delivery, make sure to click the button not press enter.)

Ah there localhost isn't valid, but I can disable it and get further in. I think once we merge this on stage we can successfully get what we need.

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
davinotdavid: Hmm not a super big lift no but apparently there's a few UI changes that were applied today (copy + action menu decisions):

https://zpl.io/OXoAXkk

There is going to be a "(Re-)verify Domain" option in the kebab menu

Sancus:
For spf, lets do `v=spf1 include:spf.thundermail.com -all`

Issue #344, opened 2025-10-28
Accounts Dashboard Mitigation Adjustments


davinotdavid:
## TODO list
- Hide any unimplemented UI elements
- Hide “Manage MFA” tab
- “Change Payment Method” button should link to correct Paddle page
- Ensure the Thundermail/Send/Appointment icons go to their respective websites
- Update pricing information
- Update support article links

PR #345, opened 2025-10-28
Account Dashboard Mitigation

davinotdavid:
## Description of changes
- Hides "Manage MFA" from the navigation links
- Hides MFA related content from the Dashboard page
- Hides "Manage devices" from the Dashboard page
- Implements the subscription plan information with real data
- Implements the "Manage Subscription" to point to the related Paddle page
- Makes Appointment / Send links real

## Screenshots
image

- Since the subscription plan information is real, we might need to update the Plan inside of the Paddle Dashboard / Django Admin to match the design copy (as in, change the name to Early Bird, etc).
- This PR doesn't hide / remove the delete account part yet as there is a tentative implementation here https://github.com/thunderbird/thunderbird-accounts/pull/365

## Related issues
Solves https://github.com/thunderbird/thunderbird-accounts/issues/344
Solves https://github.com/thunderbird/thunderbird-accounts/issues/368

### DIFF
--- a/assets/app/vue/composables/i18n.ts
+++ b/assets/app/vue/composables/i18n.ts
@@ -19,6 +19,17 @@ const datetimeFormats = {
   },
 };
 
+const numberFormats = {
+  en: {
+    currency: {
+      style: 'currency' as const,
+      currency: 'USD',
+      minimumFractionDigits: 0,
+      maximumFractionDigits: 2,
+    },
+  },
+};
+
 const instance = createI18n({
   legacy: false,
   globalInjection: true,
@@ -26,6 +37,7 @@ const instance = createI18n({
   fallbackLocale: 'en',
   messages,
   datetimeFormats,
+  numberFormats,
 });
 
 export default instance;
--- a/assets/app/vue/views/DashboardView/utils.ts
+++ b/assets/app/vue/views/DashboardView/utils.ts
@@ -1,16 +1,6 @@
 import type { ComposerNumberFormatting, ComposerTranslation, ComposerDateTimeFormatting } from 'vue-i18n';
 import type { SubscriptionData } from './types';
 
-export const getCurrencySymbol = (currency: string, n: ComposerNumberFormatting) => {
-  const formatted = n(0, {
-    style: 'currency',
-    currency,
-    currencyDisplay: 'narrowSymbol',
-  });
-
-  return formatted.replace(/[\d\s.,]/g, '');
-}
-
 /**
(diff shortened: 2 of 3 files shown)


assets/app/vue/views/DashboardView/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Formats subscription data from the backend into a display-friendly format
+ * - Converts currency code to symbol
+ * - Converts price from cents to dollars
+ * - Converts yearly pricing to monthly equivalent
+ * - Converts storage bytes to GB
+ * - Removes unnecessary decimal places
+ */
+export const formatSubscriptionData = (
```
MelissaAutumn: Can you add context for the single letter variables in the docblock? Mention they should be filled with vue-i18n's n,t,d vars.

MelissaAutumn (commented):
Few nits. We probably should leave number / currency formatting to Intl.NumberFormat or vue-i18n if that handles it the same.

src/thunderbird_accounts/subscription/views.py
```diff
@@ -167,3 +167,51 @@ def handle_paddle_webhook(request: Request):
...
+
+    # Get the subscription item (contains the product and price information)
+    subscription_item = subscription.subscriptionitem_set.first()
+
+    if not subscription_item or not subscription_item.product or not subscription_item.price:
+        return JsonResponse({'success': False, 'error': 'Subscription information incomplete'}, status=500)
+
+    # Get the plan associated with the product
+    try:
+        plan = Plan.objects.get(product=subscription_item.product)
```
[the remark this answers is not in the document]
davinotdavid: Nice! Makes sense, changed in the latest commit, thanks!

assets/app/vue/views/DashboardView/utils.ts
davinotdavid: Definitely! Added in latest commit, thanks!

assets/app/vue/composables/i18n.ts
```diff
@@ -9,12 +9,35 @@ const messages = {
...
+  en: {
+    long: {
+      year: 'numeric' as const,
+      month: 'long' as const,
+      day: 'numeric' as const,
+    },
+  },
+};
+
+const numberFormats = {
```
MelissaAutumn: Just a note #350 will probably eat all of this up.

MelissaAutumn (approved):

PR #362, opened 2025-10-30
Pulumi updates

### LINKED ISSUE #354: [stage] Can't access private stalwart instance
Stage can't access mail-i, and thus can't do any api things.
https://thunderbird.sentry.io/issues/6982822422/?project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

MelissaAutumn:
I noticed Keycloak takes like 4-5 minutes to boot up to a working state. 10 might be too big of a grace period, but it's at least working. 

Feel free to suggest smaller ones. This is on prod, so no rush.

Fixes #354

### DIFF
--- a/pulumi/config.prod.yaml
+++ b/pulumi/config.prod.yaml
@@ -258,7 +258,7 @@ resources:
       desired_count: 1
       ecr_resources:
         - arn:aws:ecr:eu-central-1:768512802988:repository/thunderbird/accounts*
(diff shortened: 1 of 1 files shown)


MelissaAutumn:
Note: I think @ryanjjung is taking over this PR for the prod infra work.

ryanjjung:
That is true! I'll be adding a peering connection change in the long run after the mailstrom-prod environment is back online later today.

Issue #348, opened 2025-10-28
Handle Subscription Deactivations


MelissaAutumn:
Related to #176 but there's a part of #337 wrt deactivating a subscription and what that entails that needs filling in when we receive a cancel request from paddle. 

We probably want to see if we can deactivate an account on Stalwart's end, otherwise we might be able to remove permissions from that Stalwart account in order to put it in stasis for x number of days before removal.

Issue #316, opened 2025-10-10
Pausing or cancelling a subscription causes webhook error


kewisch:
When you manually pause a subscription in paddle, the webhook event fails with `AttributeError: 'NoneType' object has no attribute 'get'`

https://thunderbird.sentry.io/issues/6935887542/?referrer=alert_email&alert_type=email&alert_timestamp=1760089351639&alert_rule_id=15849274&notification_uuid=8d6e5af6-7439-416c-a280-2ee4bc711114&environment=stage

https://github.com/thunderbird/thunderbird-accounts/blob/e1fc04a8340895f4e6d48ab464bcc452e01e5575/src/thunderbird_accounts/subscription/tasks.py#L188

I'm not 100% sure if this was the "Pause" event, or me immediately cancelling a subscription via paddle UX.

PR #364, opened 2025-10-31
Fix paddle's cancel event on subscription update, and add a test case for it.

MelissaAutumn:
This is part of #348 

Subscription updates without billing times were breaking the update code. Make that optional and I've added a test with a slimmed down version of the webhook call.


davinotdavid (approved):
LGTM 🚀 

Also, I believe this solves https://github.com/thunderbird/thunderbird-accounts/issues/316

PR #369, opened 2025-10-31
Implement Verify Custom Domain + DKIM records

### LINKED ISSUE #366: Implement Verify Custom Domain functionality
As part of the [Custom Domain functionality PR](https://github.com/thunderbird/thunderbird-accounts/pull/343), we were not entirely sure what an actual verification process look like since we were not able to do it for localhost.

There were a few leftover items for this issue:
- DKIM records should be retrieved from Stalwart and presented to the user alongside the frontend generated ones
- A correct verification should pass the parts of the troubleshooting process that are critical
- A failed verification should present useful error messages to the user related to it

davinotdavid:
## Description of changes
- Added a GET to `/custom-domains/dns-records` to fetch the DNS records that come from Stalwart. This is necessary since we don't generate the DKIM records (`_mta.sts` is always the same per [Matrix conversation](https://matrix.to/#/!RtubSrsIayrghrJkrD:mozilla.org/$LNUWIya-dga2MfJt8ihQ1pjusLW3H_IYjAIrvGPWjLM?via=mozilla.org&via=matrix.org&via=4d2.org))
  - On the frontend, after creating the custom domain and after clicking on "View DNS Records", we make a request to this endpoint and extract only the DKIM record(s) and merge them with the list of frontend generated DNS records.

- Updated the `/custom-domains/verify` endpoint to actually parse through Stalwart's Troubleshooting API and access whether or not we should define the custom domain as verified or not.
  - For localhost / development purposes, I am always verifying successfully.
  - For non-localhost / stage / prod purposes, it is parsing through the [multi-step verification](https://github.com/stalwartlabs/webadmin/blob/main/src/pages/manage/troubleshoot.rs#L806-L1003) and sorting the results in two buckets: warnings and critical issues. For more details, [please check this part of the changes](https://github.com/thunderbird/thunderbird-accounts/pull/369/files#diff-4275e5cae1fb5775414e5d3d9807a7bc14c5f22c9699386cc6a156b6b7ea455fR208-R231).

- Other fixes include:
  - Wrapped the Add Custom Domain input in a form so that hitting "enter" works
  - Added notice-bars all around to show warnings and critical errors separately

## Screenshots

When verifying from the kebab menu
image

When verifying from the "verify step" of adding a custom domain
Screenshot 2025-10-31 at 3 25 50 PM

## Related issues
Fixes  #366

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -161,23 +161,11 @@
           "verify": "Verify",
           "reVerify": "Re-verify",
           "domainAlreadyExists": "Domain already exists",
-          "verificationWarnings": {
-            "mtaStsNotFound": "MTA-STS policy not published (optional security feature)",
-            "tlsRptNotFound": "TLS-RPT record not published (optional reporting feature)",
-            "tlsaNotFound": "TLSA/DANE records not found (optional security feature)",
-            "mtaStsFetchError": "Failed to fetch MTA-STS policy",
-            "mtaStsVerifyError": "MTA-STS verification failed",
-            "tlsRptLookupError": "TLS-RPT lookup failed",
-            "tlsaLookupError": "TLSA lookup failed",
-            "daneVerifyError": "DANE verification failed",
-            "startTlsError": "TLS upgrade failed (degraded security but may work)"
-          },
           "verificationCriticalErrors": {
             "mxLookupError": "Unable to find mail servers for domain",
             "ipLookupError": "Unable to resolve IP addresses",
             "connectionError": "Failed to connect to mail server",
-            "readGreetingError": "SMTP greeting failed",
-            "ehloError": "EHLO command failed"
+            "readGreetingError": "We could not verify that the domain records connect to Thundermail"
           }
         },
         "securitySettings": {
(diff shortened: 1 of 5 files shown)


assets/app/vue/locales/en.json
```diff
@@ -160,7 +160,25 @@
...
+            "daneVerifyError": "DANE verification failed",
+            "startTlsError": "TLS upgrade failed (degraded security but may work)"
+          },
+          "verificationCriticalErrors": {
+            "mxLookupError": "Unable to find mail servers for domain",
+            "ipLookupError": "Unable to resolve IP addresses",
+            "connectionError": "Failed to connect to mail server",
+            "readGreetingError": "SMTP greeting failed",
+            "ehloError": "EHLO command failed"
+          }
```
davinotdavid: The copy can be tweaked of course but I wasn't too sure on what would be the most helpful message for each specific verification step warning / error so please advise if changes are needed.

Sancus: One thing that concerns me about this is just confusing and spamming people with all the warnings. Fastmail doesn't check any of these types of optional records as far as I can tell, they only do SPF, DKIM, and MX records.

That said I don't think changes should block this pull request, it's something we can iterate on.

MelissaAutumn: EHLO is just a self connection test on SMTP afaik. We can probably adjust that to "We could not verify that the domain records connect to Thundermail".  (And bundle readGreetingError with it.)

davinotdavid: Sounds good. I've removed the warnings from the frontend (they are still being returned from the backend / can be seen in Network tab) and mashed ehlo and readGreetingError together. Thanks!

Issue #370, opened 2025-11-03
Mail / Thundermail Mitigation


davinotdavid:
Wishlist / TODO:
- Make the storage usage slider real
- Implement Security Settings' Account Activity
- Implement Security Settings' Allow List

Mitigation:
- Hide Security Settings entirely

davinotdavid:
Closed per https://github.com/thunderbird/thunderbird-accounts/pull/379

PR #378, opened 2025-11-04
Implement Account Activity

davinotdavid:
## Description of changes
- Added two endpoints and connected them to the frontend (with some minor formatting for display purposes):
  - GET `api/v1/auth/get-active-sessions/` for getting the active sessions
  - POST `api/v1/auth/sign-out-session/` for logging out from a session
    - This deletes the Keycloak session and, in case you are logging yourself out, it also deletes the Django session
- Added two OIDC settings for the deletion to work:
  - `OIDC_STORE_ID_TOKEN = True`, needed to store the ID token in the session for session verification ([ref](https://mozilla-django-oidc.readthedocs.io/en/stable/settings.html#OIDC_STORE_ID_TOKEN))
  - `OIDC_USE_NONCE = False`, needed for .verify_token() check on the OIDC backend for session verification (OIDC_USE_NONCE)

## Screenshots
image

## How to test
- Login with two different browsers (or a normal browser + a private window), the list is desc order so if you do it in the same minute, the top one will be the second device.
- If you sign out from the session that you are not currently on, the page should refresh and you should see the entry disappear
- If you sign out from the session that you are currently on, you should be redirected to the home page without being logged in

- Apparently Keycloak doesn't store device information on a session but this might be achievable using a plugin of sorts so for now we are showing "Unknown Device" for all.
- It is easier to refresh the page for now so that data is automatically re-fetched and redirect to non-logged in / non-authenticated route works automatically. However, can definitely be enhanced for a more SPA feel in the (near) future.

## Related issues
https://github.com/thunderbird/thunderbird-accounts/issues/370


markstos:
Closing as superceded by https://github.com/thunderbird/thunderbird-accounts/pull/1162 but first a note:

> Apparently Keycloak doesn't store device information on a session but this might be achievable using a plugin of sorts so for now we are showing "Unknown Device" for all.

Keycloak has two different APIs to get session information.

There's an "Admin" API which doesn't doesn't return the sessions details ([but there's a ticket to request that](https://github.com/keycloak/keycloak/issues/47980)). This API is necessary for long-lived sessions, what we would call a "Connected App" to users, like Thunderbird Desktop. Technically those are OAuth sessions with "offline" scope. 

What was really in scope here was *short-lived* OAuth sessions and for *that* there's a user-level API for users to retrieve their own sessions that already /does/ contain device details. So on that technicality about the scope, we don't need to change anything in Keycloak quite yet. That's the approach taken in #1162.

PR #379, opened 2025-11-04
Mail / Thundermail Mitigation

davinotdavid:
## Description of changes
- Hides Security Settings header / footer links and Mail's Dashboard section
- Implements the progress bar / plan name on the Mail's Dashboard header

## Screenshots
image

## Solves
https://github.com/thunderbird/thunderbird-accounts/issues/370

### DIFF
--- a/assets/app/vue/views/MailView/sections/DashboardSection/components/WelcomeHeader.vue
+++ b/assets/app/vue/views/MailView/sections/DashboardSection/components/WelcomeHeader.vue
@@ -22,6 +22,10 @@ const planStorageProgress = computed(() => {
   const usedQuota = parseFloat(planInfo.value?.usedQuota);
   const mailStorage = parseFloat(planInfo.value?.features.mailStorage);
 
+  if (mailStorage === 0) {
+    return '0%';
+  }
+
   return `${(usedQuota / mailStorage) * 100}%`;
 });
 const mailStorageQuotaFormatted = computed(() => formatBytes(planInfo.value?.features.mailStorage));


src/thunderbird_accounts/subscription/views.py
```diff
@@ -198,6 +199,17 @@ def get_subscription_plan_info(request: Request):
     # Get price information
     price = subscription_item.price
 
+    # Used quota comes from Stalwart and it is optional
+    used_quota = None
+
+    try:
+        stalwart_client = MailClient()
+        account = stalwart_client.get_account(request.user.stalwart_primary_email)
+        used_quota = account.get('used_quota', 0)
```
MelissaAutumn: We should pull their quota value as well, as some folks might have specially set ones in the future.

MelissaAutumn (changes requested):
Small nits, but looks good overall.

src/thunderbird_accounts/subscription/views.py
davinotdavid: Hmmm isn't this already set from our side into Stalwart? Upon user creation apparently we:
- Add the quota from our plan into Stalwart (https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/mail/utils.py#L40-L46)
- Update the quota into Stalwart during a task (https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/mail/tasks.py#L125-L143)

As in, isn't it safe to assume that the `Plan`'s `mail_storage_bytes` is accurate?

MelissaAutumn: mail_storage_bytes should be accurate for most folks, but the value is set on every single account manually. Since you're already pulling in the used_quota, you should pull in and use the quota value as well. 

There may be cases in the future where we give someone an extra gig, or two and that might not be reflected on the plan.

assets/app/vue/views/MailView/sections/DashboardSection/components/WelcomeHeader.vue
```diff
@@ -1,13 +1,52 @@
...
+const errorMessage = ref<string>(null);
+const planInfo = ref<SubscriptionData | null>(null);
+
+const planStorageProgress = computed(() => {
+  if (!planInfo.value) return '0%';
+
+  const usedQuota = parseFloat(planInfo.value?.usedQuota);
+  const mailStorage = parseFloat(planInfo.value?.features.mailStorage);
+
+  return `${(usedQuota / mailStorage) * 100}%`;
```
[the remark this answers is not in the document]
davinotdavid: Omg 🤦 thanks and sorry for the oversight! Changed, thanks!

PR #382, opened 2025-11-05
Add payment pending step after a user subscribes which is cleared after paddle webhook. (Fixes #352)

### LINKED ISSUE #352: Payment verification step
After you complete a purchase on #337 Paddle needs to actually send the webhook saying "hey this person paid us successfully". This normally happens immediately, but there's times when this could not happen immediately. 

We should have some sort of "Payment is pending verification" step.

Something along the lines of

Image

MelissaAutumn:
This adds a payment pending flag, it's set once a user has paid. Since someone could potentially fake the end of payment api call, or a webhook could be delayed we then show them the "Payment pending" screen. Once the webhook comes from Paddle we know they're legit, and we clear that flag. 

## Screenshots
image

### DIFF
--- a/assets/app/vue/router.ts
+++ b/assets/app/vue/router.ts
@@ -102,9 +102,9 @@ router.beforeEach((to, from) => {
   }
 
   // Don't let unsubscribed users anywhere except the subscribe view
-  if (isAuthenticated && sendToSubscribe && to.name != 'subscribe') {
+  if (isAuthenticated && sendToSubscribe && to.name !== 'subscribe') {
     return { name: 'subscribe' };
-  } else if (isAuthenticated && !sendToSubscribe && to.name == 'subscribe') {
+  } else if (isAuthenticated && !sendToSubscribe && to.name === 'subscribe') {
     return { name: 'dashboard' };
   }
 


src/thunderbird_accounts/subscription/tasks.py
```diff
@@ -261,12 +272,6 @@ def paddle_subscription_event(self, event_data: dict, occurred_at: datetime.date
             plan = product_obj.plan
             if not plan:
                 logging.warning(f'Product {product.get("id")} has no plan attached!')
-            else:
-                try:
-                    user = User.objects.get(pk=user_uuid)
-                    activate_subscription_features(user, plan)
-                except User.DoesNotExist:
-                    logging.warning(f'Product {product.get("id")} has no valid user attached!')
```
davinotdavid: Ah yes, as you've pointed out in Matrix, this is still required to link the user to the plan

MelissaAutumn:
Accidentally nuked activate_subscription_features. Oops!

I added it back in, and noticed we're actually skipping a bit in the tests. So now each subscription/transaction test will have thier product and plan available to them which does increase test coverage.

assets/app/vue/router.ts
```diff
@@ -58,53 +58,57 @@ const router = createRouter({
...
+    return { name: 'home' };
   }
 
   // Don't let unsubscribed users anywhere except the subscribe view
-  if (isAuthenticated && !hasActiveSubscription && to.name != 'subscribe') {
-    return {name: 'subscribe'};
+  if (isAuthenticated && sendToSubscribe && to.name != 'subscribe') {
+    return { name: 'subscribe' };
+  } else if (isAuthenticated && !sendToSubscribe && to.name == 'subscribe') {
+    return { name: 'dashboard' };
```
davinotdavid: Nit: This works but it would be nice to have `===` and `!===` instead for the strict equality goodness

MelissaAutumn: whoops, thanks for the catch

Issue #397, opened 2025-11-06
Force login on home view of accounts.tb.pro


kewisch:
Right now we have an empty view. Behavior should be as follows:

1. If the person is logged in, redirect to https://accounts.tb.pro/dashboard
2. If not, trigger the login flow

Further it appears if you visit the dashboard while not logged in you get redirected to https://accounts.tb.pro/ , so instead we should trigger the login process there as well and return to the dashboard.

PR #406, opened 2025-11-10
Guard auth needed routes in Django

davinotdavid:
## Description of changes
- If this is a Vue route, check if it is part of the list of public routes in Vue's router, if not, redirect to Login. By doing so, we avoid rendering the Vue app at all preventing some undesirable side-effects. Also, we remove the "flashing" effect (briefly showing the Vue app skeleton) of the redirect.
- Removed the `isAuthenticated` guards on the component level in Vue-land
- Added tests

- IMHO it is a bit bad to have to maintain a list of Vue public routes within the Django app but also I don't see it increasing in side in the foreseeable future so it should be fine?

## Related issues
https://github.com/thunderbird/thunderbird-accounts/pull/405
https://github.com/thunderbird/thunderbird-accounts/issues/397

### DIFF
--- a/src/thunderbird_accounts/mail/tests/test_views.py
+++ b/src/thunderbird_accounts/mail/tests/test_views.py
@@ -10,6 +10,90 @@ from thunderbird_accounts.authentication.models import User
 from thunderbird_accounts.mail.models import Account
 
 
+class HomeViewRedirectTestCase(TestCase):
+    """Test redirect behavior for authenticated and unauthenticated users."""
+
+    def setUp(self):
+        self.client = RequestClient()
+        self.user = User.objects.create(username=f'test@{settings.PRIMARY_EMAIL_DOMAIN}', oidc_id='1234')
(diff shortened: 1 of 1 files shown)


src/thunderbird_accounts/mail/views.py
```diff
@@ -95,6 +95,12 @@ def home(request: HttpRequest):
         if request.user.plan:
             max_custom_domains = request.user.plan.mail_domain_count
             max_email_aliases = request.user.plan.mail_address_count
+    else:
+        # Check if path is included in Vue's public routes (assets/app/vue/router.ts)
+        public_routes = ['/privacy', '/terms']
```
MelissaAutumn: I wonder if we make this a json file that's accessible in both frontend and backend. 🤔

MelissaAutumn (commented):
One suggestion here, but if it's not easily obtainable we can punt it to a future ticket.

MelissaAutumn (approved):

Issue #412, opened 2025-11-12
Pre-select a domain when adding an alias


kewisch:
When you add an alias, it asks you for a domain via dropdown. The default appears to be empty. It should pre-select the first custom domain, and if there is none then thundermail.com.

Also to clarify: Do users automatically get username@tb.pro, or do they need to add aliases? I don't see tb.pro as a choice for aliases.

PR #420, opened 2025-11-12
Pre-select domain when adding custom alias + add tb.pro to ALLOWED_EMAIL_DOMAINS env var

davinotdavid:
## Description of Changes
- Pre-selects the first domain (if any) in the allowed domains list upon rendering of the domain `select` / dropdown inside the Custom Alias section when adding a custom alias.
- Adds `tb.pro` to the `ALLOWED_EMAIL_DOMAINS` env var in prod.

As a reminder of the logic behind the domains that show up in the dropdown: it is a combination of the domains in the `ALLOWED_EMAIL_DOMAINS` env var and the _verified_ Custom Domains. Non verified custom domains do not show up in the dropdown.

## Screenshots / Video

https://github.com/user-attachments/assets/f48838ae-b491-4b37-960d-9a8d0bf1e6c2

- Do we have a domain that looks like `tb.pro` but for stage? Currently the `ALLOWED_EMAIL_DOMAINS` for stage is just `stage-thundermail.com`.

## Related issues
Resolves https://github.com/thunderbird/thunderbird-accounts/issues/412

### DIFF
--- a/src/thunderbird_accounts/mail/views.py
+++ b/src/thunderbird_accounts/mail/views.py
@@ -101,7 +101,7 @@ def home(request: HttpRequest):
             'connection_info': settings.CONNECTION_INFO,
             'app_passwords': json.dumps(app_passwords),
             'user_display_name': user_display_name,
-            'allowed_domains': settings.ALLOWED_EMAIL_DOMAINS,
+            'allowed_domains': settings.ALLOWED_EMAIL_DOMAINS if settings.ALLOWED_EMAIL_DOMAINS else [],
             'custom_domains': json.dumps(custom_domains),
             'email_addresses': json.dumps(email_addresses),
             'max_custom_domains': max_custom_domains,


assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliasForm.vue
```diff
@@ -18,7 +18,7 @@ const emit = defineEmits<{
 
 const step = ref<EMAIL_ALIAS_STEP>(EMAIL_ALIAS_STEP.INITIAL);
 const emailAlias = ref(null);
-const selectedDomain = ref(null);
+const selectedDomain = ref(props.allowedDomains[0] ?? null);
```
MelissaAutumn: is allowedDomains always a non null value?

MelissaAutumn (approved):
Looks good, just want to confirm that allowedDomains won't error out if empty.

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliasForm.vue
davinotdavid: Good question! Yes, it is passed as a prop but the `computed` property that feeds into it comes from here: https://github.com/thunderbird/thunderbird-accounts/blob/main/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue#L31-L41

But you are right, there is a case where `allowed_domains` coming from `settings.ALLOWED_EMAIL_DOMAINS` is something that is not an array, which would fail the spread `...` operator so I've added new guards around there. Thanks!

ryanjjung:
Task definition changes look good to me. Since Mel had some commentary, I'll let her do the approval when she's ready.

MelissaAutumn (approved):

Issue #331, opened 2025-10-22
Remove old self-serve


MelissaAutumn:
This includes the routes, the sign up page (which is no longer needed), and anything else attached

davinotdavid:
Solved by https://github.com/thunderbird/thunderbird-accounts/pull/421

Issue #313, opened 2025-10-09
Add E2E tests for the main TB Accounts dashboard


rwood-moz:
Add new TB Accounts E2E tests for the latest UI changes, more specifically:

- Add tests for the new dashboard (#309)
- Add tests for the new Privacy & Data page (#311)
- Add tests for the new Manage MFA page (#312)

rwood-moz:
Depends on #391 to be finished first, as that will add TB Accounts auth and the various Browserstack config files etc.

rwood-moz:
Using this issue for E2E tests for the main TB Accounts dashboard page only; created #741 for adding E2E tests for the TB Accounts Mail dashboard.

PR #421, opened 2025-11-12
Remove old self-serve related code

davinotdavid:
## Description of Changes
- Removed all Django related URLs, controllers and templates + Vue forms related to self-serve, including:
  - Sign in / Sign up
  - Checkout Flow
  - Account Info
- Also removed E2E tests and related files related to self-serve, more on that on `Known Issues / Things to improve`.
- Updated the `KC_TBPRO_HOME` in docker-compose to lose the `self-serve/`
- Small update to the docs

## Known Issues / Things to Improve
- There is an auth step that happens in the E2E tests that relies heavily on the self-serve so I've deleted / commented out that whole section and left a TODO there since the only E2E tests left is the `contact/` form which is non-authenticated route. cc @rwood-moz 
- Docs are quite outdated all around, it would be great to set aside some time to do some proper brain dump on all the changes done in the past few months.

## Related issues
Solves https://github.com/thunderbird/thunderbird-accounts/issues/331

### DIFF
--- a/docs/mail.rst
+++ b/docs/mail.rst
@@ -2,7 +2,7 @@
 Mail App
 ===========
 
-The Mail App is a module of code located in ``/src/thunderbird_accounts/mail/`` which deals with our integration with to a Stalwart mail server and directory server / user management. We host part of their external directory server sql schema within Django with some minor tweaks to make things play nice. Currently we only have admin panel access, and basic user management via a self-serve page.
+The Mail App is a module of code located in ``/src/thunderbird_accounts/mail/`` which deals with our integration with to a Stalwart mail server and directory server / user management. We host part of their external directory server sql schema within Django with some minor tweaks to make things play nice.
 
 Admin Panel
 -----------


MelissaAutumn (approved):
Haven't spotted any breaking changes here, thank you!!

docs/mail.rst
```diff
@@ -1,12 +1,12 @@
+===========
 Mail App
---------
+===========
 
-The Mail App is a module of code located in `/src/thunderbird_accounts/mail/` which deals with our integration with to a Stalwart mail server and directory server / user management. We host part of their external directory server sql schema within Django with some minor tweaks to make things play nice. Currently we only have admin panel access, and basic user management via a self-serve page.
+The Mail App is a module of code located in ``/src/thunderbird_accounts/mail/`` which deals with our integration with to a Stalwart mail server and directory server / user management. We host part of their external directory server sql schema within Django with some minor tweaks to make things play nice. Currently we only have admin panel access, and basic user management via a self-serve page.
```
davinotdavid: Nit: I think we should remove the self-serve page mention here (I forgot to do it myself)

rwood-moz:
>     * There is an auth step that happens in the E2E tests that relies heavily on the self-serve so I've deleted / commented out that whole section and left a TODO there since the only E2E tests left is the `contact/` form which is non-authenticated route. cc @rwood-moz

Thanks @davinotdavid Yes the E2E tests will need to be re-written for all of the latest changes; I have an issue already on the go for this #313.

Issue #414, opened 2025-11-12
Contact support URL is incorrect on login screen


kewisch:
It points to https://accounts.tb.pro/self-servecontact

The template uses `${client.baseUrl}contact`. So either we need to set the base url to remove self-serve, or we need to change the template. Changing the base url may have some other implications, but I believe we plan on not using self-serve anymore.

davinotdavid:
@kewisch this should be fixed in production now, could you check again / close if solved? Thanks!

PR #423, opened 2025-11-13
Fix contact support link in Keycloak screens

davinotdavid:
## Description of changes

The URL for the contact page needs the extra `/` at the end because it is currently a Django route.

This PR won't be necessary / needs to be reverted once https://github.com/thunderbird/thunderbird-accounts/pull/271 gets merged since it moves the Zendesk form within the Vue app (outside of Django), hence removing the extra `/` at the end.

- In terms of the `baseUrl` in prod containing the `self-serve/` part, it seems that this is controlled in Keycloak itself as I couldn't find references to it in the codebase 🤔 

## Related issues
Related PR https://github.com/thunderbird/thunderbird-accounts/pull/271
Solves https://github.com/thunderbird/thunderbird-accounts/issues/414

### DIFF
--- a/keycloak/themes/tbpro/login/login-username.ftl
+++ b/keycloak/themes/tbpro/login/login-username.ftl
@@ -5,7 +5,7 @@
     <script>
       window._page['currentView'] = {
         formAction: '${url.loginAction}',
-        supportUrl: '${client.baseUrl}contact',
+        supportUrl: '${client.baseUrl}contact/',
         clientUrl: '${client.baseUrl}',
         // <#if realm.password && realm.registrationAllowed && !registrationDisabled??>
         registerUrl: '${url.registrationUrl}',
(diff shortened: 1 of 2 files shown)


davinotdavid:
@MelissaAutumn question about Keycloak: do you know where the `client.baseUrl` is set? I assume it is in the within the Keycloak admin > Clients but when I go there it points to yet another variable `authBaseUrl`:

image

MelissaAutumn:
It should be the home or root url for each client. You're probably looking in the master realm, which we don't use besides admin access for accounts. You'll want to swap to the tbpro realm and take a look.

They should all be in the style of:
`https://accounts.tb.pro/` or `http://localhost:8087/`.

MelissaAutumn:
Is the zendesk one ready for review? I can just review that one, and we can close this one. We have to manually push up keycloak theme changes to prod as it's not part of the CI right now.

PR #426, opened 2025-11-17
Version 1.2.0

davinotdavid:
## Description of changes

Because we have unique tagging setup, prod deploys will fail if there's not a new version tagged in pyproject.toml. So this PR bumps the version and updates the `uv.lock` after running `uv sync`.


davinotdavid:
Alright great, merging and pushing! 🚀

Issue #346, opened 2025-10-28
Add plan / subscription details to Keycloak user


MelissaAutumn:
We need to add plan and subscription details to keycloak users. 

We'll probably just need to mirror the attributes on the Plan model (note: gb is changing to bytes)
https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/subscription/models.py#L105

And add a "is_subscribed" attribute as well, so other services can redirect `!is_subscribed` users to accounts.

kewisch:
I'm curious if this is something we need for Earlybird? Are any other features or bugfixes we have planned blocked on this?

MelissaAutumn:
@kewisch we should at least add a is_subscribed to the token for early bird so other services know if they're actually suscribed, but the other data isn't required and can be hard-coded for now.

Eventually once we start allowing custom quotas it might be better to surface this information via webhooks instead of a token anyway.

MelissaAutumn:
Turns out the plan info and is_subscribed was on stage.

I've applied the following onto prod:
* Applied stage's new profile fields via json copy (https://auth.tb.pro/admin/master/console/#/tbpro/realm-settings/user-profile)
* Mapped those new fields onto the tbpro scope (https://auth.tb.pro/admin/master/console/#/tbpro/client-scopes/c8744222-c8d4-4314-9c59-5ed51872b4a3/mappers)

This function (https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/authentication/clients.py#L137) needs to be well tested since the operation is a PUT and removes all existing data and replaces it with whatever we send. I think this is the reason why I didn't finish the work. 

This function also needs to be uncommented so we can trigger the updates when a user subscribes or unsubscribes
https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/subscription/utils.py#L44

Once it's uncommented you can also test the sync via admin actions (https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/authentication/admin/actions.py#L113)

PR #433, opened 2025-11-19
Initial work for propagating subscription data to Keycloak

### LINKED ISSUE #407: Harden account creation if celery cannot create a mail account
This happened to @aatchison, our celery worker couldn't connect to thundermail but continued along and did the rest of the steps. This caused activate_subscription to be skipped and thus account/email were never created and plan_id was never set.

To fix this I found the webhook in paddle, re-ran it. Then manually created an account object for that user. After that, I ran the fix empty stalwart id action on the newly created account object, and adjusted the quota.

That seems to have tied everything up.

MelissaAutumn:
Pt1 of #346 

Punting tasks:
* Document the updated keycloak jwt data
* Document what to do on client login
* Add additional tests
* Ensure it all works locally (only tested admin panel flow atm)
* Click-ops the Keycloak config over to stage/prod and document what changed

In order to fix #407 I'm punting the above tasks to another ticket as this became a bit of a mega PR for some big underlaying changes. 

What this PR does:
* Add initial documentation for integrating Keycloak into your service
* Clean up old outdated documentation
* Add an admin action to manually activate subscription features (this happens when we get the paddle subscription.created webhook)
* Add an admin action to run the plan info sync to keycloak (don't use this right now)
* Added the ability to sync plan information to Keycloak (via the above action)
* Cleaned up admin panel information
* Added the Stalwart Principal view to the admin panel as a custom page (it already existed, but wasn't linked)
* Cleaned up the page above
* Added some basic test factory functions for keycloak/stalwart data
* Updated a bunch of tests

### DIFF
--- a/docs/mail.rst
+++ b/docs/mail.rst
@@ -1,14 +1,15 @@
 Mail App
 --------
 
-The Mail App is a module of code located in ``/src/thunderbird_accounts/mail/`` which deals with our integration with to a Stalwart mail server and directory server / user management. We host part of their external directory server sql schema within Django with some minor tweaks to make things play nice. Currently we only have admin panel access, and basic user management via a self-serve page.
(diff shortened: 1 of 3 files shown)


davinotdavid (approved):
Awesome! LGTM 🎉

Issue #413, opened 2025-11-12
Custom Domain verification always works (when it shouldn't)


kewisch:
STR:

1. Set up a custom domain. I set mine up on a subdomain if it matters.
2. Don't actually add an DNS entries
3. Verify the domain

Expected:
* Verification fails

Actual
* Verification succeeds

davinotdavid:
From the top of my head, it is wired up to always verify if the `IS_DEV` flag is up only.

For non-development environments, it is considering verified if there are no critical errors during the troubleshooting process (warnings errors were deemed fine in the discussions around the custom domain extensive checks).

We are currently returning the full array of critical errors and warnings from the verify endpoint so I'd love to see what the Network tab says upon clicking `Verify` on that Custom Domain @kewisch

Just tried in prod and I can't verify custom domains that I own nor a subdomain because it gives a `mxLookupError` critical error.

davinotdavid:
From today's Account meeting sync, it seems like we shouldn't be using Stalwart's Troubleshooting API to verify a custom domain – Instead, we should check if the return host matches our host since Stalwart would simply identify that the domain has MX records to deliver mail to it but not necessarily _our_ host.

PR #438, opened 2025-11-20
Custom domain verification refactor

davinotdavid:
## Description of changes
- Instead of using [Stalwart's Troubleshooting API](https://stalw.art/docs/management/troubleshoot/), which only checks if the Custom Domain is accessible (not if the hostnames are set up correctly), we are now using the `dnspython` package to check the hostnames from the DNS records:
  - In `development` all verification succeeds by default as I don't think we can verify in development.
  - In other environments, it does three checks:
     1. MX Records exist and point to the correct host (Critical, fails verification)
     2. SPF Record exists and includes the correct host (Warning if missing, doesn't fail verification)
     3. DKIM Record exists (Warning if missing or wrong, doesn't fail verification)

- This is a fairly "permissive" / "loose" verification as we are essentially just checking for the `MX` record. This can be hardened as needed (moving SPF or DKIM Record checking as critical is almost a 1 line change).

## Related issues
Fixes https://github.com/thunderbird/thunderbird-accounts/issues/413

### DIFF
--- a/src/thunderbird_accounts/mail/clients.py
+++ b/src/thunderbird_accounts/mail/clients.py
@@ -24,6 +24,17 @@ class StalwartErrors(StrEnum):
     OTHER = 'other'
 
 
+class DomainVerificationErrors(StrEnum):
+    """Domain verification error codes returned by verify_domain()"""
+
+    # Critical errors (fail verification)
+    MX_LOOKUP_ERROR = 'mxLookupError'
+
+    # Warnings (do not fail verification)
+    SPF_RECORD_NOT_FOUND = 'spfRecordNotFound'
+    DKIM_RECORD_NOT_FOUND = 'dkimRecordNotFound'
+
+
 class MailClient:
     """A partial api client for Stalwart
     Docs: https://stalw.art/docs/api/management/endpoints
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/mail/clients.py
```diff
@@ -513,84 +424,129 @@ def get_dns_records(self, domain_name: str) -> list[dict]:
         return data.get('data')
 
     def verify_domain(self, domain_name: str):
```
MelissaAutumn: For the view we might want to look into rate limiting and moving this function to task in the near future

src/thunderbird_accounts/mail/clients.py
```diff
@@ -513,84 +424,129 @@ def get_dns_records(self, domain_name: str) -> list[dict]:
...
+
+            for rdata in txt_answers:
+                # rdata.strings is a list of bytes
+                txt_content = b''.join(rdata.strings).decode('utf-8')
+                if txt_content.startswith('v=spf1') and expected_spf_include in txt_content:
+                    has_spf = True
+                    break
+
+            if not has_spf:
+                warnings.append('spfRecordNotFound')
```
MelissaAutumn: We should probably define these as a enum or constant value since they're everywhere.

MelissaAutumn (approved):

Issue #285, opened 2025-09-18
Pre-fill email address when showing paddle popup


kewisch:
When a user selects a subscription in accounts, is it possible that we'd pre-fill their email address for paddle?

<img width="1372" height="912" alt="Image" src="https://github.com/user-attachments/assets/3587971d-dc7a-432a-8ae4-c37def16a280" />

MelissaAutumn:
We can prefil with the customer's email field in paddlejs
https://developer.paddle.com/paddlejs/methods/paddle-checkout-open

PR #448, opened 2025-11-21
Change how we redirect the user after the checkout process is completed

### LINKED ISSUE #417: prefill alternate email on Paddle subscription form
If we have the alternate email for an account, we should prefill it into the subscription email address field. This will make it less confusing for users.

https://developer.paddle.com/paddlejs/methods/paddle-checkout-open#params

Image

### LINKED ISSUE #441: Move checkout success event to a successUrl instead
Paddle folks mentioned we should use [successUrl](https://developer.paddle.com/build/checkout/handle-success-post-checkout#redirect-to-a-success-page) instead of the checkout completed event to handle any client-side redirects as Paypal is a separate window, so events don't propagate correctly. 

That successUrl should essentially be the POST request we do on the checkout completed event, and then redirect the user to dashboard.

MelissaAutumn:
Fixes #441 

Wow a mouthful! Paddle folks mentioned that checking for `checkout.complete` event doesn't really work for Paypal, they suggested we use the successUrl property instead. Pretty odd, but okay easy enough change. 

## Changes
* **Fix mail tests not being picked up the main test run :scream_cat:**
* Fixed some invalid mail tests
* Fixed localization and styling of the checkout heading (Payment)
* Added a generic error message if we somehow don't recieve a paddle instance
* Added a post-checkout success message
* Added a dev only route to save the checkout's transaction id to session
* Updated the checkout completed api route to a redirect route
* On dev it retrieves and pops the transaction id from session so it can retrieve the subscription webhook details manually
* I got annoyed at the Admin panel's logout button not working correctly so I fixed that. 

I intended to also fix #417 (and also I guess #285 ??) but I'm awaiting feedback from the Paddle folks on its behaviour.

### DIFF
--- a/src/thunderbird_accounts/subscription/views.py
+++ b/src/thunderbird_accounts/subscription/views.py
@@ -79,7 +79,7 @@ def on_paddle_checkout_complete(request: Request, paddle: Client):
     transaction_id = request.session.pop(SESSION_PADDLE_TRANSACTION_ID)
     user = request.user
     if not user:
-        sentry_sdk.capture_message(`
+        sentry_sdk.capture_message(
             '[on_paddle_checkout_complete] User passed login_required but did not have user object!'
         )
 


src/thunderbird_accounts/mail/tests/test_tasks.py
```diff
@@ -193,31 +193,6 @@ def test_success_account_already_exists_on_stalwart(self):
             self.assertEqual(oidc_id, task_results.get('oidc_id'))
             self.assertEqual(mock_stalwart_pkid, task_results.get('stalwart_pkid'))
 
-    def test_account_already_exists(self):
```
MelissaAutumn: We now update the account in place, this fixes an issue we're experiencing where people who have an account already but haven't subscribed (so like everyone at mzla) are currently not having their account references created.

src/thunderbird_accounts/subscription/views.py
```diff
@@ -59,28 +63,50 @@ def get_paddle_information(request: Request):
...
-    data = json.loads(request.body.decode())
-    transaction_id = data.get('transactionId')
+@require_http_methods(['PUT'])
+def set_paddle_transaction_id(request: Request):
+    """This error is used to work-around the fact we don't get a transaction id from Paddle's successUrl redirect.
+    It kinda sucks, but it works for now."""
+    if settings.IS_DEV:
+        data = json.loads(request.body.decode())
+        request.session[SESSION_PADDLE_TRANSACTION_ID] = data.get('txid')
+    return JsonResponse({'success': True})
```
MelissaAutumn: This shouldn't get called in non-dev environments but in case it does we just return true.

PR #456, opened 2025-11-26
Add a Sign-Up page to Accounts

### LINKED ISSUE #455: Port Keycloak's Sign Up form to Accounts and gate users
Bit of epic issue but I need to do the following:
* Port the Sign Up vue form from Keycloak to accounts
* Allow the user to enter their email
* Anyone not on the allow list will be signed up for the waiting list
* Anyone on the allow list will have their account created
* Send the allowed user to the subscribe screen

We'll need a new allow list object that will just be email and probably name. 

We should be explicit about the waiting list sign up too.

MelissaAutumn:
Fixes #455 

I've done some large re-organization here to help improve boot time, and allow us to import keycloak components into account's vue app. 

In the future it'd be nice to have a normalized folder structure (e.g. `assets/modules/accounts` and `assets/modules/keycloak`.)  While I'm able to simply just import the keycloak assets and have vite bake them into the build, I still had to copy over the i18n manually as keycloak uses property files for their locale. 

This also includes some additional improvements like the ability to turn off the base template for accounts for a full-page view, and also I've added support for a bulk importer for the newly created AllowListEntry model. 

## Changes
* Added AllowListEntry model
* Added bulk import for AllowListEntry (in admin panel)
* Adjusted KeycloakClient.import_user to send any aia action as an email
* Adjusted KeycloakClient.import_user to send over passwords
* Added slots to Keycloak's register view
* Refactored Keycloak's frontend vite to use `@kc` for local imports
* Added SignUpView that's mostly an import of Keycloak's RegisterView
* Only slightly sorry for not naming SignUpView as RegisterView
* Endured a few migraines
* Split Keycloak's base template from App.vue into BaseTemplate
* Injected primary domain to Keycloak's template via env variables
* Added a link to account's sign-up page on Keycloak's login page via env variables (if registration is disabled on kc)
* Added wait list url setting to accounts
* Disabled Sentry for dev / tests (it was getting annoying in stack traces)
* Add ruff check for rogue print statements in our codebase (you can use `noqa: T201` to suppress them per line)
* Switched to sublime-text with a language server, so added some helpers for typing within django

## Docker changes

I've moved frontend prod build, and collectstatic to the dockerfile's build step, instead of on boot. This should speed up container boots since we don't need to rebuild the frontend everytime we boot. Technically this step isn't even needed in local dev, but it's there so we can use the same container on stage/prod. 

If you experience any issues during testing wrt to docker please document and post here as part of the review. Thanks!

### DIFF
--- a/assets/app/vue/locales/en.kc.json
+++ b/assets/app/vue/locales/en.kc.json
@@ -18,5 +18,5 @@
   "recoveryEmail": "Recovery email",
   "recoveryEmailHelp": "You will be required to verify this email address to continue.",
   "emailVerifyInstruction": "An email with instructions to verify your recovery email address has been sent to your address.",
-  "invalidPasswordConfirmMessage": "Password confirmation doesnt match."
+  "invalidPasswordConfirmMessage": "Password confirmation doesn't match."
 }
--- a/keycloak/themes/tbpro/assets/vue/views/RegisterView/index.vue
+++ b/keycloak/themes/tbpro/assets/vue/views/RegisterView/index.vue
@@ -101,6 +101,7 @@ export default {
         name="email"
         required
         autocomplete="email"
+        type="email"
         :error="recoveryEmailError"
         :help="$t('recoveryEmailHelp')"
       >


keycloak/themes/tbpro/assets/vue/App.vue
```diff
@@ -1,164 +1,17 @@
...
-
-      <!-- Right side: Panel -->
-      <div class="right-side">
-        <div class="panel">
-          <router-view />
-        </div>
-      </div>
-    </div>
-  </section>
+  <base-template>
```
MelissaAutumn: Simplified so we can re-use the base template without having to refactor the App.vue logic.

run-tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sudo docker compose exec accounts uv run manage.py test $1
```
MelissaAutumn: ex/ `./run-tests.sh src.thunderbird_accounts.authentication.tests.SignUpViewTestcase.test_not_on_allowed_list`

I've switched to sublime-text with lsp-pyright and lsp-ruff, so no more fancy pycharm test runners for me 😅

davinotdavid (commented):
Super cool! Thanks! 🎉 I will give it a spin running locally and this review comment is just code-wise and it all lgtm, just one nit for now:

MelissaAutumn:
Haha that's odd. It's supposed to be email verification, and will look into the password matching.

MelissaAutumn:
Fixed the client side password/confirm password check, and added the actual action to the update account page

For future fixes see https://github.com/thunderbird/thunderbird-accounts/issues/466

davinotdavid (commented):
Nice! I can verify that the Verify Action (heh) now shows the "Verify Email" text, however I am seeing two small things in the Sign Up form:

1. `doesnt` instead of `doesn't`

Image

2. No `recovery email` email type validation

Adding a `type="email"` to L114 in `keycloak/themes/tbpro/assets/vue/views/RegisterView/index.vue` seems to do it

Image

PR #468, opened 2025-12-04
Create archive folder if it does not exist

### LINKED ISSUE #457: Create archive folder in Thundermail
Will add a check on login (for existing accounts) or email creation where if they don't have an archives folder it will create one.

MelissaAutumn:
Fixes #457 

This was annoying as my dev stalwart config was broken. So that's what the first commit is about. I also had some minor db revisions pending, since they're not actually db related it's not important. 

I've changed the location of where we check if we need to manually add the archives folder a few times, so please make sure it works on your end. You can create a custom route with a Mailbox/get (with only account_id and no other params) to view all your folders in jmap, or hook it up to Thunderbird or another IMAP provider to check.

## Changes:
* Add thunderbird_accounts.mail.middleware.FixMissingArchivesFolderMiddleware
* Add Account.verified_archive_folder (bool)
* Add functionality to query and create (if needed) an archives folder via jmap
* Add settings.STALWART_ARCHIVES_FOLDER_NAME which controls the name of the archives folder that is created.
* Added some tests

### DIFF
--- a/src/thunderbird_accounts/mail/tests/test_middleware.py
+++ b/src/thunderbird_accounts/mail/tests/test_middleware.py
@@ -30,7 +30,7 @@ class FixMissingArchivesFolderMiddlewareTestCase(TestCase):
                         'queryState': 'sai',
                         'canCalculateChanges': True,
                         'position': 0,
-                        'ids': ['a'] if not mailbox_ids else mailbox_ids,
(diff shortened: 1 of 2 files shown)


davinotdavid (approved):
Code LGTM! Also, I can verify that it works well with the `admin@example.org` account!

Image

PR #473, opened 2025-12-08
Use a shared json file to determine which routes are public

### LINKED ISSUE #408: Have a JSON file for routes shared with Vue and Django
## The problem

We have a Vue app that gets rendered in a specific route within the Django app. However, some of the routes in the Vue app should be guarded (`/dashboard`, for example) whereas others should not be guarded (`/privacy`, for example).

We can simply hard-code / duplicate the routes that are supposed to be guarded in the Django `urls.py` but that is not super maintainable long term and prone for errors since one would need to remember to add a public route in both sides.

## A possible solution

We could potentially attempt to have a single JSON file that would represent routes to be consumed both by the Vue app and the Django app. This would eliminate the duplication and avoid errors in the future.

MelissaAutumn:
Fixes #408 

I'm not sure I really like this solution, and I looked for existing easy solutions to pre-render this information on the frontend side. However it doesn't seem like there's a "run this javascript/typescript function with static data and replace the call with the static result". 

The solution here long-term is probably to create a vite plugin for the rollup process, but I didn't want to over-optimize/over-engineer so we'll see if this underperforms.

### DIFF
--- a/assets/app/vue/router.ts
+++ b/assets/app/vue/router.ts
@@ -1,4 +1,4 @@
-import { createRouter, createWebHistory } from 'vue-router';
+import { createRouter, createWebHistory, RouteRecordRaw, RouterOptions } from 'vue-router';
 
 // Accounts Routes
 import DashboardView from '@/views/DashboardView/index.vue';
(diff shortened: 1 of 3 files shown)


davinotdavid (commented):
Omg sorry for the super long delay on reviewing this!

Hmmm, I honestly don't know a better way of doing this at the moment to suggest but I am not sure if I particularly like this solution as it already feels quite more complex than the previous way of maintaining that array of strings on the backend.

Do we foresee having lots of patches related to public routes in accounts? I think it would be quite rare but if not, then I agree that it would definitely be more scalable to do as you proposed here!

davinotdavid:
@MelissaAutumn sounds good to me!

markstos:
I don't love the shared-JSON-file design either. Here's a PR with an alternate solution that meets the goal to keep these two lists of public Vue files in sync without it:
https://github.com/thunderbird/thunderbird-accounts/pull/1169

PR #477, opened 2025-12-09
Retrieve prices from Paddle. Calc discount as well. (Fixes #350) (Fixes #430)

### LINKED ISSUE #350: Show localized pricing on dashboard
We have pricing localization enabled for a few different currencies. So we should also show that on the dashboard. (e.g. if someone is Canadian they'll see the plan pricing converted to CAD.) 

We'll want to retrieve the pricing information from Stalwart directly (and cache it for like 5-10 minutes or so) and to display it on the dashboard. 

We sort of do this for the subscribe page on #337, but we should do it on the backend and cache per plan.

### LINKED ISSUE #430: Subscription information has paid amount even when a discount is applied
When you have a discount (e.g 100%), then the subscription info still shows the rate. It might be worth noting that they are not paying, or paying less, and then also note what the rate will be on renewal. We could also use this to indicate early if prices are changing.

MelissaAutumn:
## Description of changes

We now retrieve a user's subscription pricing in their localized currency (or at least the one they're paying in) via Paddle. We additionally retrieve and apply any discount to that price. (See screenshot for discount applied)

~~We also cache that information for an hour so we're not constantly waiting on Paddle. I encrypt the json blob with the cryptography package and a newly defined secret `CRYPTO_SECRET_KEY`. We use this in combination with a salt derived from the subscription's last update time as a string. This is just a stupid trick to invalidate old data faster, since their sub's last update time will change if they or we update their sub. Which will invalidate their cached pricing data.~~

~~I've set a default for dev environments (can't wait for about 200 emails saying "LEAKED SECRET DETECTED!"..)~~

On the frontend I use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) to format the currency string as vue-i18n can only localize pricing info based on locale code not currency code.

This is just missing some tests and a brand new secret key on stage and prod.

## How to test

- During the checkout process, change the `Country` to `Germany` (or other country of your choice that has a different supported currency)
- After the checkout process, click on the top right User Avatar to navigate back to the Accounts dashboard
- Check the Subscription Info component for the country currency specific pricing (**it should be calculated as yearly / 12, floored**)

Bonus points: re-test it using the discount code in this [Paddle link](https://sandbox-vendors.paddle.com/discounts-v2) (the credentials are in 1Password to access Paddle's Sandbox account)

## Screenshots

Example for CAD, using 100% discount
image

Example for EUR
image

## Related issues

Fixes #350 
Fixes #430


MelissaAutumn:
> This can be a follow-up issue, but what I liked in paddle checkout is that it tells you what is due now and what will be due after renewal. For those with a 100% discount it would be great if it tells you what is due at renewal, down where it says "Auto-renews on November 24,2026"

Perfect, I found the $0 was a little funny, especially a month before you renew. Mind a cutting a ticket for your vision? :smile:

src/thunderbird_accounts/authentication/migrations/0014_alter_allowlistentry_options.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ('authentication', '0013_alter_user_plan_allowlistentry'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='allowlistentry',
+            options={'verbose_name_plural': 'Allow list entries'},
+        ),
+    ]
```
MelissaAutumn: You can nix this file, it's an accidental change i think

.env.example
```diff
@@ -3,6 +3,8 @@
 
 SECRET_KEY=insecure-secret-key
 LOGIN_CODE_SECRET=insecure-login-secret-key
+# Must be created from instructions: https://pypi.org/project/cryptography/
```
MelissaAutumn: An example secret, this will need a new secret stage / prod value.

MelissaAutumn: ...i don't think this is used for anything actually.

MelissaAutumn:
> Re-worked this so we instead store the discount amount, and type on the Subscription object, and the local price on the Subscription Item object.
> 
> This needs additional work to ensure we pull the correct webhooks, but I'll cut a follow up PR for that. This is just missing tests now.

I don't remember what webhooks these would be...I should really cut the ticket before I say I'm going to.

.env.example
davinotdavid: Seems not, at least no references in the code so I will delete it

src/thunderbird_accounts/subscription/migrations/0010_subscription_discount_amount_and_more.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('subscription', '0009_remove_plan_mail_storage_gb_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='subscription',
+            name='discount_amount',
+            field=models.CharField(blank=True, help_text='Discount (if any) that is applied to the Subscription (after all items are totaled.)<br/>This field can be a flat amount or percentage. See `discount_type` for what this value represents.', max_length=256, null=True),
```
malini, line 16: should this and price override amounts be FloatFields?

src/thunderbird_accounts/subscription/admin.py
```diff
@@ -1,8 +1,54 @@
...
+            ngettext(
+                'Failed to retrieve %d discounts / localized price.',
+                'Failed to retrieve %d discounts / localized prices.',
+                errors,
+            )
+            % errors,
+            messages.ERROR,
+        )
+    if sum([success, errors]) == 0:
+        modeladmin.message_user(
```
malini, line 47: why do we do this?

davinotdavid: The Django admin command is to backfill the subscription information for existing users and this line specifically is to surface a message to the Django admin user regarding what is the result of the execution.

malini (approved):
ship with this for now, I'll cut a ticket to convert from char to float in the db if necessary

src/thunderbird_accounts/subscription/migrations/0010_subscription_discount_amount_and_more.py
davinotdavid: I believe this is just for convention / since we are using `CharFields` everywhere else. We can ask @MelissaAutumn later next week if there's a better reason!

malini:
filed https://github.com/thunderbird/thunderbird-accounts/issues/787

PR #509, opened 2026-01-16
Apply bolt's font-size only on BaseTemplate.

### LINKED ISSUE #508: Keycloak theme's font-size is being applied to TB Account's dashboard

MelissaAutumn:
Fixes #508 

I don't know why this is bolt's default size, but we can fix that in a later issue as it seems like the tbpro theme is correctly sized.


davinotdavid (approved):
Oh weird, lgtm for now pending deeper investigation later on

PR #511, opened 2026-01-19
Update App Passwords copy and styling (Fixes #485)

### LINKED ISSUE #485: App Password Design Updates
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/69168562a0332a7b892c0d9a/

Additionally the tooltip url will go to https://support.tb.pro/hc/articles/46805020320275-Create-an-App-Password

MelissaAutumn:
Fixes #485 

Updates the copy and styling to match Zeplin + text from @kewisch. Also added in a defines.ts we can use for consts like support article links. 

## Screenshots
image
image

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -93,6 +93,7 @@
           "displayName": "Display name",
           "appPassword": "App Password",
           "appPasswords": "App Passwords",
+          "close": "Close",
           "change": "change",
           "set": "Set",
           "notSet": "Not set",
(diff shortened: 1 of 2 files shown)


davinotdavid (commented):
Code lgtm! Nice! Just a few points:

The styling for the link looks a bit off from the `services-ui` link button in terms of colours.

Image


Probably these are things that were already there before but I've noticed some things on the App Password side:

1. The "Confirm Password" input doesn't get cleared if the user cancels the "Change Password" action but the "New Password" gets cleared.

2. Maybe this just requires a double checking but I've created a brand new account as  `admin@example.org` and it shows that the App Password was already set (?). I wonder if there's a bug somewhere in the logic of retrieving / saving the one app password that we currently support 🤔

PR #533, opened 2026-01-22
Fix being able to block mail to a certain domain (Fixes #531)

### LINKED ISSUE #518: Restrict deleting custom domains
...until the user has removed all aliases for that domain.

### LINKED ISSUE #531: Investigate into not using a Stalwart domain entry for dns verification
If someone creates a custom domain they can effectively block sending emails to that domain for everyone.

...that's not ideal.

MelissaAutumn:
Fixes #531 and fixes #518

* Don't create a stalwart domain for unverified domains
* Once verified then create a stalwart domain entry
* Streamline dev-mode's handling of domain verification
* Fix #518 by requiring the user to remove all aliases first
* Show domains in admin panel

### DIFF
--- a/src/thunderbird_accounts/mail/admin/__init__.py
+++ b/src/thunderbird_accounts/mail/admin/__init__.py
@@ -1,7 +1,8 @@
 from django.contrib import admin
 
-from thunderbird_accounts.mail.admin.models import AccountAdmin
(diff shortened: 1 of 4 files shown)


davinotdavid (approved):
Code lgtm! Happy to re-test this on stage when deployed

PR #535, opened 2026-01-22
Only remove a stalwart domain entry if there's a domain referencing one.

MelissaAutumn:
Small fix here, before it was attempting to remove a stalwart domain entry right before we removed the local reference. 

This changes it so we only try to remove a stalwart one if we have a local one referencing it.

### DIFF
--- a/src/thunderbird_accounts/mail/views.py
+++ b/src/thunderbird_accounts/mail/views.py
@@ -545,8 +545,13 @@ def remove_custom_domain(request: HttpRequest):
 
     stalwart_client = MailClient()
     try:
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
So fast! ⚡ I was about sending this [Sentry error](https://thunderbird.sentry.io/issues/7211388084/?alert_rule_id=15849274&alert_timestamp=1769120953774&alert_type=email&environment=stage&notification_uuid=bd81716b-f154-4a6c-a942-8fa3531e7ecc&project=4508965616680960&referrer=alert_email) to the channel 😅 

LGTM!

PR #567, opened 2026-01-29
Adjust navigation bar to use the updated design and fix some quirks (Fixes #520)

### LINKED ISSUE #520: Navigation header changes
Prototype here:
https://www.figma.com/proto/xIrjrdyb6W9NWPrKDoujem/TBPro_Web_2026_Q1_Navigation_v01?page-id=245%3A17&node-id=2034-132&viewport=1450%2C1174%2C0.5&t=FB3534wHSHYOFW1I-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2034%3A132

MelissaAutumn:
Fixes #520

This took me a long while to figure out how to properly center the links without it being affected by the logo being different widths...but it's done and I think it works nicer now? 

I re-implemented the click-away on the dropdown nav and then reverted it with a small change to make it feel a little better. (The avatar's round portion is now the cursor: pointer target instead of the bounding box which wasn't fully clickable.)

This also fixes the dropdown nav from being under the card panels.

## Screenshots

image
image
image

### DIFF
--- a/assets/app/vue/components/FooterBar.vue
+++ b/assets/app/vue/components/FooterBar.vue
@@ -46,6 +46,7 @@ const navItemsMail = [
 const currentRoute = useRoute();
 
 const isThundermail = computed(() => currentRoute.path.startsWith('/mail'));
+const isSubscribePage = computed(() => currentRoute.path.startsWith('/subscribe'));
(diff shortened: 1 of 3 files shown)


rwood-moz:
Oh wait I guess it's the `account` menu item that is hidden but not `logout`, nvm 👍

PR #587, opened 2026-02-12
Sends email in querystring param when redirecting to wait list

radishmouse:
If user tries to sign up but  is not in allow list, they are redirected to the tb.pro wait list.
[
To save the user from re-entering their email address on the wait list form](https://github.com/thunderbird/thunderbird-website/issues/1112), we send the email as a querystring param.

The tb.pro wait list page will prefill the form with that email address, after [this PR](https://github.com/thunderbird/thunderbird-website/pull/1128) is deployed.

Also updates django test ("test_not_on_allowed_list"), which was confirmed as passing with this command:
```
docker compose exec accounts uv run python manage.py test thunderbird_accounts.authentication.tests
```

### DIFF
--- a/src/thunderbird_accounts/authentication/tests.py
+++ b/src/thunderbird_accounts/authentication/tests.py
@@ -7,6 +7,7 @@ from django.contrib.admin import AdminSite
 from django.core.exceptions import PermissionDenied
 from django.forms import model_to_dict
 from django.http import HttpRequest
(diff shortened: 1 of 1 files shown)


MelissaAutumn (changes requested):

radishmouse:
Merging, since only prior feedback was how to bypass copilot scanner.

PR #592, opened 2026-02-12
Check if a user exists before adding them to the mailing list (Fixes #591)

### LINKED ISSUE #591: Mailchimp call fails if user is already on the list
This will fail most welcome emails as most folks are already on the list from via wait list sign up.

MelissaAutumn:
Fixes #591 

I've re-organized the add_subscriber_to_mailchimp_list function. It now:
* GET request to the see if the email already exists on the list
* If the email exists: POST request to add the tag
* If the email does not exist: POST request to add the email with the tag

Unfortunately this adds a lot of noisy error logic. I've also consolidated the mailchimp api request to a function to make it easier to mock.

### DIFF
--- a/src/thunderbird_accounts/subscription/tasks.py
+++ b/src/thunderbird_accounts/subscription/tasks.py
@@ -526,7 +526,7 @@ def add_subscriber_to_mailchimp_list(self, user_uuid):
                 'task_status': 'failed',
                 'reason': 'mailchimp error',
                 'error_msg_title': error_details.get('title', 'N/A'),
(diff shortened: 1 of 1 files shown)


src/thunderbird_accounts/subscription/tasks.py
davinotdavid: Question! How does the `retry` decorators here work? There are various `except` blocks to catch exceptions but they all return a dict (specially the very last one on L513) so I wonder if we should call a retry method of sorts manually in that case or if it auto-retries automagically somehow?

Issue #330, opened 2025-10-22
Queue visibility


MelissaAutumn:
I've moved some jobs to async in #328 which is great. Now the task isn't blocking. Unfortunately we don't have any visibility into the queue. 

We currently save reports to redis, so maybe something like flower would work here. It seems to have an open metrics api that should be parsable by a third party service of our choosing.
https://flower.readthedocs.io/en/latest/

Or we could pipe the job results right to the admin panel, but I had mixed results getting this to work while I was playing around with it.
https://django-celery-results.readthedocs.io/en/latest/index.html

MelissaAutumn:
Also this is merged for dev access right now. Did you want to look into setting up the infra side @ryanjjung, or should I handle that and just tag you on the review?

malini:
@MelissaAutumn to confirm it's as expected

kewisch:
@ryanjjung @MelissaAutumn what's the status on this issue? Can it be closed? Is this a target for Earlybird? Is it actually still in review?

ryanjjung:
I created [this ticket](https://github.com/thunderbird/thunderbird-accounts/issues/609) to cover the work for it, and that is complete. imo, we can close the issue, marked complete.

malini:
Instructions on viewing status is in the readme https://github.com/thunderbird/thunderbird-accounts#accessing-the-flower-web-interface

PR #606, opened 2026-02-18
Features/330 queue vis

MelissaAutumn:
Doesn't quite finish up #330 as we need to do the stage/prod part but I figured it's good to break this up. 

This adds queue visibility, so we can see what worked and what failed. I was hoping it would let us manually retry tasks but it's doable via the api so we'll expand on that in the future.

This isn't meant to replace Sentry, in fact the new TaskFailed exception is ignored when shipping exceptions to Sentry. It's meant to give us an idea on our queue health, and allow us to investigate it in finer detail if needed.

This PR also adds a new TaskFailed exception that will be used instead of returning a dict with 'task_status': 'failed'. We can clean up the task_status: success in the future, I just wanted to reduce PR / codediff noise. 

Screenshots:
image
image
image


docker-compose.yml
```diff
@@ -68,6 +68,14 @@ services:
       <<: *env
       TBA_CELERY: yes
 
+  flower:
+    <<: *accounts
+    ports:
+      - "5555:5555"
+    environment:
+      <<: *env
+      TBA_FLOWER: yes
```
davinotdavid: I am not familiar with Flower and its inner workings but I was playing around for a bit locally and when I try to refresh the queues through their UI I get this:

`FLOWER_UNAUTHENTICATED_API environment variable is required to enable API without authentication`

Image

Is this something that we might need?

Issue #609, opened 2026-02-19
Create infrastructure for hosting Flower in stage


ryanjjung:
We need to get Flower and Celery running in the same cluster, and I have recently redesigned our Fargate module to do just that. Let's get both of these things running in Stage on a new cluster that runs both services.

Stretch goal: Get Accounts running in the same cluster, condensing everything down into one place.

PR #617, opened 2026-02-26
Build new ECS cluster in stage; move celery; launch flower

ryanjjung:
There are a handful of updates wrapped up here. Let's list them:

- We update our tb_pulumi dependency to `main`, since we depend on fancy new features.
- All of the image names in our stage tb_pulumi YAML file are changed to use a variable reference. Now we can update one line at the top of the file, and it will be applied to all tasks that use the image being deployed. This helps us be sure that all of these services are running the same version and prevents a lot of YAML eyeball scanning.
- Build a new `accounts-stage` ECS cluster based on the new `AutoscalingFargateCluster` tb_pulumi pattern. This lets us condense all of our services into a single cluster and a few other nice things.
- Migrates the Celery workers in stage to this new cluster.
- Brings online the queue visibility tool Flower in the new cluster.

Ref: #609

### DIFF
--- a/pulumi/config.stage.yaml
+++ b/pulumi/config.stage.yaml
@@ -97,8 +97,7 @@
 # This forms the base of Accounts-based container definitions. The blank fields commented here must be set on any
 # inheriting container definition:
 #   environment:
-#   logConfiguration.options.awslogs-group:
-#   logConfiguration.options.awslogs-stream-prefix:
+#   logConfiguration
 .accounts_contdef: &ACCOUNTS_CONTAINER_DEFINITION
   image: *ACCOUNTS_IMAGE
   essential: true


pulumi/config.stage.yaml
```diff
@@ -175,11 +178,483 @@ resources:
       transit_encryption_enabled: True
       apply_immediately: True
   
+  tb:fargate:AutoscalingFargateCluster:
```
ryanjjung: This is what the new cluster config looks like. I think it's a little more clear (and there's a bunch of [documentation](https://thunderbird.github.io/pulumi/patterns.html#designing-autoscaling-fargate-clusters) to guide an engineer through this design process and [describing the class](https://thunderbird.github.io/pulumi/fargate.html#tb_pulumi.fargate.AutoscalingFargateCluster)).

Both the Celery and Flower configs here were copied from the original Celery task definition that is being removed in this PR. I have updated their configs accordingly where it matters, but they're mostly the same.

ryanjjung:
As a final note, I think it's worth mentioning that I don't think the CI has the permissions necessary to actually perform the next deployment. That deployment will want to destroy the old celery cluster. This is fine, I expect it, but I will probably have to perform the update myself before CI will start working again.

PR #626, opened 2026-03-06
Remove on_paddle_checkout_complete and replace it with a 2.5s poll

### LINKED ISSUE #458: Work around third party payment processors in Paddle
Unfortunately https://github.com/thunderbird/thunderbird-accounts/issues/441 simply redirects the user from the tiny pop up window and not the /subscribe page.

We should keep that checkout window redirect, but also poll on the /subscribe page for is_payment_pending or is_active_subscription. If either of those are true we should redirect the user.

MelissaAutumn:
Fixes #458 

This replaces on_paddle_checkout_complete / the use of the successUrl property in paddlejs with a 2.5s polling route. This route will query paddle directly via their api for the transaction status. 

This works around the bug in paddlejs' checkout where Paypal (and presumbly every other) third party processor does not send the checkout completed event, and redirects the small Paypal checkout window it opens to the successUrl. (Which is unusably small.) 

I've also updated paddlejs and services-ui to their latest version.

## How to test?

Ensure you have Paddle setup, and run through it. I don't think we have a sandbox paypal in our services vault (cc @kewisch : can move the test account into the services vault?)

 If all goes well, you'll see a payment complete / payment verified screen (depends on timing) and be redirected to dashboard at the end.

### DIFF
--- a/assets/app/vue/views/SubscribeView/components/CheckoutStep.vue
+++ b/assets/app/vue/views/SubscribeView/components/CheckoutStep.vue
@@ -1,9 +1,9 @@
 <script setup lang="ts">
 import { useI18n } from 'vue-i18n';
 import { LoadingSkeleton, NoticeBar, NoticeBarTypes, VisualDivider } from '@thunderbirdops/services-ui';
-import { initializePaddle, PaddleEventData, CheckoutEventNames, CheckoutEventsStatus } from '@paddle/paddle-js';
+import { initializePaddle, PaddleEventData, CheckoutEventNames } from '@paddle/paddle-js';
 import CardContainer from '@/components/CardContainer.vue';
-import { ref } from 'vue';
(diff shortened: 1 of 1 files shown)


assets/app/vue/views/SubscribeView/components/CheckoutStep.vue
```diff
@@ -18,12 +18,14 @@ const initCurrencyFormatter = (code: string) => {
 
 let currencyFormatter = initCurrencyFormatter('USD');
 let paddle = null;
+let doneCheckerHandler = null;
```
davinotdavid: I believe we should clear this timeout during unmount, something like:

```
onUnmounted(() => {
  if (doneCheckerHandler) {
    clearTimeout(doneCheckerHandler);
  }
});
```

PR #639, opened 2026-03-12
Further improvements to our third-party payment processor flow.

MelissaAutumn:
Related issue #458 

* Brings back successUrl and the backend route
* If the payment type was Paypal (or another pop-up window based type) then redirect them to a window.close template.
* If the payment was not a pop-up window based type simply redirect them.
* Added and enhanced plenty of structural window timeouts -.-.

I discovered a weird issue with new accounts and the csrftoken, but I need to test on stage to make sure I'm not chasing a weird dev issue.


davinotdavid (commented):
Other than the tests that seems to be require updating, lgtm!

PR #683, opened 2026-03-30
Revert "Revert "Remove the old Celery cluster""

ryanjjung:
Reverts thunderbird/thunderbird-accounts#681, which reverted #680. Therefore, this PR is the same as #680, and will lead to the destruction of the old Celery cluster, which has no tasks currently running.


MelissaAutumn (approved):
lgtm!

Issue #658, opened 2026-03-23
Thundermail dashboard rework


malini:
### "Connect to Thunderbird" setup step in Thundermail account page - Scope of Work

🗓️Deadline: Apr 10, 2026
✔️Approval to add to the Roadmap: Mar 30, 2026

## 🎯 The objective

The aim of this project is to support the setup of a Thundermail account by addressing the First Time User Experience (FTUE), in order to reduce its friction for subscribed TB Pro users. Success will be measured by achieving a higher percentage of active Thundermail users.

The idea is to provide a direct connection between the Thundermail Dashboard and the Thunderbird Desktop application through a "Connect to Thunderbird" button that would use a custom protocol mapping to pass token information and trigger the Account Setup in Thunderbird Desktop. This will require us to agree on the URL format and also close collaboration with the Desktop team to synchronize timelines on the OS protocol registration work.

### 🎒 Resources

- [Figma file with the current hi-fi explorations](https://www.figma.com/design/6Xn9ltELZMhY5wXPBzlayX/tMail_web_2026-Q2_tMailFirst_v02?node-id=2011-14&p=f&t=dWQUctCCh3pdbGB5-0)
- [TB Desktop - Account Hub Services Integration Notion Page](https://www.notion.so/mzthunderbird/Account-Hub-Services-integration-31f2df5d45ae801195b9c22aac0db51b?source=copy_link)
- [Phabricator issue for the Custom Mapping Protocol in TB Desktop](https://phabricator.services.mozilla.com/D294120)
- [Bugzilla issue for TB Desktop - Make use of net.thunderbird URLs for OAuth redirects](https://bugzilla.mozilla.org/show_bug.cgi?id=2033653)

### 🟢 In Scope

- One click setup of Thundermail, triggered from the Thundermail dashboard, in Thunderbird Desktop across all OSes.
- Implement the new "Welcome to Thundermail" card UI/UX
  - Desktop
    - Button to launch Thunderbird's Account Hub with Thundermail information prefilled
    - In case no Thunderbird is installed, show information that will explain to the user how to download Thunderbird
    - Fallback: If Desktop has not yet delivered the account hub protocol handler, show fallback page/information.
  - Mobile
    - QR code allows users to set up their account in Thunderbird for Android.
    - [Design/Product] Review mobile happy path to ensure that the QR code actually makes the process easier. Adjust engineering scope as necessary.
[...]

PR #689, opened 2026-04-02
Mail Dashboard UI Revamp

davinotdavid:
## Description of changes 

- Small refactor of `CardContainer` component for better future maintainability
- Added `SegmentedControlSlider` component (great candidate for a services-ui component in the future but leaving it here for now as I am not aware at any other place that uses it)
- Added `GetStartedWithThundermail` component alongside each tab in its own separate component
- Moved `View Server Settings` into `Email Settings`
- Split `ViewServerSettings` into `ServerSettingsCard` and `ServerSettingsCardItem` components so that it can be used both in `View Server Settings` and in `Get Started With Thundermail`
- Removed QR code modal and the whole "Set Up Email App" section we had before, alongside related files
- Implemented pin functionality to `Get Started With Thundermail`
- Updated the tooltip values for the `What is ...` for IMAP, JMAP and SMTP
- Updated App Password UI text + tool-tip copy
- Minor styling changes to better approximate the Zeplin designs

Zeplin:
https://zpl.io/o0Y7dyG

## Screenshots / GIFs

Get Started with Thundermail - Desktop
image

Get Started with Thundermail - Mobile
image

Get Started with Thundermail - Other Apps
image

Full Page
image

## Known issues / Things to improve

- Still need to update the tour (that popup on the top right that guides through the sections) / `useTour`
- Decide on what to do in the interim about the "Connect" button while we don't have the custom protocol mapping yet
- Header navigation changes (removal of the Manage / Thundermail) and redirecting to Thundermail instead of Accounts after payment will be done in a separate PR

## Related issues

Related https://github.com/thunderbird/thunderbird-accounts/issues/658

PR #694, opened 2026-04-10
FTUE Mail Dashboard Tour Cards

davinotdavid:
## Description of changes

- Updating the tour cards with the new copy, positioning and small UI tweaks (box-shadow, spacing inside the card)
- Added better support for screen readers and pressing ESC now closes the tour

Zeplin :https://zpl.io/romO4k3

Note that the initial and final cards are anchored to the header / user avatar where as the rest of the cards are positioned relative to their respective sections.

## Screenshots / Video

https://github.com/user-attachments/assets/1b1273ff-a67c-4d8a-9102-2e6042d60ed2

## How to test

The show / hide tour guide is controlled by a local storage key called `tb_accounts_ftue_completed` so make sure that you don't have that as `true` if you want to see it!

## Related issues

Related https://github.com/thunderbird/thunderbird-accounts/issues/658

### DIFF
--- a/assets/app/vue/composables/useTour.ts
+++ b/assets/app/vue/composables/useTour.ts
@@ -1,4 +1,4 @@
-import { ref, watch } from 'vue';
+import { ref, watch, nextTick } from 'vue';
 
 export const FTUE_STEPS = {
   INITIAL: 0,
@@ -10,6 +10,7 @@ export const FTUE_STEPS = {
 }
 
 const FTUE_STORAGE_KEY = 'tb_accounts_ftue_completed';
+const TOUR_CARD_SELECTOR = '[role="dialog"][tabindex="-1"]';
 
 // Get the initial state of the FTUE from localStorage or return true if not available
 const getInitialState = () => {
@@ -23,12 +24,31 @@ const getInitialState = () => {
 const currentStep = ref(FTUE_STEPS.INITIAL);
 const showFTUE = ref(getInitialState());
 
-// When the showFTUE state changes to false, we mark it the FTUE as completed in localStorage
+const onEscapeKey = (event: KeyboardEvent) => {
+  if (event.key === 'Escape') {
+    showFTUE.value = false;
+  }
+};
+
+const focusTourCard = () => {
+  nextTick(() => {
+    const card = document.querySelector<HTMLElement>(TOUR_CARD_SELECTOR);
+    card?.focus();
+  });
+};
+
 watch(showFTUE, (value) => {
-  if (!value && typeof localStorage !== 'undefined') {
+  if (value) {
+    document.addEventListener('keydown', onEscapeKey);
+    return;
+  }
+
+  document.removeEventListener('keydown', onEscapeKey);
+
+  if (typeof localStorage !== 'undefined') {
     localStorage.setItem(FTUE_STORAGE_KEY, 'true');
   }
-});
+}, { immediate: true });
 
 export const useTour = () => {
   const steps = [
@@ -78,7 +98,7 @@ export const useTour = () => {
     showFTUE.value = false;
   };
 
-  // Auto-scroll to target when step changes
+  // Auto-scroll to target and move focus when step changes
   watch(currentStep, (newStep) => {
     if (!showFTUE.value) return;
 
@@ -87,6 +107,7 @@ export const useTour = () => {
 
     if (newStep === FTUE_STEPS.INITIAL || newStep === FTUE_STEPS.FINAL) {
       window.scrollTo({ top: 0, behavior });
+      focusTourCard();
       return;
     }
 
@@ -99,6 +120,8 @@ export const useTour = () => {
         element.scrollIntoView({ behavior, block: 'center' });
       }
     }
+
+    focusTourCard();
   });
 
   return {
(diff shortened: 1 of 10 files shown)


assets/app/vue/views/MailView/sections/EmailSettingsSection/index.vue
```diff
@@ -29,19 +26,10 @@ export default {
...
-        :current-step="tour.currentStep.value"
-        :total-steps="FTUE_STEPS.FINAL"
-        show-back
-        @next="tour.next()"
-        @back="tour.back()"
-        @close="tour.skip()"
-      />
-
+    <card-container
+      id="email-settings"
```
Sancus: Maybe this is some Vue stuff I don't understand, but two elements shouldn't have the same id?

assets/app/vue/views/MailView/sections/CustomDomainsSection/index.vue
```diff
@@ -32,6 +32,10 @@ const customDomainFormRef = ref(null);
 const verificationCriticalErrors = ref<string[]>([]);
 const maxCustomDomains = window._page?.maxCustomDomains;
 
+const nextStepText = computed(() => {
+  return t('views.mail.ftue.nextStep', { step: t('views.mail.ftue.yourAccount') });
+})
```
Sancus:
```suggestion
});
```

There's a few more places missing semicolons which I will leave as an exercise to the reader :)

assets/app/vue/composables/useTour.ts
```diff
@@ -21,34 +24,60 @@ const getInitialState = () => {
...
       id: FTUE_STEPS.CUSTOM_DOMAINS,
       targetId: 'custom-domains',
     },
     {
       id: FTUE_STEPS.FINAL,
-      targetId: 'email-settings',
     }
   ];
 
   const start = () => {
```
Sancus: Is this actually used anywhere???

assets/app/vue/composables/useTour.ts
```diff
@@ -1,13 +1,16 @@
...
   INITIAL: 0,
-  DASHBOARD: 1,
-  CUSTOM_DOMAINS: 2,
-  FINAL: 3,
+  CONNECT_EMAIL: 1,
+  APP_PASSWORDS: 2,
+  EMAIL_ALIASES: 3,
+  CUSTOM_DOMAINS: 4,
+  FINAL: 5,
 }
```
Sancus: This benefits from `as const` I think.

assets/app/vue/views/MailView/sections/DashboardSection/components/GetStartedWithThundermail.vue
```diff
@@ -53,13 +60,26 @@ export default {
...
   <card-container
+    id="connect-email"
     :title="t('views.mail.sections.dashboard.getStartedWithThundermail.title')"
     :subtitle="t('views.mail.sections.dashboard.getStartedWithThundermail.description')"
     is-pinnable
     :is-pinned="isPinned"
     @toggle-pinned="$emit('togglePinned')"
   >
     <div class="get-started-with-thundermail-content">
+      <tour-card
```
Sancus: This same tour-card block is repeated a lot of times. Seems like it could use a wrapper component? Or maybe there's a different way to reduce the repetition, I'm not a Vue expert.

assets/app/vue/views/MailView/index.vue
```diff
@@ -81,24 +103,31 @@ export default {
   font-size: 0.75rem;
 }
 
-.initial-welcome {
+.header-card {
```
Sancus: Shouldn't this be a tour-card? Seems like it has a lot of redundancies with the tour card CSS.

assets/app/vue/composables/useTour.ts
```diff
@@ -1,13 +1,16 @@
...
 
 const FTUE_STORAGE_KEY = 'tb_accounts_ftue_completed';
+const TOUR_CARD_SELECTOR = '[role="dialog"][tabindex="-1"]';
```
Sancus: We're selecting tour cards by their role and tab index? That seems a little wild. Why not a data-tour-card attribute or something? Feels like this could easily break in the future.

Sancus (changes requested):
Looks pretty good overall although I do wonder if there should be more visual connection between the cards and the section they're referencing. It feels a little disjointed to me right now.

That feedback is not a blocker of this PR though. Other than that, review comments are below.

assets/app/vue/views/MailView/sections/EmailSettingsSection/index.vue
davinotdavid: Whoops, not Vue stuff, just me forgetting to clean that up 😅 Good catch!

assets/app/vue/composables/useTour.ts
davinotdavid: Not currently no, I might be misremembering but I think there was some conversation before to be able to re-trigger the tour through the UI. Removed!

assets/app/vue/views/MailView/sections/DashboardSection/components/GetStartedWithThundermail.vue
davinotdavid: Ah yes there is and I've just used this [last week in this PR](https://github.com/thunderbird/thunderbird-accounts/pull/689/changes#diff-325e249050db6929b6373cb36d7230283210e33dda944363c51941d965ffca5bR38-R43) (and seemingly forgotten to make this better, so thanks for the reminder!).

For future reference, we can use the [Teleport feature in Vue](https://vuejs.org/guide/built-ins/teleport)!

Sancus (approved):
LGTM now.

PR #695, opened 2026-04-10
Early Bird Navigation Updates

### LINKED ISSUE #701: Make the Thundermail dashboard the first page after login/payment
When logging in we currently go to https://accounts.tb.pro/ . With the focus on Thundermail we'd like that initial page to be https://accounts.tb.pro/mail so they can see the Thundermail settings and our welcome onboarding.

This is true both for the step after paddle payment, but also when logging in to accounts.tb.pro.

### LINKED ISSUE #702: Remove getting started box from account dashboard
Remove the old getting started box from https://accounts.tb.pro/

Image

davinotdavid:
## Description of changes

- Removed `GetStartedCard` from the Accounts dashboard
- Removed header & footer navigation links
- Made `UserAvatar` in Mail dashboard to go to Accounts dashboard instead of showing the dropdown (kept the dropdown for Accounts dashboard so that the user can logout)
- Changed the root path to be Mail instead of Accounts. Also updated the `KC_TBPRO_HOME` to be `/` so that we can simply update the `router.ts` file in the future (mimicking what we have in the dev setup in `docker-compose.yml` as well).

Zeplin: https://zpl.io/o0Y7dyG

## Related issues

Related https://github.com/thunderbird/thunderbird-accounts/issues/658
Fixes #702
Fixes #701


devmount (approved):
Ahhh that's a nice diff ratio again:

Image

Thanks for updating this, lgtm!

kewisch:
> * Made `UserAvatar` in Mail dashboard to go to Accounts dashboard instead of showing the dropdown (kept the dropdown for Accounts dashboard so that the user can logout)

I think the UserAvatar can stay as is, or am I misunderstanding? We should retain the possibility to log out and go to the account via the dropdown, as before.

davinotdavid:
@kewisch hmmm given that the behaviour that in Zeplin was very recent, I thought it was decided to change it into a link to Accounts and then, in Accounts, show the dropdown to logout. IMHO I also think we should maintain the dropdown but asked for confirmation here: https://zpl.io/jpkOAmq

davinotdavid:
Confirmed with @laurelterlesky and @solangevalverde earlier today that the PR is what's intended UX-wise for now.

As we implement other sections / different pages, then we will potentially put the dropdown back if need be. Merging this then!

PR #696, opened 2026-04-11
Add Keycloak event polling for PostHog account activity metrics

Sancus:
A Celery beat task (using celery-redbeat to ensure only one worker at a time runs it) runs every 15 minutes. It polls the Keycloak Admin API for recent authentication events (logins, logouts, token operations, registrations) within the polling window.

For each event that has a userId, it submits to PostHog via a small telemetry module. The distinct_id is sha256(keycloak_uuid),to provide a unique id without exposing any PII. Only an allowlisted set of properties is sent: clientId, keycloak_event_type, keycloak_event_id, and is_error. Everything else from the Keycloak event (IP addresses, session details, etc.) is discarded.

Events without a userId (e.g. failed logins before user identification) are skipped.


keycloak/data/import/tbpro-realm.json
```diff
@@ -1500,9 +1500,10 @@
   "accountTheme" : "",
   "adminTheme" : "",
   "emailTheme" : "",
-  "eventsEnabled" : false,
+  "eventsEnabled" : true,
```
Sancus, line 1503: I know this doesn't affect the currently running instances but figured it might as well be in there. I can take it out if preferred. Might be good for local setups though? idk.

scripts/entry.sh
```diff
@@ -13,7 +13,7 @@ fi
 
 # Run the app with the appropriate command
 if [[ "$TBA_CELERY" == "yes" ]]; then
-    CMD="uv run celery -A thunderbird_accounts worker -l INFO"
+    CMD="uv run celery -A thunderbird_accounts worker -l INFO -B"
```
MelissaAutumn:
```suggestion
    CMD="uv run celery -A thunderbird_accounts worker -l INFO"
```

Unless there's a good reason to use bpython, please don't sneak it in. Thanks!

keycloak/data/import/tbpro-realm.json
MelissaAutumn: It's fine we should ensure our local setup matches at least stage. This won't apply on existing realms though, but it's not hard to follow along to these changes or blow away your db.

pulumi/config.prod.yaml
```diff
@@ -398,6 +400,8 @@ resources:
                   value: "yes"
                 - name: TBA_FLOWER
                   value: "no"
+                - name: POSTHOG_HOST
+                  value: "https://us.i.posthog.com"
```
MelissaAutumn, line 404: ...we should migrate to the eu server.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+
+import sentry_sdk
+from celery import shared_task
+from django.conf import settings
+from django.core.cache import cache
+
+from thunderbird_accounts.authentication.clients import KeycloakClient
+from thunderbird_accounts.telemetry import capture, flush
+
+logger = logging.getLogger(__name__)
```
MelissaAutumn, line 12: I believe we currently use global logger iirc. This is fine but may not be properly logged, so please test this.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    'LOGOUT': 'accounts.logout',
+    'CODE_TO_TOKEN': 'accounts.activity',
+    'CODE_TO_TOKEN_ERROR': 'accounts.activity',
+    'INTROSPECT_TOKEN': 'accounts.activity',
+    'REFRESH_TOKEN': 'accounts.activity',
+}
+
+SEEN_EVENTS_CACHE_KEY = 'keycloak_poll:seen_event_ids'
+SEEN_EVENTS_CACHE_TTL = 3600
+EVENTS_PAGE_SIZE = 500
```
MelissaAutumn: These should probably live in settings.py

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    'LOGIN': 'accounts.login',
+    'LOGIN_ERROR': 'accounts.login_error',
+    'REGISTER': 'accounts.register',
+    'REGISTER_ERROR': 'accounts.register_error',
+    'LOGOUT': 'accounts.logout',
+    'CODE_TO_TOKEN': 'accounts.activity',
+    'CODE_TO_TOKEN_ERROR': 'accounts.activity',
+    'INTROSPECT_TOKEN': 'accounts.activity',
+    'REFRESH_TOKEN': 'accounts.activity',
+}
```
MelissaAutumn: These should probably live in settings.py

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+
+@shared_task(bind=True, retry_backoff=True, retry_backoff_max=300, max_retries=3)
+def poll_keycloak_events(self):
+    """Poll the Keycloak Admin API for recent user events and submit them to PostHog.
+
+    Keycloak's eventsExpiration is set short (30 min) to keep the events table small.
+    This task runs every 15 minutes, deduplicating via a cached set of seen event IDs.
+    Each event is hashed, stripped of PII, and submitted directly to PostHog.
+    """
+    posthog_api_key = getattr(settings, 'POSTHOG_API_KEY', '')
```
MelissaAutumn: This is not how django settings works? 


```suggestion
    posthog_api_key = settings.POSTHOG_API_KEY
```

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    if not posthog_api_key:
+        logger.debug('POSTHOG_API_KEY not configured, skipping Keycloak event polling')
+        return {'task_status': 'skipped', 'reason': 'POSTHOG_API_KEY not configured'}
+
+    now = datetime.now(timezone.utc)
+    today = now.strftime('%Y-%m-%d')
+
+    poll_interval = int(getattr(settings, 'KEYCLOAK_EVENT_POLL_INTERVAL_SECONDS', 900))
```
MelissaAutumn:
```suggestion
    poll_interval = settings.KEYCLOAK_EVENT_POLL_INTERVAL_SECONDS
```

src/thunderbird_accounts/settings.py
```diff
@@ -467,6 +467,23 @@ def oidc_logout(request):
...
 
+# Celery Beat — use redbeat for distributed locking across multiple workers.
+# redbeat_redis_url defaults to broker_url, so no separate config needed.
+CELERY_BEAT_SCHEDULER = 'redbeat.RedBeatScheduler'
+
+CELERY_BEAT_SCHEDULE = {}
+
+KEYCLOAK_EVENT_POLL_INTERVAL_SECONDS = int(os.getenv('KEYCLOAK_EVENT_POLL_INTERVAL_SECONDS', '900'))
+
+POSTHOG_API_KEY = os.getenv('POSTHOG_API_KEY', '')
```
MelissaAutumn: This should really be None if it doesn't exist.

src/thunderbird_accounts/telemetry/__init__.py
```diff
@@ -0,0 +1,66 @@
+import hashlib
```
MelissaAutumn: Move this out of __init__ into a file. I'd like to keep init's clean of code unless required (99% that means imports only.)

src/thunderbird_accounts/telemetry/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def _get_client() -> Posthog | None:
+    """Return a singleton PostHog client, or None if not configured."""
+    global _client  # noqa: PLW0603
+    if _client is not None:
+        return _client
+
+    api_key = getattr(settings, 'POSTHOG_API_KEY', '')
+    host = getattr(settings, 'POSTHOG_HOST', 'https://us.i.posthog.com')
```
MelissaAutumn: We have defaults in settings.py so we can just call settings.POSTHOST_API_KEY for example.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    try:
+        submitted = 0
+        skipped = 0
+        for event in all_events:
+            event_id = event.get('id', '')
+            if event_id in seen_event_ids:
+                skipped += 1
+                continue
+
+            kc_type = event.get('type', 'UNKNOWN')
```
MelissaAutumn: We should probably const that default type somewhere.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+            user_id = event.get('userId', '')
+            event_name = EVENT_MAP.get(kc_type, 'accounts.activity')
```
MelissaAutumn: ditto

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+        submitted = 0
+        skipped = 0
+        for event in all_events:
+            event_id = event.get('id', '')
+            if event_id in seen_event_ids:
+                skipped += 1
+                continue
+
+            kc_type = event.get('type', 'UNKNOWN')
+            user_id = event.get('userId', '')
```
MelissaAutumn: We should default to None here.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+        all_events = [e for e in raw_events if e.get('time', 0) >= cutoff_ms and e.get('type') in EVENT_MAP]
+
+    except Exception as exc:
+        sentry_sdk.capture_exception(exc)
+        logger.error(f'Failed to fetch Keycloak events: {exc}')
+        raise self.retry(exc=exc)
+
+    if not all_events:
+        logger.info('No Keycloak events found in polling window')
+        return {'task_status': 'success', 'events_submitted': 0}
```
MelissaAutumn: Should normalize it with the return value at the end.

src/thunderbird_accounts/telemetry/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    client.capture(distinct_id=distinct_id, event=event, properties=event_properties)
+
+
+def flush():
+    """Flush any pending events and reset the client."""
+    global _client  # noqa: PLW0603
+    if _client is not None:
+        _client.flush()
```
MelissaAutumn: Fyi, their docs mention to not use flush()
https://posthog.com/docs/references/posthog-python#PostHog-flush

src/thunderbird_accounts/telemetry/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+from django.conf import settings
+from posthog import Posthog
+
+logger = logging.getLogger(__name__)
+
+_client = None
+
+
+def _get_client() -> Posthog | None:
```
MelissaAutumn: Why the global when `@cache` exists?

MelissaAutumn (changes requested):
I arise from my migraine cave after sensing a disturbance in my github notifications. (I need to ban github at my router level when I'm sick.) 

Thanks for handling this! I believe we'll also need celery beat on prod as I think it's only implemented on dev right now. See https://github.com/thunderbird/thunderbird-accounts/issues/697

scripts/entry.sh
Sancus: That's only bpython on the shell. On the worker command, it's --beat or -B, which turns the Beat scheduler on. I could change it to --beat to be more explicit but then we should probably change all the arguments to the lowercase verbose form. (Edit: I went ahead and did this, idk who has these argument meanings memorized anyway)

pulumi/config.prod.yaml
Sancus: Hmm yeah I guess. Ugh. That requires creating an entire new org I think. Not sure it's even worth it right now... the plan is no PII in posthog whatsoever so I'm not super concerned about privacy stuff. There is a latency penalty, but doesn't seem important for telemetry.

src/thunderbird_accounts/authentication/tasks.py
Sancus: true

src/thunderbird_accounts/telemetry/__init__.py
Sancus: Ahha. I will call shutdown instead, then. I don't think there's any reason to leave the batching up to the SDK, we just want everything delivered as soon as this task is complete.

src/thunderbird_accounts/telemetry/__init__.py
Sancus: I'm just not used to @cache tbh.

src/thunderbird_accounts/authentication/tasks.py
Sancus: This is actually a vestige of a previous implementation, the earlier filter means that only types in EVENT_MAP can even end up in this list, so None is impossible.

https://github.com/thunderbird/thunderbird-accounts/pull/696/changes/BASE..0a4f417fd6a921d104923586e422540f4b184d11#diff-5a83112cb4c722010fc6c5fcaa4ba3cbdaedd2812515c5bcd8b04db8845ebad2R84

Will remove.

src/thunderbird_accounts/authentication/tasks.py
Sancus: Events can only be a type listed in `EVENT_MAP` so the default also doesn't do anything.

src/thunderbird_accounts/authentication/tasks.py
Sancus: We have `"disable_existing_logging": False` in settings so module level loggers should work. Also they are the standard django recommendation so I'm inclined to just keep this... but I don't care too much either way. In any case I tested the logging and it works as expected.

pulumi/config.prod.yaml
Sancus: I'm punting on this for now.

src/thunderbird_accounts/settings.py
Sancus: Removed the '' so this will now return None if it doesn't exist.

davinotdavid (approved):
oh sorry I was late for the review but lgtm!

since Mel reviewed first, let's wait for a re-review before merging this

scripts/entry.sh
MelissaAutumn: This is why i don't review with a migraine...thanks!

MelissaAutumn (approved):
tyvm!

Sancus:
Before I merge this I'm going to test that the beat-dedupe stuff work as the docs say it should with multiple celery containers running. It SHOULD but I've never used celery-redbeat before.

docker-compose.yml
```diff
@@ -78,6 +78,13 @@ services:
   celery:
     <<: *accounts
     ports: []
+    # Workaround for embedded beat (--beat) spinning at 100% CPU in Docker.
+    # Docker inherits a huge SC_OPEN_MAX (~1B) causing close_open_fds() to loop.
+    # https://github.com/celery/celery/issues/8306
+    ulimits:
+      nofile:
+        soft: 10000
+        hard: 10000
```
MelissaAutumn, line 87: Since we don't use these in stage/prod we might want to monitor what the values are and figure out a way to pulumi these up.

Sancus: Fargate's defaults for these are 65K so it's not an issue, I believe.

Sancus: `SC_OPEN_MAX: 65535	RLIMIT_NOFILE soft: 65535 hard: 65535`

yep we're good

PR #709, opened 2026-04-14
ToS / Privacy Policy Feature

### LINKED ISSUE #705: Create Terms of Service / Privacy Policy page
[Copied over from the parent]

### 🟢 In Scope

- Route the user to the acceptance screen after accounts dashboard login / registration
- Check on critical endpoints if the user has accepted the terms, and if not either log them out or route them back to the terms.

### 🔴 Out of Scope

We’ll need most of this later, but excluding this from scope for now.

- Enforce terms acceptance for already logged in sessions on a per-product basis, for example:
	- Accept terms for logged in users before they use Send to upload a file
	- Accept terms for appointment users before they accept/decline a meeting
	- Disallow sending email by responding with a mailer-daemon type response that they need to accept the terms.
- Banner to ask the user to review new terms before we require them to accept it.
- Mechanism to send email to the user to notify them of substantial changes that need acceptance, and to notify them the grace period has ended and we’ll limit their functionality.

### 🚧 Technical and/or Business Requirements

- The system must maintain a unique timestamp (UTC) when the user has accepted the terms
- With acceptance, it must be stored which version of the terms/privacy policy have been accepted, along with a clear mapping so it cannot be confused which version has been accepted.
- If there is a source context, this would also be worth saving (e.g. they were logging in to the accounts dashboard, or a specific service, etc)
- When the user has the chance to accept the terms they have the following actions:
	- If the user has not yet paid: Decline and cancel registration, or decline and delete account (confirm)
	- If the user is currently paying/subscribed: Decline and cancel account (confirm)
	We may offer them a pro-rated refund in this case if required, though that might have to be manual for the start.
	- Accept terms and privacy policy
	- Contact support
- The system should be designed in a way that we can verify they saw the terms of service before they continue, in a way that we could defend this in a court if needed.
- If a user accepts the terms in another tab, reloading the first tab should detect acceptance and continue without further interaction
- If the user declines the terms, we should record the timestamp (UTC) when they have declined. We may need to gate features/analysis on declining terms.

Zeplin: https://zpl.io/m81o56m

### LINKED ISSUE #706: Implement "Decline and cancel account" button
Based on the [Zeplin designs](https://zpl.io/m81o56m), if the user is already subscribed and signed in but haven't accepted the latest legal documents, we should allow them to decline and cancel the account.

This is currently a technical challenge and it may involve a cascading effect across services (Appointment, Send, etc). Since this is quite a destructive action, I've raised this to the Design team to confirm what should happen after clicking said button.

### LINKED ISSUE #707: Port current legal documents (Terms of Service / Privacy)
Based on the current implementation proposal, we will be able to add / edit / delete the legal documents through Django admin (for better flexibility) so first we need to port these over.

It would be great to also determine the i18n flow!

davinotdavid:
## Description of changes

Introducing the 🌟 Terms of Service / Privacy Policy 🌟 feature!

[Frontend]
- A new `/tos-privacy` page that requests the current version of the Terms of Service and Privacy markdown text and presents an option to accept / decline
- Upon accepting, we make a record in the database
- Upon declining, we also make a record in the database and logs out from Django / Keycloak
  - Declining and signing in again brings the legal page again for a brand new opportunity to accept the terms
- Every route is now gated by a `needsTosAcceptance` global variable that is set from the backend, if it is `true` you can only see that page

[Backend]
- A new `legal` Django app that is hooked into Django admin for toggling which document is the current version
- An `en.md` version for the Terms of Service and Privacy Policy pre-seeded in a migration based off of [this comment (should match the one in the tb.pro website as well)](https://github.com/thunderbird/thunderbird-website/issues/1123#issuecomment-3871919346)
  - The content is source-controlled in markdown so that it can be openly available for people to see instead of in the database
  - i18n can be achieved by creating a `de.md` file and requesting with a `locale`
- In the `home` view, we check to see if the authenticated user needs to accept the most current terms, changing the `needs_tos_acceptance` flag accordingly

## Screenshots

image

- At the time of this PR, we are still deciding on what / how to off-board a user when they decide to decline the terms so for now it records the decline decision and logs out the user. If they decide to log in again, they can and are prompted with the terms again.
- Top navigation will be changed when [this other PR](https://github.com/thunderbird/thunderbird-accounts/pull/695) lands but even if one clicks on links there, it doesn't navigate away.

## Related issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/705
Fixes https://github.com/thunderbird/thunderbird-accounts/issues/707
Fixes https://github.com/thunderbird/thunderbird-accounts/issues/706


assets/app/vue/router.ts
```diff
@@ -158,21 +164,36 @@ router.beforeEach((to, _from) => {
...
 
-  // Don't let unsubscribed users anywhere except the subscribe view
-  if (isAuthenticated && sendToSubscribe && !['subscribe', 'contact'].includes(to.name.toString())) {
+  // --- Authenticated users below ---
+  // Guards are ordered by priority: TOS > subscription > normal access.
+
+  const allowedFor = {
+    tos: new Set(['tos-privacy', 'contact']),
+    subscription: new Set(['subscribe', 'contact', 'tos-privacy']),
+  };
```
MelissaAutumn, line 185: Why sets? Arrays should do here.

src/thunderbird_accounts/core/views.py
```diff
@@ -111,6 +113,19 @@ def home(request: HttpRequest):
...
+        current_docs = LegalDocument.objects.filter(is_current=True)
+        if current_docs.exists():
+            accepted_doc_ids = set(
+                LegalDocumentResponse.objects.filter(
+                    user=request.user,
+                    document__in=current_docs,
+                    action=LegalDocumentResponse.Action.ACCEPTED,
+                ).values_list('document_id', flat=True)
+            )
+            needs_tos_acceptance = current_docs.exclude(uuid__in=accepted_doc_ids).exists()
```
MelissaAutumn: Could this instead be:
```python
legal_doc_count = LegalDocument.objects.filter(is_current=True).count()
needs_tos_acceptance = LegalDocument.objects.filter(is_current=True, legal_document_response__user=user=request.user, legal_document_response__action=LegalDocumentResponse.Action.ACCEPTED).count() == legal_doc_count
```

(You may need to adjust the relationship spanning filters, see here: https://docs.djangoproject.com/en/5.2/topics/db/queries/#lookups-that-span-relationships)

src/thunderbird_accounts/legal/models.py
```diff
@@ -0,0 +1,75 @@
...
+        help_text=_('Relative path to content directory under assets/legal/, e.g. "tos/v1.0"'),
+    )
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=['document_type', 'version'], name='unique_document_type_version'),
+        ]
+        indexes = [
+            *BaseModel.Meta.indexes,
+            models.Index(fields=['document_type', 'is_current']),
```
MelissaAutumn: This will create a composite index I believe. We'll probably want is_current by itself though.

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+    return _record_response(request, LegalDocumentResponse.Action.ACCEPTED)
+
+
+@login_required
+@require_http_methods(['POST'])
+def decline_legal_docs(request):
+    _record_response(request, LegalDocumentResponse.Action.DECLINED)
+
+    # TODO: Offboard the user / Delete user data (?)
+
```
MelissaAutumn, line 115: Do we have a ticket for offboarding from tos?

MelissaAutumn (changes requested):
Thanks! Seems like a lot of effort was put into this. 

However, we should parse and check in the legal html from the markdown. We can also store the markdown in repo, but we shouldn't be parsing anything on the fly as it's slow and depending on the security vulnerability of the week might lead to some malicious parsing. 

This would be similar to how appointment handles it I believe.

assets/app/vue/router.ts
davinotdavid: Premature optimization here that now that you've mentioned it is indeed unlikely to make us suffer given the number of elements but originally added because `.has` would perform better than `.includes` as we are doing this on `router.beforeEach`. Will change it to arrays, thanks!

src/thunderbird_accounts/legal/views.py
davinotdavid: Yes (although the name of the ticket is not great):

https://github.com/thunderbird/thunderbird-accounts/issues/706

src/thunderbird_accounts/core/views.py
davinotdavid: Ah nice, that's a lot better, thanks!

davinotdavid:
@MelissaAutumn thank you for the review! Makes sense to serve the HTML directly so now I've:

- Added a command to convert the legal docs from MD to HTML
- Added documentation around how to add / update the legal docs in README.md
- Serving the HTML directly instead and removed any markdown parsing from the frontend

assets/app/vue/router.ts
MelissaAutumn: I think sets only beat out arrays if we're using really large datasets. I could be wrong, but thanks for the change!

assets/app/vue/views/TosPrivacyView/index.vue
```diff
@@ -0,0 +1,286 @@
...
+const router = useRouter();
+const route = useRoute();
+
+const declineButtonLabel = window._page?.hasActiveSubscription ? t('views.tosPrivacy.declineAfterPayment') : t('views.tosPrivacy.declineButton');
+
+const tosSelectedTab = ref<LEGAL_DOCUMENT_TYPES>(LEGAL_DOCUMENT_TYPES.TOS);
+const loading = ref(true);
+const accepting = ref(false);
+const declining = ref(false);
+const error = ref('');
```
MelissaAutumn: I'd prefer if "no value" is null so we can distinguish between the value is a literal empty string vs definitely no errors here.

assets/app/vue/views/TosPrivacyView/index.vue
```diff
@@ -0,0 +1,286 @@
...
+const sourceContext = computed(() => {
+  return (route.query.source as string) || 'accounts-dashboard';
+});
+
+async function handleAccept() {
+  accepting.value = true;
+  error.value = '';
+  try {
+    await acceptLegalDocs(sourceContext.value);
+    window._page.needsTosAcceptance = false;
```
MelissaAutumn: I'd rather we treated window._page as immutable. This might just warrant a location.href change to re-serve the page.

...as annoying as that sounds. The other option is we have a way to slice in a new window._page. 😅

assets/app/vue/views/TosPrivacyView/index.vue
```diff
@@ -0,0 +1,286 @@
...
+      <div class="content legal-content">
+        <div
+          v-if="tosSelectedTab === LEGAL_DOCUMENT_TYPES.TOS"
+          v-html="renderedTosHtml"
+        />
+        <div
+          v-else
+          v-html="renderedPrivacyHtml"
+        />
+      </div>
```
MelissaAutumn, line 131: Hmm we can potentially avoid v-html with the use of iframes pointed at our backend. Can you give that a quick try, and if it's too much trouble we can drop it in favour of this.

src/thunderbird_accounts/legal/management/commands/convert_legal_docs.py
```diff
@@ -0,0 +1,48 @@
...
+    help = 'Converts legal document markdown files under assets/legal/ to HTML.'
+
+    def handle(self, *args, **options):
+        legal_dir = settings.ASSETS_ROOT / 'legal'
+
+        if not legal_dir.exists():
+            self.stderr.write(self.style.ERROR(f'Legal assets directory not found: {legal_dir}'))
+            return
+
+        md = markdown.Markdown(extensions=['tables'])
```
MelissaAutumn:
```suggestion
        md = markdown.Markdown(extensions=['attr_list', 'tables'])
```

We don't use it but MoCo legal does for some super minor stuff. If we ever end up pulling from their legal repo then we should have this just in case. 

https://python-markdown.github.io/extensions/attr_list/

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+    """Read pre-rendered HTML content for a legal document, falling back to English."""
+    safe_locale = Path(locale).name
+    content_dir = settings.ASSETS_ROOT / 'legal' / content_path
+    content_file = content_dir / f'{safe_locale}.html'
+
+    if not content_file.exists():
+        content_file = content_dir / 'en.html'
+
+    if not content_file.exists():
+        return ''
```
MelissaAutumn: I feel we should actually error out here, as accepting a blank TOS/PP might cause some weird legal problems. 

Any thoughts @kewisch ?

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+from django.contrib.auth.decorators import login_required
+from django.http import JsonResponse
+from django.views.decorators.http import require_http_methods
+
+from thunderbird_accounts.legal.models import LegalDocument, LegalDocumentResponse
+
+
+def _read_legal_content(content_path: str, locale: str) -> str:
+    """Read pre-rendered HTML content for a legal document, falling back to English."""
+    safe_locale = Path(locale).name
```
MelissaAutumn: What's safe about this op?

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+    content_dir = settings.ASSETS_ROOT / 'legal' / content_path
```
MelissaAutumn: We should make this an explicit path rather than rely on settings.ASSETS_ROOT to always be a path.

MelissaAutumn:
```suggestion
    content_dir = Path(settings.ASSETS_ROOT, 'legal', content_path)
```

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+    return JsonResponse({'responses': created})
+
+
+@login_required
+@require_http_methods(['GET'])
+def get_current_legal_docs(request):
+    """Returns current TOS and Privacy documents with their pre-rendered HTML content
+    and whether the authenticated user has accepted them."""
+    current_docs = LegalDocument.objects.filter(is_current=True)
+    locale = request.GET.get('locale', 'en')
```
MelissaAutumn: Would be a good idea to normalize our default language/locale code to `settings.DEFAULT_LANGUAGE`.

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+    _record_response(request, LegalDocumentResponse.Action.DECLINED)
+
+    # TODO: Offboard the user / Delete user data (?)
+
+    if settings.AUTH_SCHEME == 'oidc' and request.user.oidc_id:
+        try:
+            from thunderbird_accounts.authentication.clients import KeycloakClient, RequestMethods
+
+            kc_client = KeycloakClient()
+            kc_client.request(f'users/{request.user.oidc_id}/logout', RequestMethods.POST)
```
MelissaAutumn: I don't know if we have backchannel logouts enabled. Worth a test once we hit stage, otherwise I'll have to adjust a config somewhere.

MelissaAutumn (approved):
Look into the following but it should be good to merge right after.

assets/app/vue/views/TosPrivacyView/index.vue
davinotdavid: There might be some gymnastic that could be done but I don't think it is very worth it. A quick try of making a new endpoint locally to serve the HTML directly and put an iframe there got me this:

image

After working around this, I'd have to either inject with some hackery or duplicate some styling for the content. Perhaps we can punt this to another ticket as an enhancement if necessary?

src/thunderbird_accounts/legal/views.py
davinotdavid: `locale` comes from a query param so perhaps someone could attempt to pass in a relative path as locale and potentially retrieve other .html files

assets/app/vue/views/TosPrivacyView/index.vue
davinotdavid: Ah that makes sense, I've change this to `window.location.href = '/'` and added a comment so we remember why we are doing this. Good catch, thanks!

src/thunderbird_accounts/legal/management/commands/convert_legal_docs.py
davinotdavid: TIL, thanks!

src/thunderbird_accounts/legal/views.py
davinotdavid: I've added a `logging.error` at least in this case but interesting point. If we error out here, do we skip the acceptance altogether or hard-block all users?

src/thunderbird_accounts/legal/views.py
davinotdavid: Ah my bad yes, we already have that defined so changed here. Thanks!

assets/app/vue/views/TosPrivacyView/index.vue
MelissaAutumn: Might be something to do with missing or incorrect header. This is fine, thanks for looking!

src/thunderbird_accounts/legal/views.py
MelissaAutumn: Mind adding a comment?

davinotdavid: Yep! Added, thanks!

MelissaAutumn (changes requested):
Actually for the language code, since it is used in the path we should check if the locale is an exact match for supported languages. otherwise we should error out.

Something like: `locale in ['en']` (and eventually more, also this should come from settings.)

That should fix the bot yellings. (gosh their code suggestions are garbo.)

davinotdavid:
@MelissaAutumn thanks for the suggestion! I've added a `settings.SUPPORTED_LEGAL_LANGUAGES` that seems to have solved some of the bot yelling but not all 😞 Tempted to dismiss it as I think these are enough guards. WDYT?

src/thunderbird_accounts/legal/tests/test_views.py
```diff
@@ -0,0 +1,371 @@
...
+        cleanup_dir = settings.ASSETS_ROOT / 'legal' / 'tos' / 'v2.0'
+        if cleanup_dir.exists():
+            shutil.rmtree(cleanup_dir)
+
+    @override_settings(SUPPORTED_LEGAL_LANGUAGES=['en', 'de'])
+    def test_returns_localized_content_when_available(self):
+        (self.content_dir / 'de.html').write_text('<h1>AGB</h1>', encoding='utf-8')
+        (self.content_dir / 'en.html').write_text('<h1>TOS</h1>', encoding='utf-8')
+
+        from thunderbird_accounts.legal.views import _read_legal_content
```
MelissaAutumn: I think this can live at the top of the test

src/thunderbird_accounts/legal/views.py
MelissaAutumn: Hard block, it's more of a @kewisch question because there might be legal problems if we allow them through. Best to raise in the next meeting, we can follow up on this though.

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,111 @@
...
+
+from thunderbird_accounts.legal.models import LegalDocument, LegalDocumentResponse
+
+
+def _read_legal_content(content_path: str, locale: str) -> str:
+    """Read pre-rendered HTML content for a legal document, falling back to English."""
+    if locale not in settings.SUPPORTED_LEGAL_LANGUAGES:
+        locale = settings.DEFAULT_LANGUAGE
+
+    content_dir = Path(settings.ASSETS_ROOT, 'legal', content_path)
```
MelissaAutumn:
```suggestion
    base_path = Path(settings.ASSETS_ROOT, 'legal')
    content_dir = Path(base_path, content_path)
    if not content_dir.resolve().is_relative_to(base_path):
        logging.error('directory traversal attack!')
        return ''
```

This should prevent someone from injecting `../` into content_path (like if someone managed to get db write access then they can inject this to get container fileaccess.) 

The github botql was basically suggesting this but for locale, which is no longer possible as locale is checked against a list. But it's still possible with content_path. 

(Although I guess locale is still possible if they've modified the settings.py but then they'd probably already have file access.)

MelissaAutumn (changes requested):
Just a few things

src/thunderbird_accounts/legal/views.py
davinotdavid: Hmm makes sense, better make the guards here even though having db write access sounds way more scarier than perusing the container files themselves. Thanks!

src/thunderbird_accounts/legal/views.py
kewisch: If we localize the terms of service or privacy policy, this isn't something translators or an agency could do. We'd need to run it through a lawyer to ensure the translation doesn't create obligations. This isn't something we'd do now as the terms may still have some tweaks over time. I'm guessing it would be at least another year out before we consider it.

So I think it is fair to say that we can just return the English text we have now unconditionally, and when we do translate the ToS we can change the code to use the localized versions.

Good call though on the blank ToS, I agree this seems like a think we shouldn't allow the user to see. Ultimately if it comes to a dispute, we need to prove the user has seen and accepted the terms, and if there is a path where blank text could be shown they could argue that they saw blank text.

assets/app/vue/locales/en.json
```diff
@@ -417,6 +417,21 @@
...
       "cancelButtonLabel": "Cancel"
     },
+    "tosPrivacy": {
+      "title": "One quick step before you continue",
+      "description": "Please review the Terms of Service and Privacy Policy, then accept to continue using Thunderbird Pro.",
+      "tosTab": "Terms of Service",
+      "privacyTab": "Privacy Policy",
+      "acknowledgment": "By accepting, you acknowledge that you have read and understood our policies.",
+      "declineButton": "Decline and cancel registration",
+      "declineAfterPayment": "Decline and cancel account",
```
MelissaAutumn, line 434: @kewisch for now can we have this read: "Decline and log out"? 

Maybe add a mention of "If you want to cancel, please contact support."

Since we don't have automated account cancellation yet, this reads and feels very odd.

assets/app/vue/locales/en.json
```diff
@@ -417,6 +417,21 @@
       "deleteAccountButtonLabel": "Request to delete data",
       "cancelButtonLabel": "Cancel"
     },
+    "tosPrivacy": {
+      "title": "One quick step before you continue",
+      "description": "Please review the Terms of Service and Privacy Policy, then accept to continue using Thunderbird Pro.",
+      "tosTab": "Terms of Service",
+      "privacyTab": "Privacy Policy",
+      "acknowledgment": "By accepting, you acknowledge that you have read and understood our policies.",
+      "declineButton": "Decline and cancel registration",
```
MelissaAutumn, line 433: Currently this does not remove the account either. I guess we'll have a stale account remover at some point though.

assets/app/vue/views/TosPrivacyView/index.vue
```diff
@@ -0,0 +1,287 @@
...
+
+      <p class="acknowledgment">{{ t('views.tosPrivacy.acknowledgment') }}</p>
+
+      <div class="buttons">
+        <primary-button variant="outline" :disabled="declining" @click="handleDecline">
+          {{ declining ? t('views.tosPrivacy.loading') : declineButtonLabel }}
+        </primary-button>
+        <primary-button :disabled="accepting" @click="handleAccept">
+          {{ accepting ? t('views.tosPrivacy.loading') : t('views.tosPrivacy.acceptButton') }}
```
MelissaAutumn: Let's not change the text of the button as it makes it appear as if it's flashing on my end.

MelissaAutumn (changes requested):
Finally have time to test it in person, works great just a few tiny things one of which needs @kewisch or @malini input.

assets/app/vue/locales/en.json
MelissaAutumn: (or @malini if you're available to answer!)

kewisch: I suspect you mean specifically the after payment case? If we just tell them to log out, we have no indication they don't actually want to accept. We should ideally be piping them to support for the meanwhile.

Maybe we can keep it "Decline and cancel account", and when they click we open the support form? Bonus points for showing a banner at the top when people come in this way that says e.g. "We're sorry to see you go! Please contact support to cancel your account".

assets/app/vue/locales/en.json
kewisch: Yeah, we'll need to monitor this for a bit. If they have not yet registered and actively click this button, would it be difficult to simply remove the keycloak account? The other services shouldn't be set up yet, so we'd likely not have any stale data?

assets/app/vue/locales/en.json
davinotdavid: For both pre/post payment cases, if the person clicks on the "decline" we are recording in the DB that they declined the terms (implemented this since it was in the spec in the SoW).

I can make this change to the instead of logging the user out, redirecting them to the `/contact` route with a banner of sorts though!

assets/app/vue/locales/en.json
davinotdavid: Ok cool, I've extracted the current delete user that we do in Django Admin to a function in `util.ts` and calling that in case the user -> **does not** <- have an active subscription and seems to work well.

In the case the user ->**does**<- has an active subscription, then I am redirecting to the contact page with a notice bar warning message for now.

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,121 @@
...
+def accept_legal_docs(request):
+    return _record_response(request, LegalDocumentResponse.Action.ACCEPTED)
+
+
+@login_required
+@require_http_methods(['POST'])
+def decline_legal_docs(request):
+    _record_response(request, LegalDocumentResponse.Action.DECLINED)
+
+    if not request.user.has_active_subscription:
```
MelissaAutumn:
```suggestion
    if not request.user.is_awaiting_payment_verification and not request.user.has_active_subscription:
```

MelissaAutumn: is_awaiting_payment_verification is flipped when a user has paid but we haven't gotten the webhook that verifies it yet. Good to also add this into the mix.

src/thunderbird_accounts/legal/views.py
```diff
@@ -0,0 +1,121 @@
...
+        delete_user_data(request.user)
+        django_logout(request)
```
MelissaAutumn, line 113: Might need to logout before deleting. Not sure it  matters though.

assets/app/vue/locales/en.json
MelissaAutumn: I can live with that. Thanks!

MelissaAutumn (approved):
Minor changes, but it looks good to go after they're applied.

Tested locally and it does indeed delete the account I just created.

src/thunderbird_accounts/legal/views.py
davinotdavid: Ah good call, thanks for the reminder and suggestion!

src/thunderbird_accounts/legal/views.py
davinotdavid: Hm when I tried switching these two, it gave me an error in `delete_user_data` when trying to get the `user.oidc_id` as the `user` no longer existed 🤔 so I will keep it as is since it is working as expected.

That made me realize that I missed one translation string though, so good catch!

PR #720, opened 2026-04-17
Connect now / Custom Protocol URL

### LINKED ISSUE #704: Add actual link to "Connect now" button in Thunderbird for Desktop setup
We currently don't link to anything in the "Connect now" button just yet. Once we have a custom protocol mapping setup and agreed upon with the necessary query params for the Accounts Hub in Desktop, we should update this button.

Image

Also, this issue included handling the "fallback" case when Desktop is not installed. For reference, other applications simply add a tooltip on the button with a "Didn't open application X? Download it here." text.

Marking as `design-needed` as the fallback case is still not defined UI/UX-wise.

### LINKED ISSUE #708: Hide "Connect now" button through a local storage flag
To facilitate the development on the Desktop side for the "Connect now" button and not to confuse our existing users, we should hide the button's section under a local storage flag and make the "Download" button be the primary action there.

davinotdavid:
## Description of changes

- Updated docs with the new `show-connect-now` waffle feature flag.
- Added a `POST desktop-connect/token` endpoint that current only returns the user's access token.
- Changed from link to button for the "Connect now" to make the request and change the `window.location` with the formatted custom protocol mapping in the format that was suggested by the Desktop team.
- Improved types for waffle flags so that we avoid typos in the future.

- We don't have design input yet on the fallback scenario so that's not implemented in this PR.

## How to test

Prerequisite: Have TB Desktop v153 or higher installed

- Login with `admin@example.org / admin`
- Access http://localhost:8087/admin/waffle/flag/
- Add a flag named `show-connect-now` and turn it on for everyone
- Navigate to http://localhost:8087/mail
- Click the `Connect now` button to trigger the Account Hub in TB Desktop

## Related issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/704


assets/app/vue/views/MailView/sections/DashboardSection/components/GetStartedWithThundermailDesktop.vue
```diff
@@ -1,16 +1,48 @@
...
+    if (!data.success) {
+      error.value = t('views.mail.sections.dashboard.getStartedWithThundermail.desktopPanel.failedToGenerateToken');
+      return;
+    }
+
+    const url =
+      'net.thunderbird://accounts/new' +
+      `?name=${encodeURIComponent(userDisplayName)}` +
+      `&email=${encodeURIComponent(primaryEmail)}` +
+      `&token=${encodeURIComponent(data.token)}`;
```
davinotdavid: Based on the Matrix discussions, would this be ok as a starting point for URL to unblock the Desktop part of the integration development?

cc @darktrojan @freaktechnik

src/thunderbird_accounts/mail/views.py
```diff
@@ -515,6 +515,23 @@ def appointment_caldav_setup(request: HttpRequest):
...
+    embedded in the page to limit its exposure in the DOM."""
+
+    access_token = request.session.get('oidc_access_token')
+    if not access_token:
+        return JsonResponse(
+            {'success': False, 'error': str(_('Authentication token not available. Please try logging in again.'))},
+            status=401,
+        )
+
+    return JsonResponse({'success': True, 'token': access_token})
```
davinotdavid: During today's Accounts/Mail meeting, it was suggested that the initial attempt to make the integration work was to pass in the access token as is for simplicity. In the future, we could generate a single use token of sorts encrypted using a secret that only TB Desktop and Accounts have.

I've made this as a dedicated endpoint since it leaving the access token in the `window._page` global wasn't a great idea.

README.md
```diff
@@ -187,13 +197,13 @@ Make sure that the containers are already running.
...
 ```shell
-docker compose exec backend uv run python manage.py test
+docker compose exec accounts uv run python manage.py test
 ```
 
 To run tests for a specific module:
 
 ```shell
-docker compose exec backend uv run manage.py test thunderbird_accounts.client.tests
+docker compose exec accounts uv run manage.py test thunderbird_accounts.mail.tests
```
davinotdavid: Changed from `client` to `mail` since we don't have a module for `client` anymore

assets/app/vue/views/MailView/sections/DashboardSection/components/GetStartedWithThundermailDesktop.vue
darktrojan: I think I'd prefer something like "net.thunderbird://add-thundermail/?…" so anyone reading the URL can immediately see what's going on. Also that avoids any potential issues if we add other "accounts" actions later on (all things named "accounts" would have to be handled in the same place, which isn't necessarily bad, but might add a few more hoops to jump through).

Note that passing the access token would only give Thunderbird access for that session, and the user would be asked to sign in again next time they started Thunderbird. If it was the refresh token instead, Thunderbird could then request the access token from the server.

davinotdavid: Sounds good to me! I've updated the PR with your suggested URL and passing down the refresh token instead.

darktrojan: After thinking about this a bit more, I've gone with net.thunderbird://thundermail/add?… in my patch at https://phabricator.services.mozilla.com/D295745.

MelissaAutumn:
@davinotdavid is this still needed? I don't remember where we landed on this feature.

davinotdavid:
@MelissaAutumn yes, with a few caveats:

This implementation uses the refresh token. This has been flagged as being not secure enough and an investigation around using Keycloak's Action Tokens / Single Use Tokens but from my little research, it seems that we would have to make a plugin for it :/

I will check with the Desktop team to see the timelines for them to ship the code on their side and sync here afterwards. cc @darktrojan

darktrojan:
Hi, we're already shipping our side in 153. The URL is not the same as it is here though (see my inline comment).

https://hg-edge.mozilla.org/releases/comm-esr153/file/18da98f8a2f3dbe49d91c8a872aed9594d5aeebb/mail/components/accountcreation/modules/Thundermail.sys.mjs

MelissaAutumn:
Sorry I forgot to reply for 2 weeks...

Anyways, I don't think we can use refresh token as we don't grant offline_access to any of the services, only mail clients. I know it would make this effectively useless, but could we just transfer the details except for the refresh token? 

The user would have to login again, but I think that's better than handing off a refresh token that will expire at some point. Otherwise we may want to use a short-lived one-off token that the desktop client can use to request a session for a user. 

It's a good thing we have a principal java engineer, lol! (@Sancus)

MelissaAutumn:
We discussed this in the Thundermail meeting, @Sancus mentioned device autorization instead of passing refresh token (aka go to this url and enter this code to login.) The spec: https://datatracker.ietf.org/doc/html/rfc8628

Keycloak implements it: https://www.keycloak.org/securing-apps/oidc-layers#_device_authorization_endpoint but I have not used it personally. We should investigate this instead, as again I don't think passing refresh token is a great idea. 

If possible we'd be doing this under the desktop client too so their login session would appear correctly, and they'd have access to offline_access.

davinotdavid:
@MelissaAutumn I believe the downsides of using the access token are minimal (just re-auth when it expires) compared with the benefits of having an expedited initial setup / perceived cohesiveness between products so I went ahead and added that.

The device authorization bit is a better implementation for sure but will take more time on both sides to be live which I am not sure if it's worth pursuing right now (IMHO).

This can be reviewed if the approach is aligned! Thanks 🙏

Sancus:
Based on my testing yesterday:
1) The refresh token passed to desktop is useless, as it's for the wrong client id, and accounts is not permitted to retrieve a refresh token for the desktop client id. At least not without some sketchy and insecure modifications to our current client id setup.
2) There is really no reason to pass a token for the common use case. When Thunderbird opens your default browser to run oAuth, if you had already signed into keycloak(for say, accounts) then sign in and callback proceed immediately with no further user input.

So this patch should actually just send a fixed literal for the token for now, and it works OK. We should remove the requirement for the token from desktop and have it run oAuth itself on Account Hub. Due to the likely pre-existing session in the user's browser, this should Just Work(tm).

I filed a [bug on the desktop side](https://bugzilla.mozilla.org/show_bug.cgi?id=2062524) but I see no reason Davi can't handle that if he's up for it.

davinotdavid:
Thanks @Sancus ! I really appreciate the thorough testing and deep dive into this.

I have applied the suggested changes in this PR but also have submitted a patch for TB Desktop to remove the token requirement. It has one approval so far so I am confident that we can land that soon™ and update this PR to not pass in a token at all.

Even though it works by passing a bogus token too, I think it would be confusing to leave it there so when the patch in Desktop gets merged I can remove it from here too and mark the PR for Ready for Review again (unless you want to review already and pre-approve heh)

Sancus:
> Thanks @Sancus ! I really appreciate the thorough testing and deep dive into this.
> 
> I have applied the suggested changes in this PR but also have submitted a patch for TB Desktop to remove the token requirement. It has one approval so far so I am confident that we can land that soon™ and update this PR to not pass in a token at all.
> 
> Even though it works by passing a bogus token too, I think it would be confusing to leave it there so when the patch in Desktop gets merged I can remove it from here too and mark the PR for Ready for Review again (unless you want to review already and pre-approve heh)

The desktop side is (already) live in ESR 153.0 isn't it? So you'll need to make sure that your desktop patch is uplifted to ESR 153.whatever and that ~most people have updated before changing thunderbird-accounts.

davinotdavid:
@Sancus the portion that is live on the desktop side in ESR / 153.x  is the one that requires the token, yes. The patch in phrabricator that removes the token is yet to be merged in https://phabricator.services.mozilla.com/D317845 and that's the one I was wrongly hoping to land it before this PR.

Makes sense to merge this in then if approved!

MelissaAutumn (approved):
Looks good. I mightt say you open a new tab or window for the connect now click, but this works!

Sancus (approved):
works for me

PR #721, opened 2026-04-17
npm audit fix and bump browserstack-node-sdk version up.

MelissaAutumn:
There's still the aws-sdk security alert but nothing we can do until browserstacks fixes their dependencies.


markstos:
Closed as outdated. Something else already updated the browserstack dep a newer version.

PR #722, opened 2026-04-17
Stalwart posthog

Sancus:


davinotdavid (approved):
code-wise lgtm and I can cURL my way into testing a hypothetical case but I didn't test the full integration, perhaps it is easier to test this in stage?

PR #723, opened 2026-04-17
New Sign Up Flow

### LINKED ISSUE #398: Keycloak: Fix login hint
Right now we recieve the login hint, but the theme doesn't do anything with it....

This causes login forms that normally would have the login username prefilled to be empty instead.

### LINKED ISSUE #1193: Implement new sign up flow
Making notion sync happy. This is the task for the new sign up flow, rather than the milestone.

MelissaAutumn:
Fixes #1193
Also fixes #398 

This completes step 1-3 (and a extra landing step after you enter your email in)
https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/69ced662f6f5e6348a41df48

To test, just make sure you have sign-ups disabled in Keycloak's tbpro realm. That redirects the subscribe link on login to account's sign up flow. (This will be fixed in the future.)

There's an additional issue with field input error messages overflowing, that's a services-ui issue I'm going to look into soon. 

This PR does not cover zeplin Step 4.

### What's changed?
* Updated sign up flow to match zeplin designs
* Improve a11y issues with flow and error messages
* Refactored Sign up view (frontend) to be more accounts focused
* Moved sign up route from a django view route to a django rest framework api endpoint
* Split authentication tests into separate files
* Updated backend tests for sign up's api endpoint (adapted from the view route.)
* Updated and added additional frontend e2e tests for sign up

## Screenshots
Screen Shot 2026-04-21 at 11 45 24-fullpage
Screen Shot 2026-04-21 at 11 45 32-fullpage
Screen Shot 2026-04-21 at 11 45 50-fullpage
Screen Shot 2026-04-21 at 11 45 55-fullpage

### Examples of field errors
Screen Shot 2026-04-21 at 11 46 05-fullpage
Screen Shot 2026-04-21 at 11 46 18-fullpage


test/e2e/const/constants.ts
```diff
@@ -19,10 +19,10 @@ export const PLAYWRIGHT_TAG_E2E_SUITE = '@e2e-suite';
...
 
 // timeouts
-export const TIMEOUT_2_SECONDS = 2000;
-export const TIMEOUT_5_SECONDS = 5000;
-export const TIMEOUT_10_SECONDS = 5000;
-export const TIMEOUT_30_SECONDS = 30000;
+export const TIMEOUT_2_SECONDS = 0;//2000;
+export const TIMEOUT_5_SECONDS = 0;//5000;
+export const TIMEOUT_10_SECONDS = 0;//5000;
+export const TIMEOUT_30_SECONDS = 0;//30000;
```
MelissaAutumn: Note to self: Fix this.

src/thunderbird_accounts/authentication/tests/test_admin.py
```diff
@@ -0,0 +1,465 @@
+from unittest.mock import MagicMock, patch
```
MelissaAutumn, line 1: Unchanged, please ignore.

src/thunderbird_accounts/authentication/tests/test_api.py
```diff
@@ -0,0 +1,155 @@
+from json import JSONDecodeError
```
MelissaAutumn, line 1: Test changes in this file! (It's mostly a port of the sign up view test.)

src/thunderbird_accounts/authentication/tests/test_middleware.py
```diff
@@ -0,0 +1,300 @@
+from unittest.mock import MagicMock
```
MelissaAutumn, line 1: Unchanged, please ignore.

src/thunderbird_accounts/authentication/tests/test_utils.py
```diff
@@ -0,0 +1,95 @@
+
```
MelissaAutumn, line 1: Unchanged, please ignore.

assets/app/vue/views/SignUpView/components/SignUpLayout.vue
```diff
@@ -0,0 +1,140 @@
...
+.buttons {
+  margin-top: var(--space-24);
+  width: 100%;
+
+  .submit {
+    margin-right: 0;
+    margin-left: auto;
+  }
+}
+</style>
```
MelissaAutumn, line 120: I need to nuke these unused styles.

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
```diff
@@ -0,0 +1,133 @@
...
+          errors.value = null;
+          window.location.href = error['href'];
+        } else {
+          step.value = SignUpSteps.USERNAME;
+        }
+      } catch {
+        errors.value = 'Unknown error';
+      }
+      return false;
+    }
```
MelissaAutumn: I need to unnest this.

assets/app/vue/views/SignUpView/components/SignUpLayout.vue
```diff
@@ -0,0 +1,140 @@
...
+    <form @submit.prevent="onSubmit()" @keyup.enter="onSubmit()" ref="form">
+      <div class="form-elements">
+        <slot name="form-elements" />
+        <slot name="form-extras" />
+      </div>
+      <div class="buttons" v-if="!hideActions">
+        <primary-button form-action="submit" data-testid="submit-button" class="submit" :disabled="submitDisabled"
+          @click.prevent="onSubmit()">
+          {{ $t('views.mail.views.signUp.continue') }}
+        </primary-button>
```
davinotdavid, line 59: It seems that the text varies between steps so we should make either the whole button as a prop or just the text in it. For example, in step 3 it is "Verify email" instead:

Image

assets/app/vue/locales/en.json
```diff
@@ -266,6 +245,41 @@
...
+            "username": "Username",
+            "password": "Password",
+            "confirmPassword": "Confirm Password",
+            "yourName": "Your Name",
+            "verificationEmail": "Verification email"
+          },
+          "step1": {
+            "title": "Create an account",
+            "subtitle": "Choose your Thundermail username.",
+            "emailSuggestions": "Try one of these:",
```
davinotdavid, line 261: Double checking, this doesn't seem to be implemented yet as this text is currently not being used. Is that correct?

assets/app/vue/views/SignUpView/components/SignUpLayout.vue
```diff
@@ -0,0 +1,140 @@
...
+import { useThrottleFn } from '@vueuse/core';
+
+const form = useTemplateRef<HTMLFormElement>("form");
+
+withDefaults(defineProps<{
+  stepId: string;
+  title: string;
+  subtitle: string;
+  submitDisabled?: boolean | null;
+  hideActions?: boolean | null;
```
davinotdavid: I think that if we want to enforce `null` and not `undefined` this should be:

```suggestion
  submitDisabled: boolean | null;
  hideActions: boolean | null;
```

Otherwise I believe it allows for `boolean | null | undefined` (although I would still lean towards removing the `| null` altogether and believe in the truthy-ness checks)

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
```diff
@@ -0,0 +1,133 @@
...
+    name.value = null;
+    username.value = null;
+    verificationEmail.value = null;
+    password.value = null;
+    confirmPassword.value = null;
+    timezone.value = null;
+    lang.value = null;
+    step.value = SignUpSteps.USERNAME;
+    // Hmm...
+    errors.value = null;
```
davinotdavid: 👀

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  const username = useSessionStorage(`${sessionStorageKeyPrefix}/username`, null);
+  const verificationEmail = useSessionStorage(`${sessionStorageKeyPrefix}/verificationEmail`, null);
+  const timezone = useSessionStorage(`${sessionStorageKeyPrefix}/timezone`, null);
+  const lang = useSessionStorage(`${sessionStorageKeyPrefix}/lang`, null);
+  // In-memory only!
+  const step = ref(SignUpSteps.USERNAME);
+  const password = ref(null);
+  const confirmPassword = ref(null);
+
+  const errors = ref(null);
```
davinotdavid: Perhaps a nit here but maybe we should either define the type as being `string | null` explicitly or change the name to `errorMessage` or something as `errors` sounds like this would be an array / list of sorts.

assets/app/vue/views/SignUpView/views/Step3Verify/index.vue
```diff
@@ -0,0 +1,59 @@
...
+const onSubmit = async () => {
+  loading.value = true;
+  const response = await submit();
+
+  if (response === true) {
+    window.location.href = '/sign-up/complete';
+  } else {
+    window.setTimeout(() => {
+      loading.value = false;
+    }, 0);
```
davinotdavid: Is the `setTimeout` here necessary? Could this potentially be replaced with a [`nextTick`](https://vuejs.org/api/general.html#nexttick) call?

keycloak/themes/tbpro/assets/vue/views/LoginView/index.vue
```diff
@@ -30,6 +30,7 @@ const passwordError = computed(() => {
 
 // CheckboxInput requires a valid ref as a model to show the check icon
 const rememberMeChecked = ref(rememberMe);
+const username = ref(window._page.currentView?.loginHint);
```
davinotdavid, line 33: ❤️

src/thunderbird_accounts/mail/api.py
```diff
@@ -0,0 +1,47 @@
...
+    # So EmailValidator.__call__ will raise a ValidationError if it fails, but they're the wrong
+    # ValidationError...So catch this ValidationError so we can raise DRF's ValidationError.
+    try:
+        email_validator(full_username)
+    except DjValidationError as ex:
+        raise ValidationError(ex.message)
+
+    user = User.objects.filter(username=full_username).first()
+    if user:
+        raise ValidationError(_('This username is already taken. Try another one.'))
```
davinotdavid, line 45: We might want to add some random valid variations based off of the username as suggestions and return that to the frontend here

davinotdavid (changes requested):
Nice! Happy path case works well, I think we need to update some other pages as well as the logout screen since it is showing double logos:

Logout page
Image

After clicking on the link in the verify email (Do we need to have this intermediate screen of 'performing actions'?)
Image

Image

Also, not sure if this will still be done in this PR or in a follow up but I am assuming that your comments will be tackled too:

(Unused styles cleanup)
https://github.com/thunderbird/thunderbird-accounts/pull/723/changes#r3119696498

(Unnesting this)
https://github.com/thunderbird/thunderbird-accounts/pull/723/changes#r3119697472

assets/app/vue/locales/en.json
MelissaAutumn: I'm punting that to a future ticket, thanks for the reminder.

assets/app/vue/views/SignUpView/components/SignUpLayout.vue
MelissaAutumn: I agree with removing | null.

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
MelissaAutumn: Hmm...

Turned out to be fine.

assets/app/vue/views/SignUpView/views/Step3Verify/index.vue
MelissaAutumn: Probably, will take a look.

src/thunderbird_accounts/mail/api.py
MelissaAutumn: Punting that to a future issue.

assets/app/vue/views/SignUpView/components/SignUpLayout.vue
MelissaAutumn: Ah I knew that "submitText" was for something right after I removed it. 

Thanks for the spot.

MelissaAutumn:
Double logo isn't ideal, I'll fix that up. Thanks!

rwood-moz (changes requested):
Wow this is so cool (and exciting!). I gave it a try locally and have some notes:

- On the first step I notice for username it has `@example.org` for the domain but I'm assuming this will be `@thundermail.com` when on stage/prod
- On the first step, under the username field, the email address displayed `yourname@example.org` should be in italics according to Zeplin
- On the `Set a password` screen the place holder text is missing from all of the fields. In Zeplin I see `Create a password`, `Confirm password`, and `Your name` place holder text in the Password, Confirm Password, and Your Name fields respectively
- On the `Add a verification email` screen, the Verification email field is missing place holder text (Zeplin shows `yourname@example.com`)
- Verification email screen button label is `Continue` but Zeplin shows `Verify email`
- Awesome work adding/updating the E2E tests! All the tests passed for me locally except the first test `form successfully submits but is not on allow list` fails as after the form is submitted it just goes to `http://localhost:8087/sign-up/complete` and not the wait list form (is the allowed email / wait list not implemented in this new flow yet maybe?)

MelissaAutumn:
@rwood-moz @davinotdavid I've fixed most of what was outlined, thanks for catching them!

I've reached out to design for some questions on excessive placeholders, but we shouldn't be blocked on that. Additionally I've added a new template variable for any backend feature flags, and exposed `USE_ALLOW_LIST`. This allows us to retrieve that information and potentially skip a e2e test if it's disabled (which solves the test issue Rob mentioned.)

rwood-moz:
> @rwood-moz @davinotdavid I've fixed most of what was outlined, thanks for catching them!
> 
> I've reached out to design for some questions on excessive placeholders, but we shouldn't be blocked on that. Additionally I've added a new template variable for any backend feature flags, and exposed `USE_ALLOW_LIST`. This allows us to retrieve that information and potentially skip a e2e test if it's disabled (which solves the test issue Rob mentioned.)

Very cool re: Skips the E2E test if the allow list is not enabled. Works great for me locally (tried with and without `USE_ALLOW_LIST=True`) thanks!

rwood-moz (approved):

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
```diff
@@ -23,29 +29,56 @@ import OrbGraphic from '@kc/images/orb-graphic.png';
...
     .left-side {
       display: none;
+      position: relative;
     }
 
     .right-side {
       position: relative;
       display: flex;
       flex-direction: column;
-      justify-content: center;
```
davinotdavid, line 48: I believe this should be back to solve the vertical alignment

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
```diff
@@ -1,20 +1,26 @@
...
-        <img :src="OrbGraphic" alt="Thunderbird Orb" class="orb-graphic" />
-      </div>
+      <!-- Left side: Featured image -->
+      <div class="left-side"></div>
 
       <!-- Right side: Panel -->
       <div class="right-side">
         <div class="panel">
-          <slot />
+          <a :href="clientUrl" class="logo-link">
```
davinotdavid, line 18: I think we are missing the CSS that was removed here https://github.com/thunderbird/thunderbird-accounts/pull/723/changes#diff-8ac2837516d35e427bb66048f345bd691d1d54d665c26e0f5548c952acd8f9a6L107-L121

davinotdavid (changes requested):
Sorry that I didn't catch this before I think but the responsiveness is a bit broken compared with what's on prod. I believe the spacing between the logo and the body + being vertically aligned is off:

Image

Mobile
Image

--

For these, I don't think we have specific Zeplin designs for it but perhaps we should also tackle? If not, of course we can punt this to another ticket!

The email verification screen / re-send email seems quite different style-wise from the rest 

Image

The page expired as well:

Image

MelissaAutumn:
That's not ideal huh. Thanks for catching that! I'll look into the playwright docs and see if we can catch these through e2e testing too.

rwood-moz:
> That's not ideal huh. Thanks for catching that! I'll look into the playwright docs and see if we can catch these through e2e testing too.

IMO I probably wouldn't write an E2E test for that when it's probably just a one time change/fix, it's best to keep the E2E tests to main user flows since they take a lot of overhead & time to run and maintain.

MelissaAutumn:
> > That's not ideal huh. Thanks for catching that! I'll look into the playwright docs and see if we can catch these through e2e testing too.
> 
> IMO I probably wouldn't write an E2E test for that when it's probably just a one time change/fix, it's best to keep the E2E tests to main user flows since they take a lot of overhead & time to run and maintain.

Good point, I believe I was thinking about using the screenshot feature to detect visual inconsistencies, but that's better suited for a component library like services-ui.

MelissaAutumn:
I'll be normalizing the styling for the keycloak theme in the second half of this ticket btw. I want to avoid this PR from becoming bigger than it is.

davinotdavid (approved):
tyvm! lgtm 🚀

PR #682, opened 2026-03-28
Introduce StrEnum for consistent Stalwart types

MooseTheRebel:
Small subset of work, in support of https://github.com/thunderbird/thunderbird-accounts/issues/251 .  

This PR introduces the `StalwartPrincipalType` enum and replaces all usages.  I've tried to keep this change small, so that the implementation and style could be discussed and refactored if needed. 

---

A future commit (or a follow-on PR) could address the following:

Creating enums for `StalwartPatchAction` and `StalwartPrincipalField` 

<details><summary>Click to expand </summary>
<p>

```python
class StalwartPatchAction(StrEnum):
    SET = 'set'
    ADD_ITEM = 'addItem'
    REMOVE_ITEM = 'removeItem'
    CLEAR = 'clear'


class StalwartPrincipalField(StrEnum):
    TYPE = 'type'
    NAME = 'name'
    DESCRIPTION = 'description'
    QUOTA = 'quota'
    SECRETS = 'secrets'
    EMAILS = 'emails'
    URLS = 'urls'
    MEMBER_OF = 'memberOf'
    ROLES = 'roles'
    LISTS = 'lists'
    MEMBERS = 'members'
    ENABLED_PERMISSIONS = 'enabledPermissions'
    DISABLED_PERMISSIONS = 'disabledPermissions'
    EXTERNAL_MEMBERS = 'externalMembers'
```

</p>
</details> 

Replacing magic strings in `MailClient` methods

<details><summary>Click to expand </summary>
<p>

| Method | Old string | Replacement |
|---|---|---|
| `_create_principal` | `'quota'`, `'secrets'`, `'emails'`, etc. (default dict) | `StalwartPrincipalField.*` |
| `_update_principal` | `patch_schema` keys (`'type'`, `'name'`, ...) | `StalwartPrincipalField.*` |
| `_update_principal` | `patch_schema` values (`'set'`, `'addItem'`, `'removeItem'`) | `StalwartPatchAction.*` |
| `delete_app_password` | `'removeItem'`, `'secrets'` | `StalwartPatchAction.REMOVE_ITEM`, `StalwartPrincipalField.SECRETS` |
| `save_app_password` | `'addItem'`, `'secrets'` | `StalwartPatchAction.ADD_ITEM`, `StalwartPrincipalField.SECRETS` |
| `save_email_addresses` | `'addItem'`, `'emails'` | `StalwartPatchAction.ADD_ITEM`, `StalwartPrincipalField.EMAILS` |
[...]

PR #724, opened 2026-04-20
Enumify stalwart principal fields

MooseTheRebel:
Relates to https://github.com/thunderbird/thunderbird-accounts/issues/251 .  

I'm curious what you may think about the name of this enum, as StalwartPrincipalField may be redundant.  We could just call it "PrincipalField", as an option.

See also: https://github.com/thunderbird/thunderbird-accounts/pull/682


markstos (commented):
Here, it's less clear to me that the value of the strictness of enumification adds more value than we lose by the worse readability.

@MelissaAutumn What's your take on this one?

MooseTheRebel:
Hi @markstos 👋 That's understandable, let me know what you all decide about this PR and about the linked issues. 

If it helps, I'm happy to work on another issue. Is there a longer-term goal (e.g., a 1-6 month effort) that I can support (Python, Go, maybe Rust)? Happy to chat via [matrix](https://matrix.to/#/%23tb-services:mozilla.org) as well.

markstos:
Thanks for your contribution to our fast moving project.

I'm closing this because the 0.16 upgrade has progressed to the point where this code is now "legacy" and is expected to be removed after not too long, so there's not much value to enumify it to prevent mistakes during further development as there won't be much here.

Enumification in v0.16-active code remains of interest.

PR #727, opened 2026-04-20
Add UTM tags to download buttons

### LINKED ISSUE #698: Add UTM tags to Desktop/Mobile download links
Some attribution for downloads via the dashboard would help us understand how many people are downloading Thunderbird via Thunderbird Pro.

davinotdavid:
## Description of changes

Added UTM tags to the download button in the sections below:

Desktop

image

Mobile

image

## Related issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/698

### DIFF
--- a/assets/app/vue/defines.ts
+++ b/assets/app/vue/defines.ts
@@ -3,6 +3,6 @@ export const OTHER_APPS_SUPPORT_URL = 'https://support.tb.pro/hc/en-us/articles/
 export const WHAT_IS_IMAP_SUPPORT_URL = 'https://support.tb.pro/hc/articles/46108465880467-What-is-IMAP';
 export const WHAT_IS_JMAP_SUPPORT_URL = 'https://support.tb.pro/hc/articles/46109214513939-What-is-JMAP';
 export const WHAT_IS_SMTP_SUPPORT_URL = 'https://support.tb.pro/hc/articles/46106891841043-What-is-SMTP';
-export const DOWNLOAD_THUNDERBIRD_DESKTOP_URL = 'https://www.thunderbird.net/thunderbird/all/';
-export const DOWNLOAD_THUNDERBIRD_MOBILE_URL = 'https://play.google.com/store/apps/details?id=net.thunderbird.android';
+export const DOWNLOAD_THUNDERBIRD_DESKTOP_URL = 'https://www.thunderbird.net/thunderbird/all?utm_campaign=main&utm_medium=tb_pro&utm_source=thundermail_dashboard&utm_content=desktop_download';
+export const DOWNLOAD_THUNDERBIRD_MOBILE_URL = 'https://play.google.com/store/apps/details?id=net.thunderbird.android&referrer=utm_campaign%3Dmain%26utm_medium%3Dtb_pro%26utm_source%3Dthundermail_dashboard%26utm_content%3Dmobile_download';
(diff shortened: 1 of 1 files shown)


assets/app/vue/defines.ts
```diff
@@ -3,6 +3,6 @@ export const OTHER_APPS_SUPPORT_URL = 'https://support.tb.pro/hc/en-us/articles/
 export const WHAT_IS_IMAP_SUPPORT_URL = 'https://support.tb.pro/hc/articles/46108465880467-What-is-IMAP';
 export const WHAT_IS_JMAP_SUPPORT_URL = 'https://support.tb.pro/hc/articles/46109214513939-What-is-JMAP';
 export const WHAT_IS_SMTP_SUPPORT_URL = 'https://support.tb.pro/hc/articles/46106891841043-What-is-SMTP';
-export const DOWNLOAD_THUNDERBIRD_DESKTOP_URL = 'https://www.thunderbird.net/thunderbird/all/';
-export const DOWNLOAD_THUNDERBIRD_MOBILE_URL = 'https://play.google.com/store/apps/details?id=net.thunderbird.android';
+export const DOWNLOAD_THUNDERBIRD_DESKTOP_URL = 'https://www.thunderbird.net/thunderbird/all?utm_campaign=main&utm_medium=tb_pro&utm_source=thundermail_dashboard&utm_content=desktop_download';
+export const DOWNLOAD_THUNDERBIRD_MOBILE_URL = 'https://play.google.com/store/apps/details?id=net.thunderbird.android&referrer=utm_campaign%3Dmain%26utm_medium%3Dtb_pro%26utm_source%3Dthundermail_dashboard%26utm_content%3Dmobile_download';
```
radishmouse:
```suggestion
export const DOWNLOAD_THUNDERBIRD_MOBILE_URL = 'https://play.google.com/store/apps/details?id=net.thunderbird.android&utm_campaign=main&utm_medium=tb_pro&utm_source=thundermail_dashboard&utm_content=mobile_download;
```

radishmouse, line 7: I'm not sure if the mobile URL needs the referrer querystring arg? I' ve suggested a version that removes the `referrer=` and just  drops in the UTM from Natalie's comment on the ticket.

radishmouse (changes requested):
Please double check if the `referrer` querystring arg is necessary.

assets/app/vue/defines.ts
davinotdavid: Hmm that was a suggestion from @kewisch here https://github.com/thunderbird/thunderbird-accounts/issues/698#issuecomment-4283155160 so perhaps it is a particularity for the mobile?

radishmouse: @kewisch if we keep the `referrer`, should it have a specific value?

inattee: Yes I believe 'referrer' is necessary for the Play Store

radishmouse (approved):
All good now!
It seems that the entire UTM querystring is the value of `referrer`.

kewisch:
> All good now! It seems that the entire UTM querystring is the value of `referrer`.

Yes this is correct, it is specific to the play store. See https://ga-dev-tools.google/ga4/campaign-url-builder/play/  for future use.

davinotdavid:
Cool! Thanks all for the inputs and reviews! 🚀

PR #742, opened 2026-04-23
Clean up contact support E2E test

### LINKED ISSUE #737: Update the 'contact support' E2E test to check for the primary Thundermail email not recovery email
The `contact.spec.ts` E2E test signs into TB Accounts and then verifies that the email address field on the contact support form is auto-populated. This form used to auto-populate with the TB Accounts recovery email address, but was changed awhile back to now auto-populate with the actual Thundermail primary email address associated with the TB Account. The test is checking for `E2E_ACCTS_RECOVERY_EMAIL` but should be checking for `E2E_PRIMARY_THUNDERMAIL_EMAIL`. Update the test accordingly. Note that I previously temporarily updated the `E2E_ACCTS_RECOVERY_EMAIL` GHA secret on stage and prod to actually be the primary Thundermail email in order for the tests to pass in their current state; but the tests should be updated regardless.

rwood-moz:
Clean up the contact support E2E test:

- The contact form pre-populates the email address with the primary Thundermail email (test was using recovery email)
- Recovery email .env var is no longer needed anywhere in these E2E tests so remove all references to it and the related GHA secret
- After this lands remove the `E2E_ACCTS_RECOVERY_EMAIL` GitHub actions secret from the repo as it won't be used anymore

Fixes #737.

BrowserStack [link here](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Firefox+Desktop/86?tab=sessions&testListView=spec) running the updated E2E tests on stage (all tests passed).

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -357,7 +357,6 @@ jobs:
     env:
       ACCTS_OIDC_EMAIL: ${{ secrets.E2E_ACCTS_OIDC_EMAIL }}
       ACCTS_OIDC_PWORD: ${{ secrets.E2E_ACCTS_OIDC_PWORD }}
-      ACCTS_RECOVERY_EMAIL: ${{ secrets.E2E_ACCTS_RECOVERY_EMAIL }}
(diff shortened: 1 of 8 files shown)


MelissaAutumn (approved):
woot, thanks!

Issue #659, opened 2026-03-23
New sign up flow


malini:
We have a new FTUE flow we want to set up, as shown in [this Figma](https://www.figma.com/design/uTmD40AGx1CUb6ma29evpz/Thundermail_Exploration_2026_Q1_Product_Direction_v0.0.1?node-id=7-1141&t=2ZsXPkfmkDRKDlpy-0).

MelissaAutumn:
It's ready and in zeplin format
https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/69ced662f6f5e6348a41df48

MelissaAutumn:
Mostly complete with https://github.com/thunderbird/thunderbird-accounts/pull/723 

Will need to follow up with required action nonsense on keycloak templates.

MelissaAutumn:
#723 is merged on stage. Working fine, but I need to fix a step change at the end before the redirect, update the verified email screen in keycloak, and work on normalizing the layout with the rest of keycloak.

kewisch:
@MelissaAutumn are all issues required for follow up filed? Looks like this isn't completely done, but good enough for the scope we had for Earlybird.

MelissaAutumn:
I believe so, I couldn't find any follow up issues that weren't tackled.

PR #745, opened 2026-04-23
Update the keycloak info template to show verify_email specific content.

MelissaAutumn:
Part of #659 

This is annoyingly hard coded in keycloak. So we still need some sort of "Click this button" action and landing page for that. I'm looking into it but we may need to wait until we start making modifications to keycloak itself to be able to match the design.

This change updates the generic info template / sfc, detects if "VERIFY_EMAIL" is the only required action, and if so then displays this custom text. If "VERIFY_EMAIL" is paired with other required actions (like set otp, or change password) then it will be displayed in the old list format. 

We can continue to iterate on this as we go forward. 

## Screenshots
Screen Shot 2026-04-23 at 15 06 32-fullpage
Screen Shot 2026-04-23 at 15 06 36-fullpage


davinotdavid (approved):
lgtm!

PR #753, opened 2026-04-27
Redirect to mail after payment

### LINKED ISSUE #701: Make the Thundermail dashboard the first page after login/payment

davinotdavid:
## Description of changes

We are controlling the post-payment redirect through a `beforeEach` hook in `router.ts` and I've missed this during my [navigation changes PR](https://github.com/thunderbird/thunderbird-accounts/pull/695).

In that PR, I've made it redirect to `mail` after logging in as part of changing the Keycloak + Vue router root path but this `beforeEach` hook was forcibly redirecting it to `dashboard` instead.

Added a comment for clarification!

## Related issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/701 (for real this time)

### DIFF
--- a/assets/app/vue/router.ts
+++ b/assets/app/vue/router.ts
@@ -174,7 +174,7 @@ router.beforeEach((to, _from) => {
   if (isAuthenticated && sendToSubscribe && !['subscribe', 'contact'].includes(to.name.toString())) {
     return { name: 'subscribe' };
   } else if (isAuthenticated && !sendToSubscribe && to.name === 'subscribe') {
-    return { name: 'dashboard' };
+    return { name: 'mail' };
   }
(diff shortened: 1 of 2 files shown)


rwood-moz (approved):
Tried locally on clean stack and can confirm after subscribing and waiting for the payment to go through, I was automatically redirected to the Thundermail page (http://localhost:8087/mail) 🥇

Issue #446, opened 2025-11-21
Aliases should support catchalls but ONLY for custom domains


Sancus:
A catch-all is an alias that applies to all addresses on a domain. These can be added in Stalwart by not specifying any name and just adding the domain by itself like "tb.pro".

We should allow this for custom domains, and in fact, adding it on behalf of the user as a default after their domain is verified would also be great.

However for our domains this should not be permitted.

kewisch:
@laurelterlesky I'm curious what the UX for this could look like. The very technical way to solve this might be to just not make the Name field required, and accept either empty or * with some explanatory text. Then maybe a pill in the list of aliases that says "catch-all". I'm sure there is a better way to do this though. I suspect we might need to explain catch-all to some folks via support article.

<img width="952" height="537" alt="Image" src="https://github.com/user-attachments/assets/903f92d5-f1bc-4e9e-b486-7be7d85ee8a0" />

MelissaAutumn:
Should we maybe have this as a domain option instead, something like "Allow this domain to recieve emails from any address? (Catch all emails.)"

laurelterlesky:
Hello! My eyes are on this. We did some previous work in this area. Going to review that today and see if we have a working solve or if design needs to do a quick pass.

malini:
looks like the current design is [here](https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/69e6b6d96f5a7d3132da6351/approval/69e7d30f8e7c1f6a92cce71e?aprid=673f97e4316f2f5809d3373c)

ryanleesipes:
Maybe to add another thing here we've said in meetings a few times but I don't see an issue for. The alias limit should not apply to aliases attached to custom domains.

MelissaAutumn:
The design (https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/69e6b6d96f5a7d3132da6351) describes a new flow just for catch-all. This work might slip into next week since I want to ensure quality and a11y support. 

So I'm splitting this into two parts, the first part is simply adding support for `*` in the textbox. That should be up by EOD if there's no issues on Stalwart's end.

The second part is adding the flow.

MelissaAutumn:
Part 2 of this ticket will also include a store that communicates with custom domain settings. Right now if you add or remove a domain you'll need to refresh the page in order see those changes affect the page. 

That's not ideal, but with the timeline of a few days there's not much time to fix that.

[...]

PR #761, opened 2026-04-28
Custom domains should support catch alls

MelissaAutumn:
Part of #446 

This is the simplest method of adding catch all domains. I've re-used the original textbox and adjusted the code so if you enter an empty textbox or `*` on a custom domain that doesn't already have a custom domain it will add one as a catch-all.

Note the second help line disappears if that domain already has a catch-all. 

The second part of #446 will be about implementing the flow described in https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/69f10f8bf4ab174d8ae6563f. It will happen following Thunderweek.

Additionally I've adjusted the limit check so only aliases on shared domains (thundermail.com for example) count towards it. Also I noticed we don't have a backend check for that?! So now we do. 

## Screenshots
image
image
image


assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
```diff
@@ -170,7 +216,7 @@ const onDeleteAliasError = (error: string) => {
       min-height: 60px;
       background-color: var(--colour-neutral-lower);
 
-      & + .alias-item {
+      &+.alias-item {
```
MelissaAutumn, line 220: npm lint strikes again!

assets/app/vue/locales/en.json
```diff
@@ -162,6 +162,9 @@
           "emailAliasesDescription": "All emails will deliver mail to your primary email mailbox.",
           "emailAliasesDescriptionTwo": "{aliasUsed} of {aliasLimit} aliases used",
           "emailAliasesPrimaryChangeWarning": "Changing your primary email may affect app sign-ins.",
+          "catchAllTooltip": "Tool tip: All available domains already have a catch-all email. No additional catch-all aliases can be created. {supportUrl}",
```
MelissaAutumn: TODO: Remove tool tip:

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
```diff
@@ -74,11 +97,29 @@ const onDeleteAliasError = (error: string) => {
...
+        <p class="email-aliases-count-text">
+          {{ t('views.mail.sections.emailSettings.emailAliasesDescriptionTwo', {
+            aliasUsed: usedAliases, aliasLimit
+          }) }}
+        </p>
+        <span class="email-aliases-count-text">
+          {{ t('views.mail.sections.emailSettings.catchAllUsed', {
+            catchAllUsed: catchAlls.length, catchAllLimit: customDomains.length,
+          }) }}
+          <aside aria-labelledby="catch-all-tooltip" class="info-tooltip-trigger" tabindex="0">
```
MelissaAutumn: TODO: Only show tooltip if catch all aliases are maxed out.

MelissaAutumn:
Sorry fixed some last second issues I spotted. I explicitly look for an empty string on the backend, but we were sending null. I've fixed the frontend, so now it will submit an empty string as intended.

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
```diff
@@ -179,6 +226,11 @@ const onDeleteAliasError = (error: string) => {
         margin-block-end: 0;
       }
     }
+
+    .catch-all-badge {
+      /* Mixing hexcode with transparency... */
+      background-color: color-mix(in srgb, var(--colour-accent-purple), transparent 80%);
+    }
```
davinotdavid, line 233: I wonder if we should cut a ticket to add this to services-ui as a specific variant of the Badge component?

https://thunderbird.github.io/services-ui/?path=/story/components-badge--subscription

IMHO I prefer having the flexibility in the services side to update colors instead of having a fixed set in services-ui though!

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
```diff
@@ -121,11 +165,8 @@ const onDeleteAliasError = (error: string) => {
       </template>
     </notice-bar>
 
-    <email-alias-form
-      v-if="emailAliases.length < aliasLimit"
-      :all-domain-options="allDomainOptions"
-      @add-alias="onAddAlias"
-    />
+    <email-alias-form v-if="emailAliases.length < aliasLimit" :all-domain-options="allDomainOptions"
```
davinotdavid: I think this `v-if` needs some adjustment because if I reach the aliasLimit, I think we still be able to do a catch-all right? However, this blocks the add alias button to show up.

Image

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
```diff
@@ -21,39 +21,62 @@ const aliasLimit = window._page?.maxEmailAliases;
...
+
 const onAddAlias = async (emailAlias: string, domain: string) => {
   isAddingEmailAlias.value = true;
 
   try {
     const response = await addEmailAlias(emailAlias, domain);
+    const isCatchAll = emailAlias === '*' || emailAlias === '';
+    if (isCatchAll) {
+      // This is cheating but it's the easiest way to do it.
+      emailAlias = '';
```
davinotdavid, line 69: Heh that's cool actually

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
davinotdavid: Similarly, if there's aliases available but the catch-all limit is used, it is also hiding this button:

Image

davinotdavid (changes requested):
Nice! Happy path works well (adding a catch-all, the count being updated once its added, validation on attempting to add a second catch-all for either the ALLOWED_DOMAINS or the custom domain that already has a catch-all).

However, I believe the calculation / logic to show and hide the "Add alias" button / form needs some tweaking.

--

Also, we can probably punt this to another ticket (to be tackled sooner than later) but since the catch-all is closely related with the custom domains, it would be great if we have a Pinia store for the custom domains so that we can update the catch-all calculation on the fly.

It is a little misleading to see `0 of 0` at the beginning / after FTUE and after verifying the domain, it is still `0 of 0` until a full page refresh.

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
MelissaAutumn: I didn't want to block this with another services-ui update. I'm hopeful once we get a design token to css pipeline setup we can specify it in RGB like: ` --colour-accent-whatever: 123, 200, 64;` then we can just do `rgba(var(--colour-accent-whatever), 0.2)`.

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
MelissaAutumn: D'oh

MelissaAutumn:
I'll punt the store to round 2 of this ticket.

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/EmailAliases.vue
```diff
@@ -74,11 +97,30 @@ const onDeleteAliasError = (error: string) => {
...
-          aliasUsed: emailAliases.length, aliasLimit
-        }) }}
-      </p>
+      <div class="email-aliases-totals">
+        <p class="email-aliases-count-text">
+          {{ t('views.mail.sections.emailSettings.emailAliasesDescriptionTwo', {
+            aliasUsed: usedAliases, aliasLimit
+          }) }}
+        </p>
+        <span class="email-aliases-count-text" v-if="customDomains.length > 0">
```
davinotdavid, line 106: Nice!

davinotdavid (approved):
lgtm!

UI nit for the follow up PR is to right-align the aliases used / catch-all used

Image

PR #766, opened 2026-04-29
Sign up check for recovery email in use

### LINKED ISSUE #759: Add a debounced api check for recovery email
Currently if there's a keycloak error during the sign up flow about the recovery email matching an existing one it requires a submit to catch. We should add a rate limited api function that queries our db for existing recovery emails, and use a debounced function (similar to our username check) to see if this recovery email is already in use. 

This is part of the feedback from the monday walkthrough.

davinotdavid:
## Description of changes

Added a throttled new endpoint `POST api/v1/mail/is-recovery-email-in-use/` to preemptively check if the verification email already belongs to an existing user before clicking the submit button.

## Screenshots / Video

https://github.com/user-attachments/assets/aebe7354-82e9-4ca2-9401-14dd942a2b12

## Related issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/759

### DIFF
--- a/src/thunderbird_accounts/mail/api.py
+++ b/src/thunderbird_accounts/mail/api.py
@@ -11,10 +11,15 @@ from django.utils.translation import gettext_lazy as _
 
 
 from thunderbird_accounts.authentication.models import User
+from thunderbird_accounts.mail.utils import is_address_taken
 
 
 class UsernameAvailableThrottle(UserRateThrottle):
     scope = 'is_username_available'
+
+
+class RecoveryEmailInUseThrottle(UserRateThrottle):
+    scope = 'is_recovery_email_in_use'
 
 
 @api_view(['POST'])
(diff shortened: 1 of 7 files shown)


src/thunderbird_accounts/mail/api.py
```diff
@@ -45,3 +50,17 @@ def is_username_available(request: Request):
...
+@api_view(['POST'])
+@permission_classes([AllowAny])
+@throttle_classes([RecoveryEmailInUseThrottle])
+def is_recovery_email_in_use(request: Request):
+    recovery_email = request.data.get('recoveryEmail')
+    if not recovery_email:
+        raise ValidationError(_('You need to enter a recovery email.'))
+
+    if is_address_taken(recovery_email):
+        raise ValidationError(_('User exists with same email.'))
```
davinotdavid, line 64: [This is the same check & error message that we are doing upon submit in the sign-up flow](https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/authentication/api.py#L67-L68).

Just confirming, is that what we were looking for?

MelissaAutumn (commented):
We might not need this if I can safely switch some keycloak settings. Apologies for not catching that before writing the ticket up, it's been a week.

MelissaAutumn:
Closing this as it's not needed anymore.

PR #772, opened 2026-04-30
Compile legal docs into app/legal templates folder. Read it via the django template loader.

### LINKED ISSUE #773: Update legal docs folder structure
While the ToS is working locally, there's a bug in stage / prod where the docker container can't reference the `/assets/` folder to serve the .html files for the legal docs.

MelissaAutumn:
The hope of using the django template loader would be that it'll handle any automatic file remapping. It was a bit annoying though.

Works on my machine! I regenerated the legal docs via command. 

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/773


src/thunderbird_accounts/legal/tests/test_views.py
```diff
@@ -23,37 +25,38 @@ def setUp(self):
 
 class ReadLegalContentTestCase(TestCase):
     def setUp(self):
-        self.content_dir = settings.ASSETS_ROOT / 'legal' / 'tos' / 'v2.0'
+        self.legal_app_path = apps.get_app_config("legal").path
+        self.version = 'v999.0'
```
MelissaAutumn, line 29: so we don't conflict with any future docs

src/thunderbird_accounts/legal/tests/test_views.py
```diff
@@ -23,37 +25,38 @@ def setUp(self):
...
         (self.content_dir / 'de.html').write_text('<h1>AGB</h1>', encoding='utf-8')
         (self.content_dir / 'en.html').write_text('<h1>TOS</h1>', encoding='utf-8')
-        result = _read_legal_content('tos/v2.0', 'de')
+        result = _read_legal_content(f'tos/{self.version}', 'de')
         self.assertEqual(result, '<h1>AGB</h1>')
 
     def test_falls_back_to_default_for_unsupported_locale(self):
         (self.content_dir / 'en.html').write_text('<h1>TOS</h1>', encoding='utf-8')
-        result = _read_legal_content('tos/v2.0', 'fr')
+        result = _read_legal_content(f'tos/{self.version}', 'not-a-language')
```
MelissaAutumn, line 48: samesies here. We'll eventually support french.

src/thunderbird_accounts/settings.py
```diff
@@ -190,22 +190,34 @@ def before_send(event: Event, hint: Hint) -> Event | None:
...
+    ("django.template.loaders.cached.Loader", [
+        "django.template.loaders.filesystem.Loader",
+        "django.template.loaders.app_directories.Loader",
+    ])
+]
+if IS_TEST:
+    loaders = [
+        "django.template.loaders.filesystem.Loader",
+        "django.template.loaders.app_directories.Loader",
+    ]
```
MelissaAutumn, line 203: cached loader was breaking tests.

davinotdavid: TIL, thanks for that really 🙏

src/thunderbird_accounts/legal/management/commands/convert_legal_docs.py
```diff
@@ -24,21 +26,29 @@ class Command(BaseCommand):
...
         if not md_files:
             self.stdout.write(self.style.WARNING('No markdown files found.'))
             return
 
         for md_file in md_files:
-            html_file = md_file.with_suffix('.html')
+            file_name = md_file.stem
+            # Retrieve the last two folders (as of writing this it's 'privacy' -or- 'tos' and 'v1.0')
+            type, version =  md_file.resolve().parent.parts[-2:]
+            html_file = Path(legal_dir_out, type, version, f'{file_name}.html')
```
davinotdavid, line 51: In case we don't have the `vX.X` folder (for instance for a `v2.0`), this would error in attempting to convert the .md to .html as the folder would exist so I believe we should:

```suggestion
            html_file = Path(legal_dir_out, type, version, f'{file_name}.html')
            html_file.parent.mkdir(parents=True, exist_ok=True)
```

davinotdavid (changes requested):
Nice! Thank you so much for working on this 🙏 

Would you mind also updating the [README.md](https://github.com/thunderbird/thunderbird-accounts/blob/main/README.md)?

davinotdavid (approved):
lgtm! tyvm!

works on my machine too so we have at least two machines working (jk) 🚀

PR #774, opened 2026-04-30
Move sign up's email validation to utils, and call it in add_alias

### LINKED ISSUE #85: Sanitize the email address on sign-up
As suggested by Mel, we should further sanitize the email address input (no @'s, no spaces, etc..) provided in the sign-up form.

MelissaAutumn:
Fixes #85 

Technically email aliases are limited to like 40 characters via frontend validation but we can sort that out later.


davinotdavid (approved):
lgtm!

PR #779, opened 2026-05-01
Add Android mobile to nightly tests

### LINKED ISSUE #740: Port existing E2E tests to mobile
Currently the E2E tests run only on desktop browsers; port the E2E tests over to work on mobile (real Android and iOS devices on BrowserStack). Create the corresponding GHA workflows. Follow the structure used in the Appointment E2E test jobs where each platform has it's own workflow which allows us to manually trigger the tests on any single platform when needed (instead of only being able to manually trigger the tests on all platforms at once).

There aren't that many Accounts E2E tests yet but expanding to mobile now will allow us to add new E2E tests on both desktop and mobile in the future.

rwood-moz:
## What changed?

Adding Android mobile devices to the nightly E2E tests job that runs on BrowserStack.

## Why?

To expand the platforms that the nightly tests run on (currently they only run on desktop browsers).

## Limitations and Notes

Scheduled to run nightly via GHA. Added a single GHA workflow file for each platform so that we can easily manually trigger the test job on one single platform if need be.

## Applicable Issues

Fixes #740.

## Screenshots

BrowserStack links running the nightly test jobs (all tests passed):

- [Android Chrome](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Android+Chrome/19?tab=sessions&testListView=spec) on Google Pixel 10
- [Desktop Firefox](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Nightly+Tests+Firefox+Desktop/101?tab=sessions) (regression test since moved the desktop tests to new location)

### DIFF
--- a/.github/workflows/e2e-mobile-nightly-android.yml
+++ b/.github/workflows/e2e-mobile-nightly-android.yml
@@ -1,0 +1,56 @@
+name: e2e-mobile-nightly-android
+
+concurrency:
+  group: e2e-mobile-nightly-android
+  cancel-in-progress: true
+
+on:
+  schedule:
+    # run every day at 3:30am UTC (10:30PM EST)
+    - cron:  '30 3 * * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+permissions:
+  contents: read    # This is required for actions/checkout
+
+jobs:
+  mobile-nightly-android:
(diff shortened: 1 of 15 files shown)


aatchison (approved):
let's go for it

PR #781, opened 2026-05-01
Restrict some words on sign up

### LINKED ISSUE #780: name reserve list not hooked up to sign up

MelissaAutumn:
Fixes #780 

Further clean up items:
* Normalize http status codes (alias uses 403, sign up uses 400)
* Swap alias to use is_email_reserved

### DIFF
--- a/src/thunderbird_accounts/authentication/reserved.py
+++ b/src/thunderbird_accounts/authentication/reserved.py
@@ -1,7 +1,7 @@
 import re
 
 # brand related names, and also help/support
-brands = '(thunderbird|thunderbirdpro|mzla|mozilla|firefox)?'
+brands = '(thunderbird|tbpro|thundermail|thunderbirdpro|mzla|mozilla|firefox)?'
 
 # Brand names plus common tokens
 names_with_brands = [
--- a/src/thunderbird_accounts/authentication/reserved.py
+++ b/src/thunderbird_accounts/authentication/reserved.py
@@ -6,6 +6,7 @@ brands = '(thunderbird|thunderbirdpro|mzla|mozilla|firefox)?'
 # Brand names plus common tokens
 names_with_brands = [
     brands + '?',
+    brands + '[_|-]?admin?',
     brands + '[_|-]?support?',
     brands + '[_|-]?customer[_|-]?support?',
     brands + '[_|-]?help?',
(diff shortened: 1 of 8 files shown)


src/thunderbird_accounts/authentication/reserved.py
davinotdavid: I think this should also have `thundermail` in it as it is possible to create `thundermail@thundermail.com` currently

src/thunderbird_accounts/authentication/tests/test_api.py
```diff
@@ -97,6 +97,24 @@ def test_user_already_exists(self, mock_import_user: MagicMock):
...
+    def test_user_using_reserved_emails(self):
+        """Test to ensure that a user does not use a reserved username"""
+        User(email=self.wait_list[0], username='hello@example.org').save()
+
+        # All examples from reserved.py
+        for local_part in ['thunderbird_admin', 'admin', 'postmaster', 'root']:
+            response = self.client.post(
+                '/api/v1/auth/sign-up/',
+                self.make_sign_up_data(email=self.wait_list[0], partialUsername='thunderbird_admin'),
+            )
```
davinotdavid, line 109: Shouldn't this be tested against the `reserved.py` entries? Also, we are not using `local_part` in this loop... shouldn't this be

```suggestion
        for local_part in ['thunderbird_admin', 'admin', 'postmaster', 'root']:
            response = self.client.post(
                '/api/v1/auth/sign-up/',
                self.make_sign_up_data(email=self.wait_list[0], partialUsername=local_part),
            )
```

davinotdavid (changes requested):

src/thunderbird_accounts/authentication/tests/test_api.py
MelissaAutumn: Fixed. There's already tests for the reserved function these are just some examples from there.

src/thunderbird_accounts/authentication/reserved.py
MelissaAutumn: Good point I'll add it to the list

src/thunderbird_accounts/authentication/tests/test_api.py
davinotdavid: Ah true, for future me the tests are here: src/thunderbird_accounts/authentication/tests/test_utils.py

davinotdavid (approved):
lgtm! ty! 🚀

Issue #792, opened 2026-05-04
Mailchimp integration can't differentiate recovery email vs thundermail


davinotdavid:
After the OIDC token changes / Keycloak field re-mappings, the Mailchimp integration broke as tags that were previously separated between recovery email vs thundermail now are all going to the thundermail address.

MelissaAutumn:
sneaking in here to mention, i think accounts adds the users email field to the mailchimp list as a post subscription task. 

Please make sure to update that too! if it actually sends over the username, then you can ignore it.

PR #793, opened 2026-05-04
Add recovery_email to User model for Mailchimp integration

davinotdavid:
## What changed?

- New migration to add `recovery_email` to `User` model
- Using `User.recovery_email` in the Mailchimp integration for proper tagging

## Why?

We recently made it so that the `User.email` field is remapped in Keycloak as the `username` which is the `thundermail.com` address. However, we were relying on that to be the `recovery_email` for the Mailchimp integration.

To avoid confusion between field names, and for solving this integration problem for now, adding a discrete `recovery_email` field in `User` should suffice. In case we want to further remove it, it would be a clean removal

## Applicable Issues

Solves https://github.com/thunderbird/thunderbird-accounts/issues/792

### DIFF
--- a/src/thunderbird_accounts/authentication/api.py
+++ b/src/thunderbird_accounts/authentication/api.py
@@ -40,7 +40,10 @@ def sign_up(request: Request):
     from thunderbird_accounts.mail.utils import is_address_taken
 
     data = request.data
+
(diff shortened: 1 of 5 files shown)


malini (approved):
lgtm!

PR #797, opened 2026-05-05
Allow applying discount via allowlist

### LINKED ISSUE #798: Allow applying discount via allowlist
It would be great to pre-apply the discounts so we don't need to send them to users individually.

kewisch:
## What changed?

This patch will allow us to set a discount id in the allowlist which gets pre-populated in the paddle checkout.
I've used AI to implement this feature, however I understand what it changes and why. The tests are more generated by, I've not reviewed them line-by-line.

## Why?

This makes it easier for us to set up certain users with a discount pre-applied, so they don't have to fiddle with it.

## Limitations and Notes

We're not using the per-transaction custom discount codes, but instead the standard generated ones. This could be improved on later.

Fixes #798


davinotdavid (approved):
lgtm! tested locally and works well!

PR #806, opened 2026-05-06
Allow fractional digits for plan pricing

### LINKED ISSUE #800: Add floating points to dashboard subscription price
Since we are providing discount codes, we won't necessarily have whole numbers to display for the monthly subscription cost. We should display the price to 2 floating points

davinotdavid:
## What changed?

- Updated the `Intl.NumberFormat` call to allow for m

## Why?

Not showing the fractional values may cause confusion as either flooring, ceiling and truncating the value for the sake of showing a single digit would not be accurate.

## How to test

The easiest way to test this without having to tamper with the Paddle sandbox is: when running locally, update the `priceInDollars` variable here `assets/app/vue/views/DashboardView/utils.ts`

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/800

## Screenshots

[if plan is 104 / month -> 104 / 12 = 8.66667]

image

[if plan is 96 / month -> 96 / 12 = 8]

image

[if plan is 28.8 / month ->  28.8 / 12 = 2.4 (adds a trailing 0)]

image

### DIFF
--- a/assets/app/vue/views/DashboardView/utils.ts
+++ b/assets/app/vue/views/DashboardView/utils.ts
@@ -60,12 +60,14 @@ export const formatSubscriptionData = (
     priceInDollars = priceInDollars / 12;
   }
 
+  const fractionDigits = Number.isInteger(priceInDollars) ? 0 : 2;
+  
   // n is not doing a good job in letting us specify the actual currency...so use javascript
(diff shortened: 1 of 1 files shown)


kewisch (approved):
LGTM Thanks!

PR #808, opened 2026-05-06
Enforce min-length username rule during sign-up flow

### LINKED ISSUE #807: Enforce min-length username during sign-up flow
Even though we are enforcing a min length of 3 characters for email aliases (except for custom domain aliases), we are not doing so for the sign up flow.

We should enforce this during sign up too!

davinotdavid:
## What changed?

[Frontend changes]

- Added a check during `onMounted` of the first step in the sign-up process (username) in case we pre-fill the field programmatically

[Backend changes]

- Updated `validate_email` logic to enforce min_length
  - The min_length can be overridden since the `validate_email` is used for both the sign up flow and the email aliases. For the email aliases in custom domains, it should still allow for < 3 characters.
- Added a `MinLengthValidator` to the `username` field in the `User` model
- Updated logic to allow the upper bound of the max length (we could only do 149 instead of 150 characters for username) to be accepted
- Added tests

## Why?

The HTML native `minlength` validation works well, except that it does not run when we pre-fill the field programmatically. When doing so, I needed to manually check for the length.

Just adding the `minlength` would still allow for folks to type a 1 character username, reload the page, and since the username gets stored in session storage and populated programatically, they would still be able to click the submit.

Added guards in the backend as well in case there are attempts to bypass that.

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/807

## Screenshots / GIFs

[Normal cases]

https://github.com/user-attachments/assets/71c507dc-b39a-4303-bc85-564e4e7b8d5f

[Reload case / Username saved in session storage]

https://github.com/user-attachments/assets/7c9a2168-784b-4f51-9eeb-7e15c0bc5166

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,13 +248,13 @@ Make sure that the containers are already running.
 To run all tests:
 
 ```shell
-docker compose exec backend uv run python manage.py test
+docker compose exec accounts uv run python manage.py test thunderbird_accounts
 ```
 
 To run tests for a specific module:
 
 ```shell
-docker compose exec backend uv run manage.py test thunderbird_accounts.client.tests
+docker compose exec accounts uv run python manage.py test thunderbird_accounts.mail.tests
 ```
 
 ## Running the E2E tests
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -294,7 +294,9 @@
             "emailSuggestions": "Try one of these:",
             "usernamePlaceholder": "yourname",
             "usernameHelp": "This will become your email. e.g. yourname{domain}",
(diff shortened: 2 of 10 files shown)


rwood-moz (approved):
LGTM and it worked well when I tested it out locally. One small suggestion, if you could please update the README with the correct command to run the tests (the name of the container is now `accounts`) i.e.:

```
 docker compose exec accounts uv run python manage.py test thunderbird_accounts
```

davinotdavid:
@rwood-moz thank you for the review and suggestion! Good call on updating the README, will do it as part of this PR too

PR #814, opened 2026-05-07
Run E2E tests on PRs and wait till deploys reach steady state.

Sancus:


test/e2e/playwright.config.ts
```diff
@@ -8,6 +8,13 @@ import dotenv from 'dotenv';
 import path from 'path';
 dotenv.config({ path: path.resolve(__dirname, '.env') });
 
+const ciFirefoxLaunchOptions = process.env.CI ? {
```
Sancus, line 11: This stuff is necessary or Firefox just dies trying to find localhost on runners for whatever reason.

test/e2e/tests/auth.desktop.setup.ts
```diff
@@ -7,21 +7,14 @@ import { navigateToAccountsHubAndSignIn } from '../utils/utils';
...
-
-  console.log('created auth storage directory');
-  // when use storageState in browserstack yml, browserstack requires the file to exist already even on the
-  // first time the auth-setup step is run; so must create an emtpy user.json file here
-  const filepath = path.join(directoryPath, 'user.json');
-  const emptyJsonObject = {};
-  const jsonString = JSON.stringify(emptyJsonObject, null, 2); // The '2' adds indentation for readability
-  fs.writeFileSync(filepath, jsonString);
-});
+fs.mkdirSync(directoryPath, { recursive: true });
```
Sancus: I could probably add a try block here instead, but the previous code was async and caused a potential race where Playwright could start navigation before `user.json` existed.

rwood-moz: Oh good catch! I'll update that in the E2E tests on the other repos too.

test/e2e/playwright.config.ts
rwood-moz: Ok thanks, never seen this before. Some AI info just to record some context:

```
Playwright's Firefox often fails to connect to localhost because Firefox prefers IPv6 (::1) while your local server is listening only on IPv4 (127.0.0.1), causing a connection mismatch. Disabling IPv6 in Firefox forces it to use IPv4, solving the conflict. 
Recommended Solutions:

    Change Localhost to IPv4: Update your Playwright tests to use http://127.0.0.1:port instead of http://localhost:port. This explicitly avoids the IPv6 preference.
    Disable IPv6 in Firefox (Persistent Solution):
        Open Firefox through Playwright.
        Type about:config in the address bar.
        Search for network.dns.disableIPv6.
```

.github/workflows/validate.yml
```diff
@@ -132,17 +142,32 @@ jobs:
...
+          append_env_if_set PADDLE_TOKEN
+          append_env_if_set PADDLE_API_KEY
+
+      - name: Start accounts stack via docker
         run: docker compose up -d --build -V
 
+      - name: Wait for local services
+        run: |
+          curl --retry 30 --retry-all-errors --retry-delay 2 --fail http://localhost:8087/health
+          curl --retry 30 --retry-all-errors --retry-delay 2 --fail http://localhost:9000/health/ready
```
rwood-moz, line 178: Very cool!

rwood-moz (approved):
Awesome, thanks for this @Sancus! I see the tests ran successfully on your PR, nice. LGTM (and I learned a few things!).

PR #823, opened 2026-05-08
Add email fields and username field as search_fields for UserAdmin

### LINKED ISSUE #820: Django Admin: Allow searching all email fields in user search
We need to find a user by any kind of email, not just their thundermail email. Make it so that results include recovery emails and any other email fields.

davinotdavid:
## What changed?

- Added a `search_fields` [[Django docs reference](https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields)] with the various email related fields for UserAdmin

## Why?

- It is already becoming difficult to search through existing users in Django Admin and the existing search field wouldn't match with recovery_email or last_used_email

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/820

### DIFF
--- a/src/thunderbird_accounts/authentication/admin/models.py
+++ b/src/thunderbird_accounts/authentication/admin/models.py
@@ -82,6 +82,7 @@ class CustomUserAdmin(UserAdmin):
         admin_manual_activate_subscription_features,
         admin_add_to_mailchimp_list,
     ]
+    search_fields = ('email', 'recovery_email', 'last_used_email', 'username')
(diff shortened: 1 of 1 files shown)


Sancus (approved):
Sounds good, unless the field names in this are impacted by the fix to recovery email display but I don't think they are?

davinotdavid:
@Sancus it shouldn't impact no, working on the fix for the recovery email and will test just in case before pushing a PR later on. Thanks for the review!

PR #825, opened 2026-05-08
Use recovery_email in Django Admin User forms

### LINKED ISSUE #821: Django Admin: Wrong recovery email field
Image

The recovery email is showing the wrong email in the change user form. And likely also changing the wrong one.

davinotdavid:
## What changed?

- Updated Django Admin's User to use `recovery_email` instead of `email`
- Fixed a bug on merging two dicts while updating a Keycloak user
- No changes in `ForgotPasswordView` other than lint

## Why?

We've recently added `recovery_email` for unblocking the Mailchimp integration for Early Bird and also to serve as a stable reference for a recovery email in a way that is less disruptive for the rest of the application [[PR link](https://github.com/thunderbird/thunderbird-accounts/pull/793)].
 
However, Support received a request to update the recovery_email for a user. This was already possible by updating the `email` field in the Django Admin User form but that would be overridden upon the next login (see Limitation and Notes below).

To keep a clear separation of recovery_email vs email and avoid confusion, the Django Admin User form now uses the same `recovery_email` introduced in the PR mentioned. There was a bug though that we were not applying the changed from Django Admin -> Keycloak since when merging two dicts, the ordering of the dicts were favouring the existing data instead of the updated data (this has been fixed in this PR too).

## Limitations and Notes

We are using a Keycloak Client Scope Mapper for the `email` field to be the `username` which overrides the `email` field in the Keycloak User upon login. This was done on purpose as a way to not store the recovery email in the long term (at least as far as I'm aware) and this hasn't changed.

Since we now have a separate `recovery_email`, this is a non-problem. However, if we don't want to have a separate recovery email field and want to roll back the changes (go back to using `email` as `recovery email`), we could revert this PR and [the recovery_email PR](https://github.com/thunderbird/thunderbird-accounts/pull/793)), and remove the Keycloak Client Scope Mapper.

IMHO, I think it is less confusing to have a clearly separated recovery email field though!

## How to test locally

[Pre-requisites]

- Login in Keycloak [[link](http://keycloak:8999/admin/master/console/#/master/realms)] as `admin / admin`
- Change the realm to `tbpro`
- Navigate to the Realm Settings, Login tab and change the User Registration to be `OFF` so we can sign-up using the same flow as stage / prod [[link](http://keycloak:8999/admin/master/console/#/master/realm-settings/login)]
- In `Client Scopes` [[link](http://keycloak:8999/admin/master/console/#/master/client-scopes)], select `email` -> `Mappers` tab -> `email`, and change the User Attribute to be `username` so that we have the same setup as stage / prod (this is what overrides the email field mentioned above btw)
- Login into Django Admin [[link](http://localhost:8087/admin/)] as `admin@example.org` / `admin`
- Create a Plan [[link](http://localhost:8087/admin/subscription/plan/)]
- Logout

[Test steps]

- Navigate to `http://localhost:8087/sign-up` and go through the sign-up flow as usual (your verification email will go to [Mailpit](http://localhost:8025/))
- Click in the verification link in the email and complete your account (for payment in Paddle sandbox, use the [test cards](https://developer.paddle.com/concepts/payment-methods/credit-debit-card#test-payment-method)).
- Logout & Login as `admin@example.org / admin`
- Go to Django Admin, find the user you've just signed up with and see the Recovery Email field being populated correctly

From here, you can test the Forgot Password flow, updating the user's recovery email in Django admin and trying again and verifying in Mailpit that the email is sent to the recovery email address correctly

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/821

### DIFF
--- a/src/thunderbird_accounts/authentication/clients.py
+++ b/src/thunderbird_accounts/authentication/clients.py
@@ -235,10 +235,10 @@ class KeycloakClient:
         if update_attribute_data:
             update_data['attributes'] = update_attribute_data
 
-        # Merge the two dicts together, update data being above existing data
+        # Merge the two dicts together, update data taking precedence over existing data
         update_data = {
-            **update_data,
             **user_data,
+            **update_data,
         }
 
         try:
--- a/src/thunderbird_accounts/authentication/clients.py
+++ b/src/thunderbird_accounts/authentication/clients.py
@@ -233,7 +233,10 @@ class KeycloakClient:
         )
 
         if update_attribute_data:
-            update_data['attributes'] = update_attribute_data
+            update_data['attributes'] = {
+                **(user_data.get('attributes') or {}),
+                **update_attribute_data,
+            }
 
         # Merge the two dicts together, update data taking precedence over existing data
         update_data = {
--- a/keycloak/themes/tbpro/assets/vue/views/ForgotPasswordView/index.vue
+++ b/keycloak/themes/tbpro/assets/vue/views/ForgotPasswordView/index.vue
@@ -1,4 +1,4 @@
-<script setup>
+<script setup lang="ts">
 import { TextInput, PrimaryButton, NoticeBar, NoticeBarTypes } from "@thunderbirdops/services-ui";
 import { ref, computed, useTemplateRef } from 'vue';
 import MessageBar from '@kc/vue/components/MessageBar.vue';
@@ -22,7 +22,7 @@ const usernameError = computed(() => {
 
 </script>
 
-<script>
+<script lang="ts">
 export default {
   name: 'ForgotPasswordView'
 };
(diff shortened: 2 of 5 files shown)


src/thunderbird_accounts/authentication/clients.py
```diff
@@ -235,10 +235,10 @@ def update_user(
         if update_attribute_data:
             update_data['attributes'] = update_attribute_data
```
Sancus:
```suggestion
```

src/thunderbird_accounts/authentication/clients.py
```diff
@@ -235,10 +235,10 @@ def update_user(
         if update_attribute_data:
```
Sancus, line 235: After this patch, I don't think this is safe because it will replace the entire attributes object ignoring what fields it actually contains or not. Which can completely delete attributes that weren't otherwise touched.

```suggestion
        if update_attribute_data:
            update_data['attributes'] = {
                **(user_data.get('attributes') or {}),
                **update_attribute_data,
            }
```

Sancus (changes requested):
LGTM other than the attributes issue.

We should probably have one helper function for doing keycloak puts across the whole codebase with tests so that it's harder to break.

That is not blocking for this PR but I would file an issue as keycloak puts in particular are high risk.

davinotdavid:
@Sancus oops very good catch, thanks a lot for the review! About hardening the keycloak puts, this is what's missing from this issue https://github.com/thunderbird/thunderbird-accounts/issues/346#issuecomment-4360516910 so that can be hopefully be taken care there.

PR #827, opened 2026-05-08
Add multi-factor authentication management for TOTP and recovery codes. (Fixes #810, Fixes #826)

### LINKED ISSUE #810: Implement app-based TOTP for MFA

### LINKED ISSUE #826: Implement recovery codes for MFA
Track implementation of MFA recovery codes (backup codes) for account recovery when primary MFA factors are unavailable.

Sancus:
The below documentation is AI generated, though I've made some edits, I think it's helpful to understanding what's going on here:

## Summary

Adds a self-service MFA management UI to `accounts.tb.pro`, letting users enrol an authenticator app (TOTP) and save recovery codes against their Keycloak identity. Splits cleanly into two commits:

1. **MFA functionality** (Fixes #810, #826)
2. **E2E tests** for the above (#810, #826)

The view is gated behind a `MANAGE_MFA_FEATURES` flag and the existing `FooterBar` link is still hidden, so this rolls out dark until we expose the entry point.

## Key design decisions

- **Two-step start/confirm for both TOTP and recovery codes.** Pending state lives in the Django cache; the Keycloak credential is only written after the user proves possession (TOTP) or acknowledges they saved their codes. Old recovery codes remain valid until new ones are committed, so dismissing the modal mid-flow can't lock anyone out.
- **Step-up auth on sensitive actions.** Removing or replacing an MFA credential requires a recent MFA auth (10-minute window). When stale, the API returns 403 + reauth URL; the front-end drives an OIDC step-up at `acr_values=2`, and the session timestamp only refreshes if the returned id_token actually carries that ACR.
- **After a fresh TOTP enrolment, recovery code setup immediately follows**, since that's the natural moment to also save backup codes.

### Backend (`src/thunderbird_accounts/authentication/`)

- `api.py` — REST endpoints for `start_totp_setup`, `confirm_totp_setup`, `start_recovery_codes_setup`, `confirm_recovery_codes_setup`, list/remove credentials. Includes per-action rate throttling.
- `mfa.py` — pure helpers: TOTP secret generation, recovery code generation/hashing, cache key builders.
- `clients.py` — Keycloak Admin API wrappers for credential CRUD and credential metadata reads.
- `middleware.py` — captures the `acr` claim into the session so the recent-MFA-auth check can verify it.
- `views.py` + `urls.py` — `MFAReauthenticationRequestView` (OIDC step-up entry point) and URL wiring.
- `admin/actions.py` + `templates/.../reset_totp_confirmation.html` — "Force reset TOTP credentials" admin action for ops to unblock locked-out users.
- `tests/test_mfa.py` (new) + `tests/test_admin.py` — full coverage of the API, throttles, ACR enforcement, and admin action.

### Keycloak realm (`keycloak/data/import/tbpro-realm.json`)

Adds a dedicated MFA-management step-up flow and makes `auth-otp-form` / `auth-recovery-authn-code-form` alternatives at the OTP step, so users can sign in with either factor.

### Frontend (`assets/app/vue/views/ManageMFAView/`)

- `api.ts` — typed wrappers around the new endpoints; centralises the `MFAReauthenticationRequiredError` that triggers step-up.
- `index.vue` + `components/AuthenticationMethods.vue` — the page itself: lists enrolled methods, shows setup/last-used metadata, and routes the action buttons.
- `components/ManageAuthenticatorAppModal.vue` — QR + manual-secret display and the confirm step.
- `components/ManageRecoveryCodesModal.vue` — generates codes, copy/download UI, ack-required-before-commit.
- `components/VerifyYourIdentityModal.vue` — the step-up prompt shown when recent-MFA-auth is stale.
- `features.ts` — feature flag map for which methods are enabled.

### E2E tests (commit 2)

- `test/e2e/tests/mfa.spec.ts` — two tests: TOTP enrolment + Stalwart OAuth/IMAP/SMTP XOAUTH2 verification, and the recovery-codes lifecycle (chained setup, edit-dismiss preserves, edit-confirm rotates, remove).
- `.github/workflows/validate.yml` — bootstraps the e2e environment: maps `keycloak` to `127.0.0.1` in `/etc/hosts` so Playwright's Node fetches and Firefox both see the same host the accounts container's OIDC config uses, and provisions an active subscription for the test user via a new `create_active_subscription_for_e2e_test` management command (the `/manage-mfa` page is subscription-gated).
- `playwright.config.ts` — disables Firefox's "insecure HTTP form" warning that would otherwise hang every form submit on the local HTTP stack.

## Test plan

- [x] CI green on PR #827 (it is)
- [ ] Manually exercise: enrol TOTP → confirm → recovery codes auto-chains → save → re-login with OTP works → re-login with recovery code works
- [ ] Manually exercise: try to remove TOTP after >10min idle(or logout and login again), verify step-up is required and that pressing Cancel mid-step-up doesn't break the page

Fixes #810, Fixes #826


Sancus:
Definitely not perfect, but should be ready for initial review at least. Everything works and E2E tests pass.

src/thunderbird_accounts/authentication/views.py
```diff
@@ -26,6 +28,18 @@ def start_reset_password_flow(request: HttpRequest):
     return HttpResponseRedirect(create_aia_url(KeycloakRequiredAction.UPDATE_PASSWORD))
 
 
+@method_decorator(login_required, name='dispatch')
+class MfaReauthenticationRequestView(OIDCAuthenticationRequestView):
```
MelissaAutumn, line 37: This isn't how we write django view functions fyi.

src/thunderbird_accounts/authentication/mfa.py
```diff
@@ -0,0 +1,107 @@
...
+def make_totp_code(secret: str, for_time: int | None = None) -> str:
+    if for_time is None:
+        for_time = int(time.time())
+
+    counter = int(for_time / settings.MFA_TOTP_PERIOD)
+    digest = hmac.new(secret.encode(), counter.to_bytes(8, 'big'), hashlib.sha1).digest()
+    offset = digest[-1] & 0x0F
+    truncated_hash = int.from_bytes(digest[offset : offset + 4], 'big') & 0x7FFFFFFF
+    code = truncated_hash % (10**settings.MFA_TOTP_DIGITS)
+    return str(code).zfill(settings.MFA_TOTP_DIGITS)
```
MelissaAutumn: We should be retrieving the code from keycloak itself not re-implementing TOTP codes locally.

src/thunderbird_accounts/authentication/mfa.py
```diff
@@ -0,0 +1,107 @@
...
+    if not normalized_code.isdigit() or len(normalized_code) != settings.MFA_TOTP_DIGITS:
+        return False
+
+    current_time = int(time.time())
+    for window_offset in range(-settings.MFA_TOTP_LOOK_AHEAD_WINDOW, settings.MFA_TOTP_LOOK_AHEAD_WINDOW + 1):
+        window_time = current_time + (window_offset * settings.MFA_TOTP_PERIOD)
+        if hmac.compare_digest(make_totp_code(secret, window_time), normalized_code):
+            return True
+
+    return False
```
MelissaAutumn: Again this should be done in Keycloak.

MelissaAutumn (changes requested):
One of the things with using Keycloak was to out-source authentication and security related code to a well tested and well used application. I don't want us to have to maintain TOTP generation / recovery code in accounts. I don't think that's the correct place to do so, and I don't think re-implementing TOTP from scratch is good. 

If we do for some reason want TOTP in accounts, we should be pulling in a well tested and well used library to do so instead of re-implementing it ourselves.

On the keycloak side this is mostly done, there was an issue with design ux flow that was holding up MFA iirc. Simply create a keycloak aia url (see `create_aia_url` in the authentication utils) and ship the user to a `CONFIGURE_TOTP` flow.  The user will be sent to the Keycloak totp flow and they can configure it via our keycloak template. I don't remember if there was any other problem with that atm though. Might need some testing.

Feel free and let me know if I've misinterpreted the patch.

src/thunderbird_accounts/authentication/mfa.py
Sancus: > Again this should be done in Keycloak.

After discussing in Matrix I will change this patch to pull in a Keycloak plugin they exposes REST APi endpoints for totp initiation and aurh, so we can keep the UI in accounts but accounts will no longer handle any secret generation.

assets/app/vue/views/ManageMfaView/components/AuthenticationMethods.vue
```diff
@@ -138,7 +367,7 @@ const handleEdit = (method: string) => {
   header {
     padding: 1rem 1.125rem;
     border-radius: 0.5rem 0.5rem 0 0;
-    color: #eeeef0; /* TODO: --colour-ti-base forced dark */
+    color: var(--colour-ti-base);
```
davinotdavid: The "forced dark mode" color here was on purpose since we don't really support light / dark mode in Accounts in general for now (we just had designs for actual dark mode very recently) so if your browser is in dark mode, the colour contrast is off:

Image

assets/app/vue/views/ManageMfaView/components/AuthenticationMethods.vue
```diff
@@ -21,56 +32,217 @@ interface AuthenticationMethodData {
...
+    case AUTHENTICATION_METHODS.RECOVERY_CODES:
+      return MANAGE_MFA_FEATURES.recoveryCodes;
+    case AUTHENTICATION_METHODS.RECOVERY_EMAIL_ADDRESS:
+      return MANAGE_MFA_FEATURES.recoveryEmail;
+    case AUTHENTICATION_METHODS.RECOVERY_PHONE_NUMBER:
+      return MANAGE_MFA_FEATURES.recoveryPhone;
+    default:
+      return false;
+  }
+}));
```
davinotdavid: Marking this part of the `AuthenticationMethods` as an example but I believe this component is doing way too much work and could be broken down into other smaller components / composables for easier maintenance / understanding / testing.

For example, we could have a `useMfaMethods()` composable for the logic/state!

assets/app/vue/views/ManageMfaView/components/AuthenticationMethods.vue
```diff
@@ -124,9 +306,56 @@ const handleEdit = (method: string) => {
...
+
+    <div class="remove-mfa-method-actions">
+      <link-button :disabled="isRemoving" @click="cancelRemove">
+        {{ t('views.manageMfa.actions.cancel') }}
+      </link-button>
+      <primary-button :disabled="isRemoving" @click="confirmRemove">
+        {{ t('views.manageMfa.actions.remove') }}
+      </primary-button>
+    </div>
+  </generic-modal>
```
davinotdavid: These two generic-modals are nearly identical so we can probably extract this into a `RemoveMfaMethodModal` component

davinotdavid (commented):
Sorry for taking so long to review! I've tried setting it up locally and it works, nice!

I see that the implementation is going to move to Keycloak so I'll re-review a bit deeper once that's done. Here's some comments on the frontend side of things.

Other than that, I think before we actually un-comment the manage mfa link in the footer we need to:

- Review the main navbar to add the link to manage-mfa there (or should it? might be weird to have a singular link centered in the navbar)
- Review all the "support related links" like the "Didn't get the code" and so on as none of them are linking to anywhere

assets/app/vue/views/ManageMfaView/components/AuthenticationMethods.vue
Sancus: I agree with this and did it for expediency but since I'm reworking the keycloak part of this I guess I will also look at breaking it down.

src/thunderbird_accounts/authentication/views.py
Sancus: I'm not entirely sure what this was referring to, but I think the use of a class-based view specifically makes sense here. Otherwise we need to reimplement a bunch of stuff from `OIDCAuthenticationRequestView.get()`.

assets/app/vue/views/ManageMfaView/components/AuthenticationMethods.vue
Sancus: This is further broken down in the new patch.

Sancus:
Note: This PR's tests will remain red until the Keycloak plugin jar is available.

Sancus:
https://mfa-accounts.thunderbird.dev/manage-mfa

I've set up a preview environment to make reviewing this easier. You can **login with your stage-thundermail credentials**. Ignore the errors about subscriptions and Stalwart, mail is not available in this preview env, only keycloak and accounts.

assets/app/vue/components/FooterBar.vue
```diff
@@ -10,6 +10,7 @@ const { t } = useI18n();
 const isAuthenticated = ref(window._page?.isAuthenticated);
 
 const navItemsAccounts = [
+  // TODO: Expose Manage MFA here after production validation is complete.
```
MelissaAutumn: Can we expose this with a feature flag? Local storage is fine, I think we use a feature flag somewhere else in code so just follow that convention, otherwise something like FF_SHOW_MFA as the key works.

assets/app/vue/components/GenericModal.vue
```diff
@@ -27,6 +27,7 @@ defineExpose({
   open: () => {
     modal.value.showModal();
   },
+  close: handleClose,
```
MelissaAutumn, line 24: Thx!

MelissaAutumn (changes requested):
(This is a brief review of the code live on stage btw.)

First off, thank you! It's not an easy lift, but thanks for the plugin and the integration into accounts.

Now then:

1. Modal's close button is weirdly tall
image

2. You can close the recovery codes modal before ack'ing that you saved them. If you do then the codes are still saved. Ideally it doesn't save these unless you ack them. I feel like this is going to be a pain to fix.

3. On the MFA modal the "Can't Scan?" and "Didn't get the code?" buttons don't do anything.

4. Support links don't do anything.

5. The MFA modal's error message doesn't have enough padding on the bottom. Ideally you'd use the html5 errors / the text input's own error property. See an example here: https://thunderbird.github.io/services-ui/?path=/story/components-textinput--required / https://thunderbird.github.io/services-ui/?path=/docs/components-textinput--docs

6. It's too easy to accidentally close the setup modals. I kept doing it while switching tabs to add the setup code to 1pass.

7. I can simply `X` out of the "Verify your identity" modal that pop-ups. If it's on page-load I feel like we can just use a "Back" button instead. Maybe design has opinions here? 

9. The step up completes but I still get the "Verify your identity" pop up.

assets/app/vue/components/FooterBar.vue
MelissaAutumn: Oh I see the route just works fine. Might still be a good idea to throw those behind feature flags though.

Sancus:
>     1. Modal's close button is weirdly tall
> image

will fix

>     2. You can close the recovery codes modal before ack'ing that you saved them. If you do then the codes are still saved. Ideally it doesn't save these unless you ack them. I feel like this is going to be a pain to fix.
>     6. It's too easy to accidentally close the setup modals. I kept doing it while switching tabs to add the setup code to 1pass.

I'm just going to remove the ability to click out of these entirely. I don't think modals should offer that option unless they're purely informational, which none of these are(and purely informational modals are bad anyway).

I'm open to removing the X from the recovery codes modal specifically, but I don't think it's a big deal? If you really do click that intentionally you can always remove the codes and regenerate them, and you can regenerate them at any time in the future as long as you have your TOTP. Feels minor.

>     3. On the MFA modal the "Can't Scan?" and "Didn't get the code?" buttons don't do anything.

- "Didn't get the code?" doesn't make any sense for TOTP so I have removed it.
- Not even sure what "Can't Scan?" is supposed to link to. Maybe a support article? I can remove it as well or merge it with the manual code display option.

>     4. Support links don't do anything.

^ AFAIK there are no articles for any of these to link to yet. But I don't think that should block this PR getting merged, we can add those whenever support has them ready. It's early bird, we shouldn't even HAVE any of this documentation in there at all to be a proper software beta :)

>     5. The MFA modal's error message doesn't have enough padding on the bottom. Ideally you'd use the html5 errors / the text input's own error property. See an example here: https://thunderbird.github.io/services-ui/?path=/story/components-textinput--required / https://thunderbird.github.io/services-ui/?path=/docs/components-textinput--docs

I'll take a look at it.

>     7. I can simply `X` out of the "Verify your identity" modal that pop-ups. If it's on page-load I feel like we can just use a "Back" button instead. Maybe design has opinions here?
> 
>     8. The step up completes but I still get the "Verify your identity" pop up.

^ Keycloak was misconfigured, I've patched this and it should also properly update stage and prod when we deploy next.

assets/app/vue/components/FooterBar.vue
MelissaAutumn: I think we'll definitely want to test this in prod, so making a feature flag for routes / navbar a blocker here.

assets/app/vue/views/ManageMfaView/components/ManageRecoveryCodesModal.vue
```diff
@@ -0,0 +1,233 @@
...
+  step.value = MODAL_STEPS.CODES;
+  emit('configured', response.credentials);
+
+  // Push the codes to the user immediately so they aren't lost if the modal is
+  // dismissed; the manual download button below stays as a fallback/re-download.
+  downloadCodes();
+};
+
+const copyCodes = async () => {
+  if (!codes.value.length) return;
```
MelissaAutumn: fyi https://thunderbird.github.io/services-ui/?path=/docs/patterns-copytoclipboard--docs

MelissaAutumn: You can put the codes split into a computed and just shove it into the copyValue param.

assets/app/vue/views/ManageMfaView/components/ManageRecoveryCodesModal.vue
```diff
@@ -0,0 +1,233 @@
...
+  const blob = new Blob([codes.value.join('\n')], { type: 'text/plain' });
+  const url = URL.createObjectURL(blob);
+  const link = document.createElement('a');
+  link.href = url;
+  link.download = t('views.manageMfa.modals.manageRecoveryCodes.downloadFilename');
+  document.body.appendChild(link);
+  link.click();
+  document.body.removeChild(link);
+  URL.revokeObjectURL(url);
```
MelissaAutumn, line 65: This will never get old....how is there not a better download api yet?

assets/app/vue/views/ManageMfaView/useMfaMethods.ts
```diff
@@ -0,0 +1,236 @@
...
+  const { isLoading, errorMessage, run } = useMfaAction();
+
+  const authenticationMethods = ref<Record<AUTHENTICATION_METHODS, AuthenticationMethodData>>({
+    [AUTHENTICATION_METHODS.AUTHENTICATOR_APP]: { set: false },
+    [AUTHENTICATION_METHODS.RECOVERY_CODES]: { set: false },
+    [AUTHENTICATION_METHODS.RECOVERY_PHONE_NUMBER]: { set: false },
+    [AUTHENTICATION_METHODS.RECOVERY_EMAIL_ADDRESS]: { set: false },
+  });
+
+  const reauthenticationUrl = ref('');
```
MelissaAutumn, line 77: This function sort of fits a store pattern but not reallly. We can leave this as-is but in the future if we're handling state across different modals or views, stores can be handy.

scripts/entry-keycloak.sh
```diff
@@ -4,11 +4,24 @@
...
 fi
+KC_PID=$!
+
+# Forward termination so container stop stays graceful.
+trap 'kill -TERM "$KC_PID" 2>/dev/null' TERM INT
+
+/bin/bash "$SCRIPT_DIR/apply-mfa-config.sh" &
 
-$CMD
+wait "$KC_PID"
```
MelissaAutumn, line 27: This will also affect stage/prod containers. Do we want to roll that manually or have it auto?

src/thunderbird_accounts/authentication/admin/actions.py
```diff
@@ -212,6 +216,100 @@ def admin_sync_plan_to_keycloak(modeladmin, request, queryset):
...
+    """
+    if request.POST.get('apply') != '1':
+        context = {
+            **modeladmin.admin_site.each_context(request),
+            'title': _('Force reset TOTP credentials'),
+            'queryset': queryset,
+            'action_checkbox_name': helpers.ACTION_CHECKBOX_NAME,
+            'opts': modeladmin.model._meta,
+        }
+        return TemplateResponse(request, 'admin/authentication/user/reset_totp_confirmation.html', context)
```
MelissaAutumn, line 313: Huh, didnt know you could do this.

src/thunderbird_accounts/authentication/templates/admin/authentication/user/reset_totp_confirmation.html
```diff
@@ -0,0 +1,36 @@
...
+    {% endfor %}
+  </ul>
+
+  <form method="post">
+    {% csrf_token %}
+    {% for user in queryset %}
+      <input type="hidden" name="{{ action_checkbox_name }}" value="{{ user.pk }}">
+    {% endfor %}
+    <input type="hidden" name="action" value="admin_reset_totp_credentials">
+    <input type="hidden" name="apply" value="1">
```
MelissaAutumn, line 32: Maybe include a template comment mirroring the function comment.

MelissaAutumn (changes requested):
Questions, and such. This was from another skim but code looks okay otherwise. Github wont load the api.py diff for me??? so i guess i cannot add comments there.

Sancus:
@MelissaAutumn I've updated the front-end to use the localstorage flag and made some other minor changes, and while I was at it, wired up the dashboard UI so it should now appear as it will do normally.

`localStorage.setItem('feature.show-mfa', 'true')` and preview is up to date. https://mfa-accounts.thunderbird.dev/mail

scripts/entry-keycloak.sh
Sancus: auto, but this was taken care of in #995 and is no longer part of this patch

MelissaAutumn:
Feature flag looks good, and OTP work. However I didn't notice this during code review but the recovery code feature is a bit weird huh?

Why is it requesting specific codes and not just accepting any of the codes that haven't been used before. Is that something I just haven't noticed about recovery codes on other platforms before?

Sancus:
> Why is it requesting specific codes and not just accepting any of the codes that haven't been used before. Is that something I just haven't noticed about recovery codes on other platforms before?

Yes, it is *incredibly* weird but this is the way [keycloak implemented](https://www.keycloak.org/2025/10/recovery-codes) recovery codes. "Code #1 is requested because recovery codes are used for the first time. Remember they are requested and consumed in order."

It really pissed me off when I realized, tbh, lol.

Unfortunately I don't see any way to fix this other than to write a custom authenticator in the plugin. Which is significantly more changes than the current flows and would also require more maintenance(internal keycloak changes could break it).

I decided it wasn't worth it for this patch. We could revisit it in the future though.

src/thunderbird_accounts/authentication/api.py
```diff
@@ -28,18 +39,114 @@ class SignUpThrottle(UserRateThrottle):
...
+
+
 @api_view(['POST'])
 def get_user_profile(request: Request):
     if not request.user:
         raise NotAuthenticated()
     return Response(UserProfileSerializer(request.user).data)
 
 
+@api_view(['GET'])
```
MelissaAutumn, line 64: I'm kinda onboard with moving everything to POST's by default especially with our session/cookie auth. Another prevention against csrf attacks. (I know we have csrftokens.)

src/thunderbird_accounts/authentication/api.py
```diff
@@ -28,18 +39,114 @@ class SignUpThrottle(UserRateThrottle):
...
+@authentication_classes([SessionAuthentication])
```
MelissaAutumn, line 65: Default authentication class is `OIDCAuthentication`. Any reason we're overriding that?

src/thunderbird_accounts/authentication/api.py
```diff
@@ -28,18 +39,114 @@ class SignUpThrottle(UserRateThrottle):
...
+@permission_classes([IsAuthenticated])
```
MelissaAutumn: Default permission class is already IsAuthenticated, you can drop this.

src/thunderbird_accounts/authentication/mfa_management.py
```diff
@@ -0,0 +1,455 @@
...
+        except MfaRecoveryCodesNotFoundError:
+            raise
+        except RequestException as exc:
+            sentry_sdk.capture_exception(exc)
+            raise MfaRemoveRecoveryCodesUnavailableError() from exc
+
+        return {}
+
+    def _require_oidc_id(self) -> str:
+        oidc_id = getattr(self.request.user, 'oidc_id', None)
```
MelissaAutumn: Why are we using getattr here? The user model has oidc_id, and it's set correctly in django settings so it should be available unless it's an anonymous user (in which case this shouldn't be called.)

src/thunderbird_accounts/authentication/mfa_management.py
```diff
@@ -0,0 +1,455 @@
...
+    MFA_REST_ERROR_ALREADY_CONFIGURED,
+    MFA_REST_ERROR_TOTP_NOT_CONFIGURED,
+    has_recent_mfa_management_auth,
+    make_pending_totp_cache_key,
+)
+from thunderbird_accounts.authentication.middleware import refresh_user_access_token
+
+
+# mozilla-django-oidc stores the user's access token here when OIDC_STORE_ACCESS_TOKEN is on.
+OIDC_ACCESS_TOKEN_SESSION_KEY = 'oidc_access_token'
```
MelissaAutumn: You can import `from thunderbird_accounts.authentication.middleware import OIDC_ACCESS_TOKEN_KEY`

src/thunderbird_accounts/authentication/mfa_management.py
```diff
@@ -0,0 +1,455 @@
...
+
+# Refresh slightly before the real expiry so a token that's valid now but dies mid-request
+# isn't forwarded and rejected.
+_ACCESS_TOKEN_EXPIRY_LEEWAY_SECONDS = 30
```
MelissaAutumn: Probably should be a django setting value

src/thunderbird_accounts/authentication/mfa_management.py
```diff
@@ -0,0 +1,455 @@
...
+class MfaSessionExpiredResponseRequired(MfaManagementError):
+    pass
+
+
+class MfaReauthenticationRequired(MfaManagementError):
+    pass
+
+
+@dataclass(frozen=True)
+class MfaMethodsResult:
```
MelissaAutumn, line 116: I'm okay with dataclasses, but wanted to point out if it's exclusively for drf then we also have access to serializers https://www.django-rest-framework.org/api-guide/serializers/.

src/thunderbird_accounts/authentication/clients.py
```diff
@@ -36,12 +44,28 @@ class RequestMethods(enum.StrEnum):
...
+class TotpSetupResponse(TypedDict, total=False):
+    secret: str
+    encodedSecret: str
+    otpAuthUri: str
+    digits: int
+    period: int
+    algorithm: str
+
+
+class RecoveryCodesResponse(TypedDict, total=False):
```
MelissaAutumn, line 56: We use data classes in mfa_management.py. Do you want to settle one wrt typeddict vs dataclass?

src/thunderbird_accounts/authentication/clients.py
```diff
@@ -97,6 +122,28 @@ def get_security_credentials(self, oidc_id):
...
 
+    def get_totp_credentials(self, oidc_id: str) -> list[dict]:
+        return [
+            credential
+            for credential in self.get_security_credentials(oidc_id)
+            if credential.get('type') == KEYCLOAK_OTP_CREDENTIAL_TYPE
+        ]
+
+    def delete_credential(self, oidc_id: str, credential_id: str) -> bool:
+        self.request(f'users/{oidc_id}/credentials/{credential_id}', RequestMethods.DELETE)
```
MelissaAutumn, line 133: While I sometimes don't do a great job at it, I find it's nicer if we catch request exceptions and re-raise a specific application-based exception. Kinda like how get_user has a `GetUserError` instead of just the requestexception.

src/thunderbird_accounts/authentication/clients.py
```diff
@@ -343,3 +390,98 @@ def import_user(
...
+        endpoint: str,
+        user_access_token: str,
+        method: RequestMethods = RequestMethods.GET,
+        json_data: Optional[dict] = None,
+    ) -> dict:
+        """Call the self-service provider and translate its MFA-specific failures."""
+        try:
+            response = self.request(endpoint, user_access_token, method, json_data=json_data)
+        except RequestException as exc:
+            http_response = getattr(exc, 'response', None)
```
MelissaAutumn: Seems like RequestException will always carry a response property. You can remove the getattr.
https://requests.readthedocs.io/en/latest/_modules/requests/exceptions/

src/thunderbird_accounts/authentication/mfa.py
```diff
@@ -0,0 +1,30 @@
...
+MFA_REAUTH_PENDING_SESSION_KEY = 'mfa_reauth_pending'
+PENDING_TOTP_CACHE_KEY = 'accounts:mfa:totp-setup:{user_id}'
+
+# Machine error codes returned by the keycloak-mfa-rest provider (see its MfaResource).
+MFA_REST_ERROR_ALREADY_CONFIGURED = 'mfa_already_configured'
+MFA_REST_ERROR_TOTP_NOT_CONFIGURED = 'totp_not_configured'
+MFA_REST_ERROR_STEP_UP_REQUIRED = 'step_up_required'
+
+
+def make_pending_totp_cache_key(user_id: int) -> str:
```
MelissaAutumn: User's pk is a uuid fyi

src/thunderbird_accounts/authentication/mfa_management.py
```diff
@@ -0,0 +1,455 @@
...
+        exp = int(claims['exp'])
+    except (jwt.PyJWTError, KeyError, ValueError, TypeError):
+        return False
+    return time.time() >= exp - _ACCESS_TOKEN_EXPIRY_LEEWAY_SECONDS
+
+
+def get_user_access_token(request: Request) -> str | None:
+    """Return the end user's OIDC access token for the self-service MFA provider."""
+    access_token = request.session.get(OIDC_ACCESS_TOKEN_SESSION_KEY)
+    if access_token and not _access_token_is_expired(access_token):
```
MelissaAutumn, line 148: Note that tokens can also get invalidated by keycloak. You could ping the token introspect route to see if the token is still active, but if not just be careful about assuming a token is valid after this point. (It'll still error out on the keycloak route, but frontend/backend code might not expect a token fail at that point.)

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -199,6 +247,11 @@ def authenticate(self, request, **kwargs):
         if payload:
             # New for #498
             store_tokens(self.request, access_token, id_token, refresh_token)
+            if self.request.session.pop(MFA_REAUTH_PENDING_SESSION_KEY, False):
+                # Only grant MFA management auth if Keycloak actually satisfied
+                # the requested ACR (i.e. the user completed the OTP step-up).
+                if str(payload.get('acr', '')) == settings.MFA_KEYCLOAK_ACR_VALUE:
```
MelissaAutumn: Do we need to guard against an empty mfa keycloak acr value? (server misconfiguration, or self-hosting with no need for mfa?)

MelissaAutumn (changes requested):
Some more requested changes

src/thunderbird_accounts/authentication/mfa_management.py
Sancus: Yeah, if that happens we get 401 and return `MfaSessionExpiredError`. Checking with keycloak defensively is another request and I think the code needs to be prepared to handle that error in any case, as it can happen at any arbitrary time from the perspective of accounts.

src/thunderbird_accounts/authentication/clients.py
Sancus: I've always preferred typed dicts for JSON vs data classes for internal objects, as using a data class with JSON responses means you need to deal with serialization/deserialization and it's also less flexible in terms of the attributes that can be present. So that's why this is used here.

If I had to pick one I'd probably make the data class a typed dict as that's a more minimal change than using a data class for external JSON, which gets kind of annoying.

This would all be moot if we used Pydantic though which we probably would benefit from since accounts is pretty data/schema heavy.

src/thunderbird_accounts/authentication/api.py
Sancus: Huh, using POST even for just reads? That feels pretty weird to me, never done that before, but I don't feel super strongly about it.

src/thunderbird_accounts/authentication/middleware.py
Sancus: True, we should. Though if you're self-hosting with no MFA then you'd always have MFA management auth from the back-end's perspective. But this is an easy check.

src/thunderbird_accounts/authentication/api.py
Sancus: This is an artifact of using DRF views. The default for them is bearer token auth, but we don't have that here, so we need to explicitly set authentication as session/cookie auth. This isn't necessary for regular non-DRF django views that use `@login_required`.

These don't necessarily have to be drf views but we gain the throttling stuff and whatnot so it doesn't seem worth changing.

src/thunderbird_accounts/authentication/clients.py
Sancus: The application-based exceptions are handled by the callers in this case though, in mfa_management.py -- If there's a requests error here it becomes one of the MfaAuthenticator errors depending on the case. Delete credential doesn't have enough info to itself determine the right type of MFA error I don't think.

MelissaAutumn (approved):

PR #833, opened 2026-05-12
Add `UsernameBlockListEntry` and tie it to is_reserved, and the admin panel. (Fixes #831)

### LINKED ISSUE #831: Allow the username block list to be adjustable via the admin panel
We should have an easy way to add names, patterns, etc... to the username block list. This shouldn't require a code deployment either.

So we should create a user model that our list will check along-side the hardcoded list. This should probably be cached so we're not pulling it from the db every check.

MelissaAutumn:
Fixes #831 

I don't like the part where I pull all entries and check them sequentially, but this works for now. 

I didn't use a regex lookup so we could simply use `*` for wildcard entries, the function also is case insensitive which is also good.


markstos:
(I pushed an amended commit to resolve the conflicts).

markstos (approved):
For what it's worth, The Ghost project also uses a two-tier system
for blocking. In their case it's for domains they want to ban from
registering because they are associated with spamming. They offer
both a system level defined in configuration as well as a user-editable
level stored in a database. It seems like a sensible approach here, too.

Security and validation here looks good enough for a admin-only feature for trusted users, but there some extra safeguards worth considering:

- Add max length guardrail to the field, like 1024 characters. Extremely long patterns could cause regex performance problems. But if untrusted users have admin access, there are bigger problems. Minor concern.
- I tested various edge cases which were valid, but it's not clear why need to support:          
  - Tested "**"
  - Tested "?" and "[test]", as both these are treated specially by `fnmatch`. It appears these additional characters are also treated specially as part of the regex, but it's not clear that we need to bother excluding them either. This variant of regex is forgiving and I couldn't find any combination which could cause a crash. Again, with trusted admins this is not a great concern.

## QA Log

 - Read the docs for https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch
 - Reviewed code
    - Checked that DB index is present.
-  Tested migration and /admin actions manually.
    - Used `docker compose exec accounts ./manage.py migrate` to run migration.
    - Shows up in admin area.
    - Tested adding entries with various fnmatch edge cases

PR #835, opened 2026-05-13
Show warning in Contact Support form (/contact) when user is not on allow list 

### LINKED ISSUE #834: Only allow users in allow list to submit contact form responses
The Support team is currently noticing an uptick in tickets coming from the `/contact` form from users that are not in the allow list about being able to sign in so we should prevent them from submitting while showing a message to encourage them to join the wait list instead.

This will reduce the strain in the Support team but also improve UX for folks that may not be aware of the waitlist.

cc @lisajill

davinotdavid:
## What changed?

[Frontend]
- Added a warning `notice-bar` in case the user if the email typed is not in the allow list
- Added a throttled call on a _valid_ email input to check if the email is on the allow list

[Backend]
- Updated logic on our `is_email_in_allow_list` function to check for `recovery_email` as well
- New `POST api/v1/contact/check-email-is-on-allow-list/` endpoint that just calls `is_email_in_allow_list`
- Added tests for `is_email_in_allow_list` and for the new endpoint

> [!NOTE]
> We are still allowing any user to submit the contact form, that's by design / requested by the Support team

## Why?

The Support team has been suffering from receiving tickets that are non-actionable, which causes unnecessary work on their end and unnecessary frustration for the (future? potential?) users: tickets coming from folks that are not in the allow list, questioning why they can't sign-in.

## Limitations and Notes

~~This has not been vetted by the Design team just yet and it is a mitigation / band-aid to help the support team for now. We shall discuss further if this is needed later on and this may be removed in a future iteration.~~

~~Copy / UX is pending revision but I figure that majority of the logic will be the same.~~

Solution confirmed by design & support!

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/834

## Screenshots / GIFs

https://github.com/user-attachments/assets/91caf4c0-9f01-4bab-8fe4-11fedf4d84b1

### DIFF
--- a/src/thunderbird_accounts/settings.py
+++ b/src/thunderbird_accounts/settings.py
@@ -589,3 +589,6 @@ VERIFY_PRIVATE_LINK_SSL = True if os.getenv('VERIFY_PRIVATE_LINK_SSL', 'true').l
 
 # For Appointment's CalDAV auto-setup
 APPOINTMENT_APP_PASSWORD_PREFIX: str = 'appointment-caldav-setup-'
+
+# For contact support form allow list-less users
+CONTACT_SUPPORT_ONLY_FOR_ALLOW_LISTED_USERS = False
--- a/src/thunderbird_accounts/settings.py
+++ b/src/thunderbird_accounts/settings.py
@@ -274,7 +274,11 @@ REST_FRAMEWORK = {
         'mozilla_django_oidc.contrib.drf.OIDCAuthentication',
         'rest_framework_simplejwt.authentication.JWTAuthentication',
     ],
-    'DEFAULT_THROTTLE_RATES': {'is_username_available': '30/minute', 'sign_up': '10/minute'},
+    'DEFAULT_THROTTLE_RATES': {
+        'is_username_available': '30/minute',
+        'sign_up': '10/minute',
+        'check_email_is_on_allow_list': '10/minute',
+    },
 }
 
 REDIS_URL = os.getenv('REDIS_URL')
--- a/src/thunderbird_accounts/settings.py
+++ b/src/thunderbird_accounts/settings.py
@@ -591,4 +591,4 @@ VERIFY_PRIVATE_LINK_SSL = True if os.getenv('VERIFY_PRIVATE_LINK_SSL', 'true').l
 APPOINTMENT_APP_PASSWORD_PREFIX: str = 'appointment-caldav-setup-'
 
 # For contact support form allow list-less users
-CONTACT_SUPPORT_ONLY_FOR_ALLOW_LISTED_USERS = False
+CONTACT_SUPPORT_ONLY_FOR_ALLOW_LISTED_USERS = True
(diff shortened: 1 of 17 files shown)


MelissaAutumn (commented):
Can we move the logic to the backend's submit api route? I think we're proxying the submission anyways. I just want to avoid adding api routes that let's users ping our available users. 

Also we should have this as a backend flag (static settings, no env needed) in case support wants to enable anonymous support messages in the future.

davinotdavid:
@MelissaAutumn hmm so talking with the design & support folks, they mentioned that it would be a bad experience to let the user fill in all of the contact support form and hit submit before we actually show the error message so we can't move it to the submit button unfortunately

Although it is not very ideal to allow folks to ping it, we also already expose the username available check under similar throttling conditions. Is there a more suitable compromise code-wise?

Will add the setting though!

lisajill (approved):

src/thunderbird_accounts/settings.py
```diff
@@ -585,3 +589,6 @@ def oidc_logout(request):
 
 # For Appointment's CalDAV auto-setup
 APPOINTMENT_APP_PASSWORD_PREFIX: str = 'appointment-caldav-setup-'
+
+# For contact support form allow list-less users
+CONTACT_SUPPORT_ONLY_FOR_ALLOW_LISTED_USERS = False
```
rwood-moz: Just confirming, we want this set to `False` by default?

rwood-moz (approved):
Just one question inline, otherwise LGTM (and great job on the tests!).

src/thunderbird_accounts/settings.py
davinotdavid: Ah, thank you for catching this! I was testing it manually and forgot to flip this to `True` before committing 😅

MelissaAutumn (changes requested):
Okay fair point there. Can you add an additional check on submit though? Folks can always bypass frontend logic so we need to double up on backend.

davinotdavid:
@MelissaAutumn Sorry, I just realized that I didn't update the PR title / description with the latest in-person discussions 😅 @lisajill and @kewisch decided not to block the submission of the form as that would block prospective users from asking questions about TB Pro so I've changed it in [this commit](https://github.com/thunderbird/thunderbird-accounts/pull/835/commits/4e3781e14416d21d4d3486a17de52e341d7a7bbe).

So we are essentially just showing the warning with a link to join the wait list if they are not in the allow list but anyone is free to still fill all of the form fields and submit as usual!

MelissaAutumn:
Should we just cut the PR and replace it with a single warning then? Or do we want the warning only if they're using a non-allow list email?

MelissaAutumn (approved):

davinotdavid:
@MelissaAutumn ty for the review! We want the warning only if they are using a non-allow list email, yes!

IIRC there was also some chats about attempting to mitigate this further by tweaking some copy + font-sizes to make it more prominent in our sign in page that we are allow list based at the moment but that will probably require some cross team conversations

PR #465, opened 2025-12-02
Remove 3 char limitation for custom domain email aliases

davinotdavid:
## Description of changes

- Renamed props related to `allowedDomains` vs `allDomainOptions` for clarity
- Added frontend validation to guard against less than 3 character email aliases for domains included in the `ALLOWED_DOMAINS` env var (`example.org` in local dev, `stage-thundermail.com` in stage and `tb.pro` + `thundermail.com` in prod)
- Added guard in the backend as well with tests in case someone tries through the API

## Screenshot / Video

https://github.com/user-attachments/assets/44a764f3-36c3-40db-92e2-554f648fd595

## Related PRs
Fixed https://github.com/thunderbird/thunderbird-accounts/issues/462

PR #841, opened 2026-05-15
feat(mail): lower MIN_CUSTOM_DOMAIN_ALIAS_LENGTH default to 2

nidhalsaidani:
## Summary

Lowers the default minimum length for custom-domain email aliases from 3 to 2 characters, and makes the validation error message dynamic so it always reflects the configured `MIN_CUSTOM_DOMAIN_ALIAS_LENGTH` value.

## Why

The current default rejects 2-letter aliases (e.g. initials like `ns@`), even though the `MIN_CUSTOM_DOMAIN_ALIAS_LENGTH` setting introduced in #465 already exists to make this tunable. Two-letter aliases are a common request from users who own short custom domains. Additionally, when an admin lowered the env var, users still saw a misleading 'must be at least 3 characters long' error because the message was hardcoded.

## Changes

- `.env.example`: example value 3 -> 2
- `src/thunderbird_accounts/settings.py`: default 3 -> 2
- `src/thunderbird_accounts/mail/views.py`: error message now interpolates `settings.MIN_CUSTOM_DOMAIN_ALIAS_LENGTH`
- `src/thunderbird_accounts/mail/tests/test_views.py`: `test_alias_input_invalid` uses the dynamic value; `test_custom_domain_alias_length_minimum_2` now expects the correct '2 characters' message

## Backwards compatibility

Deployments that want a stricter limit can keep `MIN_CUSTOM_DOMAIN_ALIAS_LENGTH=3` (or higher) in their environment. The existing `test_allowed_domain_alias_too_short` pins 3 via `override_settings` and continues to pass.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,7 +11,7 @@ PUBLIC_BASE_URL=http://localhost:8087
 
 # Comma separated list of tlds (e.g. ALLOWED_EMAIL_DOMAINS=example.com,example.org)
 ALLOWED_EMAIL_DOMAINS=example.org
-MIN_CUSTOM_DOMAIN_ALIAS_LENGTH=3
+MIN_CUSTOM_DOMAIN_ALIAS_LENGTH=2
 
 ADMIN_WEBSITE=https://www.thunderbird.net
 ADMIN_CONTACT=dummy@example.org
--- a/src/thunderbird_accounts/mail/tests/test_views.py
+++ b/src/thunderbird_accounts/mail/tests/test_views.py
@@ -105,7 +105,9 @@ class AddEmailAliasTestCase(TestCase):
                 self.assertEqual(
                     json.loads(response.content.decode()),
                     {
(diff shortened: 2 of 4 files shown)


kewisch:
Hi Nidhal, unfortunately we're not looking to change the minimum length to 2. You are welcome to use your custom domain where even 1 character is allowed.

kewisch:
That said I would take the hunks of your patch that adjust the error message to use the variable, if you are interested in updating let's reopen :)

nidhalsaidani:
image

same for custom domains
this is why i request update

kewisch:
Alright, that's valid. For custom domains we shouldn't be enforcing a minimum length. But I believe this needs to be fixed in a different way, so we can separate between custom domains and the built-in domains. Can you file an issue on this? If you have a pull request that allows custom domains to have any length to attach to it, that would be much appreciated!

PR #844, opened 2026-05-18
Fix E2E tests for allow list in contact form URL matching

davinotdavid:
The URL matching in the E2E tests was wrong as we insert a locale prefix when redirecting!


aatchison (approved):
thanks!

PR #851, opened 2026-05-19
docs: improve README.md for new devs

markstos:
## What changed?

- Adds DB migration command necessary to make first login work.
- Documents adding keyclock to /etc/hosts
- Documents that hot-reloading generally works-- full rebuilds are rarely needed.
- Documents setting up a subscription plan-- needed for full admin access
  to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Add the following to `/etc/hosts`:
 
 ```text
 127.0.0.1 keycloak
+127.0.0.1 stalwart
 ```
 
 ## Getting Started
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ During development it's great to have these handy!
 
 Make sure you have [uv](https://github.com/astral-sh/uv) up and running.
 
+Add the following to `/etc/hosts`:
+
+```text
+127.0.0.1 keycloak
+```
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,11 @@ During development it's great to have these handy!
 
 Make sure you have [uv](https://github.com/astral-sh/uv) up and running.
 
+Add the following to `/etc/hosts`:
+
+```text
+127.0.0.1 keycloak
```
MelissaAutumn, line 46:
```suggestion
127.0.0.1 keycloak
127.0.0.1 stalwart
```

Apparentally this was removed [in a commit](https://github.com/thunderbird/thunderbird-accounts/commit/592208aba42635bb752a05a4683af3c4fad7264b#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L41-L46) a while ago. Must have missed that in review. Thanks for adding it back!

README.md
```diff
@@ -52,8 +57,15 @@ uv run bootstrap.py
 
 This will create a virtual environment if needed and sync the latest project dependencies to your local environment.
 
-The project comes with some optional dependencies such as cli tools, tools for building the docs, and tools for working
-with our Subscription app powered by Paddle.
+After the Docker containers are running for the first time, bootstrap the Django database schema:
+
+```bash
+docker compose exec accounts ./manage.py migrate
```
MelissaAutumn: The container auto-migrates on boot so this shouldn't be required. Did you run into an issue with the auto-migrations?

MelissaAutumn (changes requested):
Looks good, just a small tweak and a question about migrations.

README.md
markstos: Yes, on first boot there was an error about Django not being able to access a session table and IIRC, this seemed to fix it. We could drop this for now and if it's really a problem, the next new new dev will help us surface it.

MelissaAutumn: Not ideal, I'll cut a ticket to test this out on a fresh install later but we can nix this part.

markstos: I've pushed a fixup commit to remove this. Ready for re-review and squash-merge.

MelissaAutumn (approved):
Perfect, thank you!

PR #852, opened 2026-05-19
chore(dev): Ensure .editorconfig gets loaded by VS Code / Cursor

markstos:
These editors appear to need to plugin to load the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
-.vscode
+.vscode/*
+!.vscode/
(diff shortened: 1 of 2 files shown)


MelissaAutumn:
I don't use vscode, so maybe @davinotdavid or @rwood-moz might have a better time testing this out?

davinotdavid (approved):
Works for me, thanks! Adding the extension does seem to apply the rules in the `.editorconfig` file.

Image

Issue #372, opened 2025-11-03
Fetch Plan / Subscription info and store it in global state / Pinia


davinotdavid:
We use plan / subscription data in a lot of different places both in Accounts and in Mail / Thundermail.

It would be super useful to introduce a global state solution / Pinia so that we can pass this data seamlessly instead of using the `window._page` in the Django template.

MelissaAutumn:
As we're starting to clean up the UI and make it more reactive this is becoming more of a priority.

(cc @malini )

Issue #347, opened 2025-10-28
Add `@subscription_required` annotation


MelissaAutumn:
Currently we use django's `@login_required` annotation for routes. We need to add a new one that checks for subscription status, so users who have an account but aren't subscribed yet can't do things like create app passwords. (That would fail anyways but it's just an example.)

MelissaAutumn:
Here's a few sentry issues that will be "fixed" with this annotation:
https://thunderbird.sentry.io/issues/7596637150/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream
https://thunderbird.sentry.io/issues/7596637132/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream
https://thunderbird.sentry.io/issues/7596663468/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

Issue #152, opened 2025-05-28
Implement model factories for tests


MelissaAutumn:
For appointment I simply wrote a wrapper function that when invoked creates an instance of the model either via faker data or by data you enter in the arguments. 

Ex/
https://github.com/thunderbird/appointment/blob/main/backend/test/factory/slot_factory.py
https://github.com/thunderbird/appointment/blob/main/backend/test/factory/calendar_factory.py
https://github.com/thunderbird/appointment/blob/main/backend/test/defines.py#L32

However there's also third party libraries for django out there that do the same thing but probably better: https://factoryboy.readthedocs.io/en/stable/index.html

We should probably investigate any third party libraries first, and fallback to our own method if they don't seem particularly great.

PR #854, opened 2026-05-19
Change App Password / Display name to API routes (Fixes #359)

### LINKED ISSUE #359: Change App Password / Display name to API routes
Instead of form submission we should switch it to an api call.

markstos:
Before, these were form submissions, causing full-page reloads.

An additional check was added to app-password changes to make sure the
user has a current subscription.

## QA Log

 - Added automated test coverage
 - Manually tested before / after behavior for both cases in a browser 

## Review notes

- Is there a way to further simplify the test coverage that's still sufficient?

### DIFF
--- a/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/UserInfoSide.vue
+++ b/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/UserInfoSide.vue
@@ -8,6 +8,7 @@ import {
   TextInput,
   LinkButton,
 } from '@thunderbirdops/services-ui';
+import { setDisplayName } from '../api';
 
 const { t } = useI18n();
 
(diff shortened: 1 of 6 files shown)


src/thunderbird_accounts/mail/views.py
```diff
@@ -41,6 +41,12 @@
 def app_password_set(request: HttpRequest):
     """Sets an app password for a remote Stalwart account"""
 
+    if not request.user.has_active_subscription:
+        return JsonResponse(
+            {'success': False, 'error': str(_('An active subscription is required to set an app password.'))},
+            status=403,
+        )
+
```
MelissaAutumn, line 49: Thanks, this is technically will be covered in https://github.com/thunderbird/thunderbird-accounts/issues/347 but good to have now.

assets/app/vue/views/MailView/sections/EmailSettingsSection/components/UserInfoSide.vue
```diff
@@ -27,26 +28,15 @@ const onSetDisplayNameSubmit = async () => {
...
-        'display-name': displayName.value,
-      }),
-    });
-
-    const data = await response.json();
+    const data = await setDisplayName(displayName.value);
 
     if (data.success) {
+      userDisplayName.value = displayName.value;
+      window._page.userDisplayName = displayName.value;
```
MelissaAutumn, line 35: Ideally we'd keep the window._page immuatable but I guess we cannot. I'll see if I can move https://github.com/thunderbird/thunderbird-accounts/issues/372 up.

MelissaAutumn: This doesn't update the name at the top of the page, but again a fix for this will be punted to #372.

MelissaAutumn (approved):
Looks good, thank you!

As we move towards all of the tickets we punted into the "next milestone" I'm expecting a lot of nice clean ups, but things might be a little broken in the mean-time. 

As for testing, that's basically how we do it. I think we'll gain a lot from #152 which is yet another "nice to have" we punted from early bird.

PR #855, opened 2026-05-19
Prevent contact form's file input from opening twice. (Fixes #849)

### LINKED ISSUE #849: TB Pro Accounts Support Form Asks to Attach Twice
## Describe the bug
Attempting to attach a .png file to the TB Pro Accounts support form results in the form not attaching, but attempting to attach it a second time works.

## To Reproduce

1. Go to https://accounts.tb.pro/contact
2. Click on "Choose files"
3. Select a .png file and confirm
4. See the dialogue box just disappear and take you back to step 3
5. Try to select the same .png file and it will now work

(Please note - I am not able to reproduce this every time, but it does happen often when testing)

## Expected behavior
Selecting a file and attaching it the first time

## Actual behavior
Randomly not attaching the file the first time, requiring you to select it again

## System

- OS: User reported on Windows 10, I was able to reproduce on MacOS Sequoia 15.7.1
- Browser Version: User reports Chrome 148 and Edge 148. I replicated in Chrome 148 as well. Couldn't replicate in Firefox.

## Additional context
Ticket 5736

MelissaAutumn:
Fixes #849 

We wrap the fileInput with a div that has an on-click event. The on-click event (as seen in changes) calls the fileInput's click event. You can see why some browsers may get confused :smile:. 

So Chrome-based browsers seem to prioritize the least deep element's events while Firefox prioritizes the deepest element's events. Interesting distinction there.

### DIFF
--- a/assets/app/vue/views/ContactView/components/ContactSupportForm.vue
+++ b/assets/app/vue/views/ContactView/components/ContactSupportForm.vue
@@ -62,7 +62,11 @@ const nativeAttrsForField = (field: TicketField) => {
   }
 };
 
-const triggerFileSelect = () => {
+const triggerFileSelect = (event: PointerEvent) => {
+  // Do not target the fileInput if they've clicked on the fileInput.
+  if (event.target === fileInput.value) {
+    return;
+  }
   fileInput.value?.click();
 };
 


markstos (commented):
The logic makes sense to me. This seems good to merge after manual re-testing in Firefox and Chrome.

Issue #628, opened 2026-03-06
Stalwart v0.16 API Changes


MelissaAutumn:
Stalwart v0.16 brings a bunch of changes. Some of those changes like the way user data is stored, and the management api affect us quite a bit. The new management/admin api is JMAP based and is not compatiable with how we wrote our mail client in tb-accounts. The changes to how user data is stored is also a challenge as we have to migrate that data to the newer format.

Thankfully Mauro has written a [migration proxy ](https://stalw.art/docs/migration/proxy/) and the jmap-based api changes seem like a step in the right direction from my brief skim. So here's some initial development tasks to be done before we can support v0.16.

## Admin api changes
See #1188 

## Local configuration
See #1168

## Readying for proxy
* [x] Investigate if we need to handle v0.15 and v0.16 admin api calls depending on user's migration status
* [ ] (If so) Add functionality to our mail client to route calls based on the [mappings](https://stalw.art/docs/migration/proxy/mappings/)

ryanjjung:
🫤

MelissaAutumn:
Looking into this now, and the new web ui is pretty slick. 

Unfortunately migration may be a pain, as it seems like we'll have to configure it via the cli tool. The new config.json is only used to store the internal datastore and its location inside the container. 

For local dev I'm thinking an entrypoint bash script that checks if the server has been provisioned, and if not then run a variety of cli commands.

MelissaAutumn:
Also unfortunately I'm starting with a blank slate on Stalwart's end, so I'll need to code the api changes before I can see if everything is working nicely.

ryanjjung:
I need to follow up on what I've been calling the "migration proxy" (I don't know what the official name will be when they release it), and that may help alleviate the migration a lot. The way it's been described to me is a service that sits in front of any Stalwart installations you have. It migrates accounts cross-schema over time while also routing calls to Stalwart to the appropriate installation to handle the account. In theory, this allows us to do migrations live with no downtime. In our case, we would have to swap DNS over to the migration proxy, wherever we install that, prior to beginning the migration. Lots of development work and some R&D about the proxy still to do on the infra side.

MelissaAutumn:
[Migration proxy was released](https://stalw.art/docs/migration/proxy/)! Seems to handle migration as mentioned on the fly with and stores a mapping in either a file, redis (preferred if we need to access that info), or an sql database. 

[...]

PR #863, opened 2026-05-22
Force sentry to capture domain not found errors during the delete process. (#690)

MelissaAutumn:
We need more information on what is failing for #690. This is really the only step I can see that can fail but still delete the Stalwart reference model. 

I don't want to spend a lot of time here as this will be re-written for #628

### DIFF
--- a/src/thunderbird_accounts/mail/views.py
+++ b/src/thunderbird_accounts/mail/views.py
@@ -288,10 +288,12 @@ def remove_custom_domain(request: HttpRequest):
             if _domain.stalwart_id:
                 try:
                     stalwart_client.delete_domain(domain.name)
-                except DomainNotFoundError:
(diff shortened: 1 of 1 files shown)


markstos (approved):
This is good, but could be further improved by notifying Sentry of all the possible domain removal failure options and including some more context about the specific domain for streamlined debugging. I've submitted a commit which does that for counter-review.

## QA Log

- Reviewed diff
- Reviewed other domain failure options
- Reviewed Sentry API docs for Python. https://getsentry.github.io/sentry-python/api.html#sentry_sdk.api.capture_exception

MelissaAutumn:
Changes may be unnecessary since we'll see different exceptions in different areas (with the exception of the one we weren't capturing lol) but they're appreciated. I think I handled Appointment's caldav issues like this.

Issue #771, opened 2026-04-30
Add E2E test for paddle payment flow


rwood-moz:
Add an E2E test for the paddle payment flow. We could run this in CI on PRs because we build a new clean stack each time there so we will have a new unsubscribed user each time.

If we want to run a paddle subscription E2E test on stage we would need a way to delete existing paddle subscription info for an existing test user (via some kind of automation script) so that the same user account can be used with the test over and over. Either that or we would need a way to automatically create a new TB Pro user on stage every time the test runs but I'm not sure that's ideal as we would end up with a large number of test accounts used one-time only. We aren't able to mock the paddle calls either (see #180). 

Regardless at the very least we could run a paddle subscription sign-up test in CI against PRs because they generate a clean dev stack each time. This issue can start with that, then we can look into solutions for running on stage in #180.

MelissaAutumn:
@rwood-moz are you in the slack channel with the Paddle folks? If not ask @Sancus for access, they might be able to help you with e2e testing.

PR #866, opened 2026-05-25
Add E2E tests for main accounts dashboard

### LINKED ISSUE #313: Add E2E tests for the main TB Accounts dashboard

rwood-moz:
## What changed?

Add E2E tests to validate the main TB Accounts dashboard.

## Why?

To expand the current E2E test suite.

## Limitations and Notes

This test will skip if it is running on local dev stack because when running on a local stack (i.e. on PRs in CI) there won't be a TB Pro subscription setup. In the future (#771) we will add a subscribe step when running the tests on a new local dev stack, and after that this test can also be enabled to run there. In the mean time this new test will run on stage after main merges, and on prod in the nightly test suite (on mobile and desktop).

My first patch primarily developed by Codex.

## Applicable Issues

Fixes #313.

## Screenshots

BrowserStack links running the new dashboard tests:

- [Desktop Firefox](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Firefox+Desktop/151?tab=sessions)
- [Desktop Chromium](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Chromium+Desktop/2?tab=sessions&testListView=spec)
- [Desktop Safari](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Safari+Desktop/1?tab=sessions)
- [Android Chrome](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Android+Chrome/20?tab=sessions)


test/e2e/pages/dashboard-page.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  async verifyManageSubscriptionOpensPortal() {
+    const [popup] = await Promise.all([
+      this.page.waitForEvent('popup'),
+      this.manageSubscriptionButton.click(),
+    ]);
+
+    await expect.poll(async () => popup.url()).not.toBe('about:blank');
+    expect(new URL(popup.url()).protocol).toMatch(/^https?:$/);
```
davinotdavid, line 130: Should we harden this assertion a little bit? I believe this would match a redirect to any HTTPS URL instead of the customer's paddle.

davinotdavid (commented):
Nice! Confirmed that all tests pass while pointing to stage and using a stage account that has a subscription:

Image

One question though:

test/e2e/pages/dashboard-page.ts
rwood-moz: Good catch, thank you, updated to verify that the URL is a paddle URL (IMO that is sufficient).

test/e2e/pages/dashboard-page.ts
```diff
@@ -0,0 +1,178 @@
...
+    expect(new URL(popup.url()).host).toBe(PADDLE_HOST);
```
davinotdavid: Perhaps we should do `toContain` here instead or change the `PADDLE_HOST` to be environment aware as stage's URL I believe is `sandbox-customer-portal.paddle.com` instead

```suggestion
    expect(new URL(popup.url()).host).toContain(PADDLE_HOST);
```

test/e2e/const/constants.ts
```diff
@@ -33,6 +40,7 @@ export const SMTP_PORT = Number(process.env.SMTP_PORT);
 export const IMAP_TLS = String(process.env.IMAP_TLS) ?? 'SSL/TLS';
 export const JMAP_TLS = String(process.env.JMAP_TLS) ?? 'SSL/TLS';
 export const SMTP_TLS = String(process.env.SMTP_TLS) ?? 'SSL/TLS';
+export const PADDLE_HOST = 'customer-portal.paddle.com';
```
davinotdavid, line 43: If we decide to do the environment specific:

```suggestion
export const PADDLE_HOST = String(process.env.PADDLE_HOST);
```

and then update the .env vars accordingly in `test/e2e/.env.stage.example` and `test/e2e/.env.prod.example`

davinotdavid (changes requested):
Thanks for the update! Sorry now the tests no longer seem to pass for me pointing to stage due to the URL not matching though :(

test/e2e/pages/dashboard-page.ts
rwood-moz: Ahh yes I tested on prod not stage, good point. Updated to use `toContain` so we don't need to use more env (and GHA secrets). Thanks :)

davinotdavid (approved):
lgtm! thanks!

rwood-moz:
> lgtm! thanks!

Thank you for your thorough review, much appreciated!

PR #868, opened 2026-05-25
Add trace and profile envs to prod (celery) and stage (accounts and celery) (Fixes #867)

### LINKED ISSUE #867: Enable sentry on stage and prod
Sentry is technically only enabled on accounts prod, not prod celery or any part of stage.

MelissaAutumn:
Fixes #867 

We must have missed shoving these two envs into stage/prod pulumi. I thought sentry too quiet on stage, and this should give us celery task errors as well.


MelissaAutumn:
By default these envs are set to 0.0, so it won't send data to sentry unless we explicitly ask sentry to capture something (or log.error.) 

So it's been somewhat working.

aatchison (approved):
Oops! Looks simple enough, approved.

PR #872, opened 2026-05-26
Raise anonymous rate limit to work around Stalwart bug

Sancus:
Until 0.16, the anon rate limit is applied to all JMAP requests no matter if authenticated or not.

Lets just use this workaround for now.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -24,6 +24,9 @@ protocol = "http"
 bind = ["[::]:8080"]
 protocol = "http"
 
+[http.rate-limit]
+anonymous = "10000/1m"
+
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):

MelissaAutumn (approved):
...i knew there was a rate limit occuring. We should maybe adjust that on stage/prod as well eh.

Sancus:
> ...i knew there was a rate limit occuring. We should maybe adjust that on stage/prod as well eh.

Yeahhh.

This can be modified with the api and maybe in the web admin so I might just bump it up.

Issue #507, opened 2026-01-16
Custom Domain DNS record improvements


kewisch:
Collection of various issues we've identified over the past few days:

* We recommend a few CNAME records pointing to mail.thundermail.com: mail.example.com, autoconfig.example.com, autodiscover.example.com, mta-sts.example.com. These won't really work out due to server certificates, and we aren't provisioning or allowing the user to store certificates. 
* All NAME/HOST have a period at the end, though some are missing (e.g. CNAME mail.example.com). I vaguely recall the dot is intentional, but we should double check and be consistent
* We should add a tip that for SRV records, some DNS providers require the priority to be included at the start of the VALUE/DATA.
* TXT SPF entry points to spf.mail.thundermail.com, where it should be spf.thundermail.com
* We're recommending a p=none rule for DMARC. A better rule would be `v=DMARC1; p=reject; rua=mailto:postmaster@example.com; ruf=mailto:postmaster@example.com`, but maybe we should tell them to do p=none first, if that works out then switch to the p=reject rule. 
* I believe `SRV _imap._tcp.says.wtf. 1 443 mail.thundermail.com` should actually be port 143, but see below. 
* We should consider making non-tls SRV records as `0 0 .`, e.g. for `_submission._tcp.example.com`. Others: _imap, _pop3, _carddav, _caldav. I don't know if this is strictly necessary though, each rule is more hassle.
* Since we don't support pop3 we should make ` 0 0 .` records for _pop3s and maybe _pop3.
* There is no dash in the last TXT line for the domainkey in the priority column.

We could split up the rules in absolutely necessary, and optional. I'm guessing caldav/carddav are not needed, and I'm not even sure all the _tcp rules are really necessary? The minimum set would be those needed to make Thunderbird (without the Pro add-on) able to find the right settings.

Let's get as many of these knocked out as easily possible for the community rollout, I'm hoping most of this is just changing the text.

@Sancus any additional thoughts?

Sancus:
For JMAP, caldav, carddav, and any other web-based API, they are an http "redirect", and those won't work unless we allow Stalwart to pull lets encrypt certs for every domain and just rely on those, which I guess we could. Otherwise we'd have to implement our own mechanism.

I'm not sure it's a good idea to recommend people CNAME the mail server without any understanding of the above, and it's not like it provides any functionality benefit.

[...]

PR #879, opened 2026-05-27
Custom domain DNS records reconciliation

### LINKED ISSUE #857: Indicate to the user that they will need to delete/update certain keys instead of just adding ours
We've landed on a more condensed set of DNS entries to not overwhelm the user (Ref also #521). However, this also means when the user moves over from e.g. Outlook, they may have a few entries that are detrimental.

From https://tbpro.zendesk.com/agent/tickets/5711 for example: 

✗ FAIL autodiscover CNAME → still points to Office 365 (blocks Thunderbird account setup)
✗ FAIL _autodiscover._tcp SRV → stale record present: _autodiscover._tcp.example.com has SRV record 0 0 443 cpanelemaildiscovery.cpanel.net.

We need to:
* Go through all the optional records, make sure they either don't exist or point to our servers
* Go through any additional records we don't need (like autodiscover?) and tell the users to delete them
* Verify that the required records exist (as we do?)

This may need some additional design, though I wanted to be clear on the requirements before we request it.

gz#5711

davinotdavid:
## What changed?

- When adding a Custom Domain, we were doing a request to the backend to get the DKIM records + merging it with a frontend list of static records. This static list has now been moved to the backend as this will also be needed for DNS conflict lookup.
- As part of the `Verify Domain` step, we are also now checking the custom domains' current DNS records to see if they match with the our expected DNS records:
  - If there is a MX record at the same priority, that's an error.
  - If SPF record doesn't contain include:spf.thundermail.com, that's an error.
  - If there's an `autodiscover.{cust_domain}` record, that's an error
  - If there's an `SRV _autodiscover.tp.{cust_domain` record, that's an error.
- Errors are displayed inline in the DNS table to better connect the error with the related error.
- Clarified error messages, like using "MX record" consistently instead of "mail server". 
- Fixed bugs related to whether records contained a trailing dot or not
- Added new CUSTOM_DOMAINS_DO_VERIFY boolean for .env to enable local verification testing.
- Added subdomain support for customer domains. Ex: `foo.mydomain.com`. It's valid to have separate mail settings  for a subdomain and gives developers useful flexibility for live-testing on a personal subdomain without impacting mail settings for the top-level domain.
- Various refactors for code organization, including merging `verify_domain` and `get_expected_dns_records_with_status`. Also, to clarify where we are referring to a customers domain vs our domains in various places, now we use `cust_domain` instead of `domain_name` in several places. 

## Why?

Configuring a custom domain requires attention to detail and may incur in problems that require lots of back and forth with Support to resolve. Having a way to highlight the problems to users proactively may help them accurately do their setup and be less frustrated.

## Limitations and Notes

We could be verifying the domain through a task but that might require some extra discussions on how / where / when to show the verification information to the user so that's out of scope for now.

## How to test

[Prerequisites]
- Have a domain with arbitrary MX, SPF, CNAME records

[Steps]
- In Thundermail Dashboard, add the domain from the prerequisite step as a new Custom Domain
- Click "Verify Domain" and check that the conflicting rows get highlighted like the screenshots below showcasing the current value

## QA Log

- [x] Complete `check_stale_dns_records` method  (src/thunderbird_accounts/mail/utils.py) as `autodiscover` records are not being fetched / compared correctly with live values
- [x] Refactor / Remove the `verify_domain` in favor of `get_expected_dns_records_with_status` as they do very similar things (`verify_domain` is now a subset of checks)
- [x] Move the domain verification into a Celery task
- [x] Test case when MX validation passes, but there are SPF and DKIM warnings.
- [x] Test that we should FAIL autodiscover CNAME → still points to Office 365 (blocks Thunderbird account setup) ex: `autodiscover.{cust_domain}` is a CNAME to `autodiscover.outlook.com`
- [x] Test we should FAIL  if there's a SRV record `_autodiscover._tcp.sub.{cust_domain}`.
- [x] Consider it an error we are the same MX priority level as another provider
- [x] SPF check should be pass as long as `include:spf.thundermail.com` is included.
- [x] Code clean-up.
- [ ] (Design needed) Show the DNS records to be removed somewhere in the UI _For now, they are displayed in-line, but the overall experience could possibly be improved. Not clear if the design needs to be perfect for Flight 2_

Manual testing process: Two categories of code changed help support manual testing. The first was the addition of the `CUSTOM_DOMAINS_DO_VERIFY` to enable live DNS verification locally.  The second was patches to support improved subdomain support. Once those were in place, I could safely test various correct and broken settings on a subdomain of a personal domain where I control the DNS without impacting the actual mail delivery of the parent domain. 

## Applicable Issues

Related with https://github.com/thunderbird/thunderbird-accounts/issues/507 and https://github.com/thunderbird/thunderbird-accounts/pull/369

Closes https://github.com/thunderbird/thunderbird-accounts/issues/857

## Screenshots

image

image

image


markstos:
I've now rebased this off the related DKIM PR I was working on, resolved the conflicts, and updated the `base` of this PR to reflect that.

markstos:
Progress report today:

* We discussed "Move the domain verification into a Celery task", but after some testing, the verification should be able to complete in about a second for common cases. At least for now I think can be a web server task. Subtracting that bit from the scope for now. 
* Rebased off the most recent other DKIM/CNAME work and resolved conflicts.
* Added new CUSTOM_DOMAINS_DO_VERIFY boolean for .env to enable  local verification testing.
* Fixed bug with SPF record value missing trailing "."
* Fixed bug with MX record generating not working when provided domain was a subdomain. 
* Completed refactor to merge `verify_domain` and `get_expected_dns_records_with_status`.

What's left to do is final testing of the autodiscovery stuff and any other design updates we want to do.

markstos:
@MelissaAutumn @davinotdavid I believe this is ready for review, with all the key cases tested with automatic and/or manual testing. There's an open question of how perfect the design should be for Flight 2 to take off.

kewisch:
Excellent! It doesn't have to be perfect, but the closer you get the less you need to touch this code again. If there is follow up work let's schedule that next.

Do you have have a screenshot of the case where someone added an extra autodiscovery domain a la outlook that doesn't point to us?

MelissaAutumn:
If it's ready for review can you remove the [WIP] from the title? I don't generally look at wip or draft PRs so this evaded my radar.

src/thunderbird_accounts/mail/views.py
```diff
@@ -193,11 +210,21 @@ def get_dns_records(request: HttpRequest):
         # Backfill only missing DKIM selectors before returning DNS records.
         stalwart_client.ensure_dkim(domain.name)
         mail_tasks.publish_hosted_dkim_dns_records.delay(domain.name)
-        dns_records = stalwart_client.get_dns_records(domain.name)
+        dns_check = stalwart_client.check_domain_dns(domain.name)
+        stale_dns_records = check_stale_dns_records(domain.name)
```
davinotdavid: I think that we shouldn't be performing the DNS checks upon fetching & generating the dkim records since, when adding a new custom domain, the custom domain won't have the records yet and we would be showing error results before verification.

src/thunderbird_accounts/mail/views.py
```diff
@@ -193,11 +210,21 @@ def get_dns_records(request: HttpRequest):
         # Backfill only missing DKIM selectors before returning DNS records.
         stalwart_client.ensure_dkim(domain.name)
         mail_tasks.publish_hosted_dkim_dns_records.delay(domain.name)
```
davinotdavid: I don't have a lot of context / knowledge on the DKIM rotation task but during my local testing, it doesn't seem that we are creating / returning the DKIM record in the list. Are we not returning / showing it in the list at all anymore?

Image

If we are handling the DKIM setting automagically somehow, I think it is confusing to show that the DKIM record has not being detected yet there's no way for the user to recover / retry.

src/thunderbird_accounts/mail/utils.py
```diff
@@ -1,20 +1,376 @@
...
+            continue
+
+        live_values = [rdata.to_text() for rdata in answers]
+        if live_values:
+            address_records[rdtype] = _unique_dns_values(live_values)
+
+    return address_records
+
+
+def check_stale_dns_records(cust_domain: str) -> list[dict]:
```
davinotdavid, line 317: I've attempted to add some autodiscover related records but they don't seem to be detected by this method. It is possible that I am misconfiguring something in Cloudflare though!

davinotdavid (commented):
Nice! Thank you for progressing in the PR! Really appreciate you taking over this 🙇 

I have a few functional comments for now. I've done a video recording of my testing which highlights two main issues (in my opinion):

1. We are doing the checking of the DNS upon viewing the DNS records (which is quite slow and we don't provide feedback on what's happening), while only showing critical errors. I think we shouldn't do this upon _viewing_ the DNS records as a brand new custom domain wouldn't have them configured anyways. The verification of the domain highlights the rest of the warnings.


2. DKIM records are not being displayed and, if it is automagically configured somehow, there's no way for the user to retry / recover in case it is wrong.

https://github.com/user-attachments/assets/231df6fc-6904-41c2-987d-2538f55f68d8

malini:
Please update the design to detail which records are conflicting, and then we can ship 🚀

src/thunderbird_accounts/mail/views.py
davinotdavid: ~~Re-tested this morning and the DKIM records are not showing up in the list for me. Is there a specific setup needed to make it happen?~~

Works for me, I was missing env vars (`HOSTED_DKIM_DOMAIN` and such)! 

image

src/thunderbird_accounts/mail/utils.py
davinotdavid: Nice! Re-tested this and the autodiscover CNAMEs are showing up for me now, according to the QA Log, thanks!

image

src/thunderbird_accounts/mail/views.py
davinotdavid: This has been solved in my latest testing, thanks!

davinotdavid (commented):
LGTM! 🚀 All comments / issues that I had were resolved and I believe we can do another set of testing in staging to verify performance on the DNS resolution (my local here was proven to be terrible 😅).

Thank you @markstos for getting this into the finish line!

Issue #839, opened 2026-05-15
Expose RSA DKIM keys in custom domain setup


rtanglao:
## Describe the bug
The RSA DKIM keys are not exposed in custom domain setup. We have at least one person with this issue (see Zendesk Ticket [5655](https://tbpro.zendesk.com/agent/tickets/5655))

## To Reproduce

1. Go to acouunts.tb.pro
2.  Scroll down to 'Custom Domains'
4. Click "View DNS Records"
## Expected behavior
* The DNS record for DKIM RSA is displayed.

## Actual behavior
* DKIM RSA DNS record is not displayed.

## System

- OS: 
- Browser Version 

## Additional context

gz#5528

gz#5655

gz#5679

MelissaAutumn:
We only create as defined [here](https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/settings.py#L400) and called [here](https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/mail/clients.py#L225).

We should make `STALWART_DKIM_ALGO` a list, and turn create_dkim into a loop that creates a key for all algo's in that list.

markstos:
To add context about why this is important: In RFC 6376 that defines DKIM, it says that 'Signers and Verifiers MUST support the "rsa" key type.'

Ed25519 is an emerging alternative, but not yet universally supported. To deal with this, we can support both RSA because we must and ed25519 because it's forward looking.

So I come to the same conclusion as @MelissaAutumn: Turn the single algorithm we support into a list and support multiple key  types.  We can two multiple keys and use different selectors for each.

MelissaAutumn:
We should also add in an admin action so we can backfill the custom domains without an rsa key. and thanks for the background info @markstos !

rtanglao:
* see also #848

markstos:
Resolved by linked issue.

PR #881, opened 2026-05-28
Publish DKIM DNS records to our own domains and share DNS CNAMEs to customers

### LINKED ISSUE #848: third-party DKIM DNS entries should be CNAMEs to support future key rotation, key type changes.
## Describe the bug

Currently we direct users to set up TXT records for DKIM keys like this:

```
TXT 202605e._domainkey.stosberg.com.  v=DKIM1; k=ed25519; h=sha256; p=shh7i1mEG2TeA/xwPKeSmDxqyBNC0Q+aDI2GOy8wmLw=
```

The problem with that if we want to later rotate the key material or change the parameters like key type or hashing, we have to contact every customer with custom DNS to do so.

## Recommended Fix

The recommended fix is instead direct customers to two or three CNAME records to their DNS, pointing to DNS entries that control, like this:

```
CNAME tm1._domainkey.stosberg.com tm1.stosberg.com.dkim.thundermail.com
CNAME tm2._domainkey.stosberg.com tm2.stosberg.com.dkim.thundermail.com
CNAME tm3._domainkey.stosberg.com tm3.stosberg.com.dkim.thundermail.com
```

Other major email provides also use CNAME delegation for DKIM records for the same reasons.

We may want to use a separate domain to host these DNS records than our main domain. 

After making this change, we can rotate these DKIM keys completely on our end, without having to contact every customer with a custom domain about updating their records.

## Related Issues and Pull Requests

*  #507 #842

markstos:
Here we improve DKIM handling.

* First, we start to publish RSA key types in addition to Ed25519. The spec requires that we support RSA, while Ed25519 is forward-looking. This incorporates #842 
* Instead of advising users to publish TXT records with hardcoded DKIM keys, instead we publish
  the keys with names like `tm{1,2,3}.CUST-DOMAIN.dkim.thunderhosted.com` advise them to host
  CNAMEs to our canonical records. This best practice allows us to later rotate the keys without
  involving customers.
* User's "custom domains" admin area is updated to reflect CNAME guidance vs TXT.

Fixes #848 #839

## Out of scope

* Handling DKIM key rotation
* Updating customer DNS directly

Separate issues can be opened to pursue these updates.

## Deploy notes

In production the following envs need to be set:

```
HOSTED_DKIM_CLOUDFLARE_ENABLED=true
HOSTED_DKIM_CLOUDFLARE_API_TOKEN=
HOSTED_DKIM_CLOUDFLARE_ZONE_ID=
# Enable after Stalwart 0.16 upgrade
STALWART_DKIM_STAGE_MANAGEMENT_ENABLED=false
```

## Stalwart 0.15 vs 0.16 notes

Stalwart 0.15 has more basic support for DKIM keys. 0.16 introduces a JMAP admin API for managing these as well "stages" used for key rotation.

Some code is gated by:

```
# Enable after Stalwart 0.16 upgrade
STALWART_DKIM_STAGE_MANAGEMENT_ENABLED=false
```

For 0.15, we create active DKIM keys immediately because that's all we can do.

In the (currently inactive) 0.16 code path, we can be more nuanced: Create a DKIM key in the pending state, then after the domain is verified moved to the "active" state.

The difference matters to Stalwart, which will sign messages with an "active" key, but not a "pending" key.

For the simple case of a brand-new domain with no existing DKIM keys the distinction is immaterial, but for customers migrating existing domains with DKIM hosted elsewhere, this more careful distinction may help Stalwart to not start signing mail until we've verified the DNS is all set.

## QA Log

- Added automated testing
- Manually tested frontend (see screenshot)
- Checked that logging indicates that Cloudflare DNS updates would be correct.
- I see my editor is adding a newline at the end of some files that didn't have one, but
  this behavior is consistent with what's in `.editorconfig` so I assume it's actually desirable.

```
celery-1      | [2026-05-28 15:51:46,184: INFO/ForkPoolWorker-16] Task thunderbird_accounts.mail.tasks.publish_hosted_dkim_dns_records[467931b4-144a-4d1f-a86c-d1170f3be4c0] succeeded in 0.00786278499981563s: {'domain_name': 'test-flight.com', 'records': [{'type': 'TXT', 'name': 'tm2.test-flight.com.dkim.thunderhosted.com', 'content': 'v=DKIM1; k=ed25519; h=sha256; p=73lYcu11LC+PdjcpmdER/gebCzuSD+tOCGDGIK9UFoU='}, {'type': 'TXT', 'name': 'tm1.test-flight.com.dkim.thunderhosted.com', 'content': 'v=DKIM1; k=rsa; h=sha256; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAuPTbGmUnPBPNoJlzexkP+iJNYJle4zB2eE2YS155MfMPbXppIl0VsfxXELC8fPdmP+fZC5DGUEPzX0gFMUu+LpIjfQiHWFvfBSwRXbG1wX/1gfaBMZRspxsa8zUnc/XvqHjoJRV/HcbOVRCLslcL0nTPDEp5lQyzGaqEZbKp6mUUhF2r86YU2BX3OR13Mbyn0+EBLHL8G7tUJEDNVXIOW1+CYmtkmyMSn92Sqcavv4w5mD9l1oa19DI2srhe5T4wuDloFmMh1D8XOfOxdB7A0gPXQ5MxJq9Sq4hkkHMTraQyDyyNg93IhqdjEMUveG7Rf9L1NbjGUpTMI2BRpu0yFwIDAQAB'}], 'skipped': True, 'task_status': 'success'}

```

Note: Actually updating Cloudflare was not tested.

image

## References

 * https://developers.cloudflare.com/api/resources/dns/subresources/records/methods/create


Sancus:
Final commit message should follow the format of historical messages with a full sentence describing the (the title of this PR is good) and (Fixes #X, Fixes #Y) in brackets following that sentence. No types at the front. JFYI.The format of commit messages in the PR itself aren't super important if we squash, but are if we rebase.

I'll review this in more detail later today but on a quick glance it looks pretty good!

src/thunderbird_accounts/mail/clients.py
```diff
@@ -223,16 +223,23 @@ def get_domain(self, domain):
...
+            }
+            response = requests.post(
+                f'{self.api_url}/dkim',
+                json=data,
+                headers=self.authorized_headers,
+                verify=settings.VERIFY_PRIVATE_LINK_SSL,
+            )
+            response.raise_for_status()
+            response_data = response.json().get('data')
+        return response_data
```
MelissaAutumn, line 268: This will only ever return the last response_data. But also the dkim route returns null so it's like kinda pointless here. 

I'd maybe catch the raise for status and separately raise a custom stalwart exception FailedToCreateDKIM(algorithm, domain) or something like that. Therefore if we have a large try/except we can catch that explicitly.

src/thunderbird_accounts/mail/clients.py
```diff
@@ -495,6 +502,64 @@ def get_dns_records(self, domain_name: str) -> list[dict]:
...
+        response.raise_for_status()
+        return response.json()
+
+    def get_dkim_signatures(self, domain_name: str) -> list[dict]:
+        domain = self.get_domain(domain_name)
+        domain_id = domain.get('id')
+        if not domain_id:
+            raise RuntimeError(f'Stalwart domain {domain_name} did not include an id')
+
+        response = self.make_jmap_admin_call(
```
MelissaAutumn, line 626: Are you mixing up Stalwart v0.16 and v0.15 (which is what we're on.) I don't believe Stalwart has an admin jmap api on v0.15. 

Looking at stalwart's admin panel on v0.15 it calls 
```
GET
	http://stalwart:8080/api/settings/group?limit=10&suffix=algorithm&prefix=signature&page=1&=
```

src/thunderbird_accounts/mail/dkim.py
```diff
@@ -0,0 +1,191 @@
...
+def _normalize_domain(domain_name: str) -> str:
+    return domain_name.strip().rstrip('.').lower()
+
+
+def _normalize_txt_content(content: str | None) -> str:
+    if content is None:
+        return ''
+    content = content.strip()
+    if len(content) >= 2 and content[0] == '"' and content[-1] == '"':
+        return content[1:-1]
```
MelissaAutumn: You can use strip here to do everything: 

```python
return content.strip('" ') # Note the white space to remove trailing spaces
```

src/thunderbird_accounts/mail/dkim.py
```diff
@@ -0,0 +1,191 @@
...
+def dkim_signatures_to_dns_records(domain_name: str, signatures: list[dict]) -> list[dict[str, str]]:
+    domain_name = _normalize_domain(domain_name)
+    records = []
+
+    for signature in signatures:
+        selector = signature.get('selector')
+        if not selector:
+            continue
+
+        if signature.get('stage', 'active') != 'active':
```
MelissaAutumn: I'm not too familiar with dns keys / dkim stuff. In what situation would stage be empty? And why would the default value be to continue as normal?

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -4,15 +4,25 @@
...
+    build_hosted_dkim_txt_records,
+    publish_hosted_dkim_txt_records,
+)
 from thunderbird_accounts.mail.exceptions import DomainNotFoundError, AccountNotFoundError
 from thunderbird_accounts.mail.models import Account, Email
 from thunderbird_accounts.celery.exceptions import TaskFailed
 from thunderbird_accounts.core.types import TaskReturnStatus
 
 
+class HostedDkimPublishRetry(Exception):
```
MelissaAutumn: See exceptions.py for how we should format these. Ideally they should be keyed and store the required info. And that should in theory give sentry a good amount of context on the exception / bug.

MelissaAutumn (changes requested):
I haven't given it a go to test as I'm still working on things but here's some starter review points.

src/thunderbird_accounts/mail/clients.py
markstos: I've reworked things so work with current APIs and be future-compatible through a feature-flag as well.

src/thunderbird_accounts/mail/dkim.py
markstos: Thanks for the tip! A nice simplification.

src/thunderbird_accounts/mail/dkim.py
markstos: There maybe some confusion here between `selector` which had the `if not` clause and `signature`, which has stages.

The `if not` check for selector was a mistake-- Good catch. DKIM Selectors are required both Stalwart and the related RFC. We can rely on them to always be present.

Signatures can be in a number of stages depending on where they are in the key rotation process, including `active`, `pending`, `retiring` and `retired`.  This is not part the DKIM spec but is state that Stalwart tracks during key rotation. Although we plan to rotate keys ourselves, we will still use Stalwart as our single source of truth. I'll be refining this to publish initially in the `pending` state and then when the `Verify` step is completed, move them to the `active` state. The distinction matters to Stalwart, as it will it sign emails with the `active` key.

src/thunderbird_accounts/mail/clients.py
markstos: Feedback addressed in newer commits.

src/thunderbird_accounts/mail/tasks.py
markstos: Good tip. Rich exceptions are useful.  Updated.

markstos: Implemented.

src/thunderbird_accounts/mail/clients.py
MelissaAutumn: Maybe you've accidentally adjusted the pinned stalwart version in docker-compose to be either latest or v0.16? 

I'm not seeing the custom stalwart capabilities in 0.15.4 (https://github.com/stalwartlabs/stalwart/blob/v0.15.4/crates/jmap-proto/src/request/capability.rs#L57 vs https://github.com/stalwartlabs/stalwart/blob/68946d4f982f60a1fa54f335547be11879ed7fea/crates/jmap-proto/src/request/capability.rs#L92) (I believe jmap explicitly mentions for clients to ignore extensions they don't understand, I'm guessing this goes for servers too but I'd have to re-read lol.) 

There also doesn't seem to be any dkim paths on the jmap api that I could find in a brief search on 0.15.4. 

Just trying to clear up any confusion so we can avoid "it works on my machine" I'm currently looking at testing this on my machine though :smile:

kewisch:
I haven't checked the code so maybe you've done this, but is there a django action that would allow us to backfill the DKIM CNAME DNS entries for current domains? Or will we do that separately as a one-off migration?

When we deploy this to prod we'll also need to gather a list of existing verified custom domains so we can reach out to users asking them to update.

We should also plan some future work to inform a user when their domain is no longer verified. This would help with either the situation where they make changes after the fact, or for cases like this where we require them to change any keys.

src/thunderbird_accounts/mail/clients.py
MelissaAutumn: Just wanted to bump this

.env.example
```diff
@@ -91,6 +91,19 @@ STALWART_BASE_API_URL=http://stalwart:8080
...
+# Hosted DKIM DNS delegation. Customers publish CNAMEs under their own domains
+# to these hosted records so Thunderbird can rotate DKIM keys later.
+HOSTED_DKIM_DOMAIN=dkim.thunderhosted.com
+HOSTED_DKIM_SELECTORS=tm1,tm2,tm3
+# True for production, useful to support false for testing, enterprise cases.
+HOSTED_DKIM_CLOUDFLARE_ENABLED=false
+HOSTED_DKIM_CLOUDFLARE_API_TOKEN=
+HOSTED_DKIM_CLOUDFLARE_ZONE_ID=
+# One hour
+HOSTED_DKIM_CLOUDFLARE_TTL=3600
```
MelissaAutumn, line 106: You might want to copy some of these values over to .env.test which is used during test runs. Might be able to slim down the `@override_settings` since they'll be defaults.

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -154,6 +161,83 @@ def update_quota_on_stalwart_account(self, username: str, quota: Optional[int]):
...
+def publish_hosted_dkim_dns_records(self, domain_name: str):
+    phase = 'initialize'
+    dkim_dns_records = []
+    hosted_records = []
+    expected_record_count = len(
+        {selector for selector in settings.STALWART_DKIM_ALGO_SELECTORS.values() if selector}
+    )
+
+    try:
+        phase = 'fetch_stalwart_dkim_dns_records'
```
MelissaAutumn, line 181: Is phase just a debugging var?

MelissaAutumn:
I'll be pushing up some pydoc comments as it's a bit lacking and @markstos is laptop troubleshooting.

MelissaAutumn (approved):
Works locally, I don't have the cloudflare keys pulled down so I haven't tested that. We can give it a go on stage and if there's issues we can work from there. 

(Note to self: This will need a pulumi key update.)

MelissaAutumn:
I've pushed some changes to dkim.py that re-organize the file to move class up top, private after, and finally public functions. 

Also I've added some caching function decorators to some commonly used functions. They only cache off of fn name + arguments so I can't add them to functions where there's a `or settings.SOME_DEFAULT_SETTING` call. 

We don't have DI in our app (maybe it's a good investigation ticket) but Python is pretty flexible that we mostly monkeypatch api clients via the import. Avoids some extra parameters.

markstos:
@kewisch I've created https://github.com/thunderbird/thunderbird-accounts/issues/891 as a placeholder to make sure that we don't lose track of the backfilling task. It use extra assignee  and other metadata filled in.

.env.example
markstos: Good idea. I found a couple that made sense to handle that way.

src/thunderbird_accounts/mail/tasks.py
markstos: It's  used when exceptions are thrown to identify the specific phase the error was thrown in when a try/catch could have more than one cause.

markstos:
@MelissaAutumn I believe open feedback has been addressed.

MelissaAutumn:
> @MelissaAutumn I believe open feedback has been addressed.

Can you reply to the comment I unresolved regarding Stalwart versioning? Thanks!

src/thunderbird_accounts/mail/clients.py
markstos: I did make any local changes to the Stalwart version on my machine's compose file and checked that I didn't have a rogue other version of Stalwart running. Please test on your machine to see if it works for you or not.

MelissaAutumn: Moving this convo to matrix.

markstos:
@Sancus @ryanjjung my final commit adds the ENV vars added here to Pulumi. The way it's structured now, it expects Cloudflare API key/secrets to AWS Secrets. That could be restructured to use Pulumi's vault feature. Either way, I'm hoping @Sancus can supply those as I've been challenged access those DNS settings.  We need one pair with permission to add DNS records under `dkim.thundermail.com` and a second one to add DNS records under `dkim.stage-thundermail.com`.

Issue #164, opened 2025-06-03
Enable PostHog


malini:
Gather metrics around usage through PostHog for accounts as well as for keycloak

MelissaAutumn:
@malini @radishmouse 
Does this include Keycloak's login theme?

MelissaAutumn:
Correcto

kewisch:
As discussed on Matrix let's put this on hold until Thunderweek.

PR #882, opened 2026-05-28
Minor fixes and enable posthog secret (#164)

MelissaAutumn:
Additional fixes for #164, we were not capturing done state, and errors like request throttles or "this user is not in the allow list".

Finally the most erroriest of errors, we don't actually have posthog enabled on the web container. We do on the celery workers, but not web. So this should fix things up.

### DIFF
--- a/assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
+++ b/assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
@@ -8,7 +8,7 @@ export const enum SIGN_UP_STEPS {
   USERNAME = 10,
   PASSWORD = 20,
   VERIFY = 30,
-  DONE = 100, // Not currently used
+  DONE = 100,
 }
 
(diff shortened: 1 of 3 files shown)


assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
```diff
@@ -67,10 +67,17 @@ export const useSignUpFlowStore = defineStore('signUpFlow', () => {
   
     if (response.status === 200) {
       $reset(false);
+      // Capture the done step. 
+      // The user will never be in this step but it's a good indicator that they've succeeded.
+      captureStep(SIGN_UP_STEPS.DONE);
```
davinotdavid, line 72: Now that we are using the step, we can remove the comment here https://github.com/thunderbird/thunderbird-accounts/blob/main/assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts#L11

davinotdavid (approved):
lgtm!

PR #888, opened 2026-05-30
Fix blank spf host for local development

lb1192176991-lab:
## What

Skips the SPF DNS check when `expected_host` is not a fully-qualified domain name (e.g. `localhost` or `None` from missing `.env` config).

## Why

When `SMTP_HOST` is set to `localhost` (common for local development) or is `None` (unset in `.env`), the SPF include string becomes `include:spf.localhost` or `include:spf.None` — both invalid. This causes a blank SPF host and produces warnings/nonsense DNS lookups.

## Testing

- Verified manually: when `expected_host` is `localhost`, the SPF check is skipped with an info log
- When `expected_host` has a dot (e.g. `mail.test.com`), existing SPF behavior is unchanged
- Existing tests pass unchanged

### DIFF
--- a/src/thunderbird_accounts/mail/clients.py
+++ b/src/thunderbird_accounts/mail/clients.py
@@ -536,19 +536,24 @@ class MailClient:
 
         # 2. Check SPF Record
         try:
-            txt_answers = dns.resolver.resolve(domain_name, 'TXT')
-            has_spf = False
-            expected_spf_include = f'include:spf.{expected_host}'
+            # Skip SPF check when expected host is not a fully-qualified domain
+            # (e.g. localhost, or None from missing .env config)
+            if not expected_host or '.' not in expected_host:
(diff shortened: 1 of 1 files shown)


markstos:
Was there an additional problem encountered besides the display of the error message? Was the error message blocking development somehow?

lb1192176991-lab:
Good point. The error wasn't blocking production, but in local dev environments where expected_host is set to 'localhost' or left empty/None, the SPF resolver would throw a cryptic DNS error instead of a clear message. This fix makes the dev experience smoother by skipping the SPF check when the host isn't a real domain.

markstos:
At this time, we are more interesting in keeping the local behavior more closely aligned with production, so we aren't accepting this change. Clear error messages for devs are nice, so if there's a narrower patch to improve the error message for this edge case we could consider that.

PR #897, opened 2026-06-02
Allow importing discounts per person with CSV

### LINKED ISSUE #900: Support discount IDs when importing allow-list entries
Allow-list imports currently do not support assigning discount IDs per imported user. This makes bulk setup tedious and error-prone when different users need different discount IDs, because the entries have to be imported first and then updated manually.

The admin import flow should support CSV rows containing an email address and an optional discount ID. Discount IDs should remain optional so existing email-only imports continue to work.

The single-entry admin form should also validate discount IDs consistently so invalid values are caught before saving.

kewisch:
Fixes #900

## What changed?

In the allow list admin, allow importing users with a discount id per user by entering CSV with the email and discount id. Discount id is optional.

Also added some validation to the discount id on the single entry form.

## Why?

Doing this manually is very annoying and error prone

## Screenshots

image
image

### DIFF
--- a/src/thunderbird_accounts/authentication/models.py
+++ b/src/thunderbird_accounts/authentication/models.py
@@ -1,13 +1,18 @@
 from functools import cached_property
 
 from django.conf import settings
-from django.core.validators import MinLengthValidator
+from django.core.validators import MinLengthValidator, RegexValidator
 from django.db import models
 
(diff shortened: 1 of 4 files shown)


davinotdavid (approved):
LGTM! Ran the e2e tests locally just in case and they pass successfully.

Also manually tested locally with common success and error cases!

PR #899, opened 2026-06-02
Set Cloudflare secret data for DKIM functions

ryanjjung:
I created two different Cloudflare API tokens, one per environment, and set them in the Pulumi config with `pulumi config set --secret` commands. I ran a targeted `pulumi up` on the secrets manager to create the AWS Secrets Manager entries referenced in the secrets section. I updated the mapping of secrets we apply to most task definitions so they would reference these secrets. I also updated the old task defintion where we don't apply this.

### DIFF
--- a/pulumi/config.stage.yaml
+++ b/pulumi/config.stage.yaml
@@ -3,7 +3,7 @@
 ### Special variables used throughout this file
 
 # Update this value to update all containers based on this thunderbird/accounts image
-.accounts_image: &ACCOUNTS_IMAGE 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:dc8d8ea6e76d8c7f25b260e964bada9d38a38e4b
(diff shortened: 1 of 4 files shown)


pulumi/config.stage.yaml
```diff
@@ -3,7 +3,7 @@
 ### Special variables used throughout this file
 
 # Update this value to update all containers based on this thunderbird/accounts image
-.accounts_image: &ACCOUNTS_IMAGE 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:dc8d8ea6e76d8c7f25b260e964bada9d38a38e4b
+.accounts_image: &ACCOUNTS_IMAGE 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/accounts:59fe695cef7f94721b34ab37cd00338890f79d6a
```
ryanjjung, line 6: This is the most recent build in ECR.

markstos (approved):
## QA Log

- Chatted with Ryan about this. LGTM. The real test will be on staging.

PR #902, opened 2026-06-02
Remove duplicate variables in stage

ryanjjung:
This removes duplicated environment variables from a stage task defintion.


markstos (approved):
## QA Log

Confirmed this is correct. 

  The dupes are not visible as block-style name: FOO entries because the
  remaining definitions are YAML anchors using inline maps:

  - config.stage.yaml:95 defines
    &SECRET_HOSTED_DKIM_CLOUDFLARE_API_TOKEN as {name:
    "HOSTED_DKIM_CLOUDFLARE_API_TOKEN", valueFrom: "arn:..."}

  - config.stage.yaml:96 defines
    &SECRET_HOSTED_DKIM_CLOUDFLARE_ZONE_ID the same way

PR #913, opened 2026-06-03
Check aliases on the debounced username check during sign-up. (Fixes #912)

### LINKED ISSUE #912: Double check sign up flow

MelissaAutumn:
Fixes #912 

I've replaced the simple username check with how we check for in-use emails during sign-up. This affects the debounced api call during step 1 of sign-up.

I took the time to add tests as there weren't any for this route. (d'oh.)


davinotdavid (approved):
lgtm!

## QA Log

- Code diff (familiar with the code flows)

PR #914, opened 2026-06-03
Create beat task to purge stale accounts

### LINKED ISSUE #964: Purge / Delete stale accounts after a X number of days
We are currently tagging stale users with the `abandoned_cart` Mailchimp tag for automations but after a certain number of days, we should be actually purging / deleting said accounts if they didn't convert.

It would be good to synchronize this with Marketing (@inattee) and UX (@laurelterlesky) since there's a concern that we need to send another follow up email upon deletion to avoid support tickets / better UX (@lisajill).

Base work if mostly done and reviewed in this PR https://github.com/thunderbird/thunderbird-accounts/pull/914 pending the necessity of crafting / sending a new email upon deletion.

Related:
https://github.com/thunderbird/thunderbird-accounts/pull/914
https://github.com/thunderbird/thunderbird-accounts/issues/896#issuecomment-4709532604

davinotdavid:
## What changed?

- Added a new celery beat task that runs **daily** on 3 AM UTC (or whichever time we set in the new `INCOMPLETE_SIGNUP_PURGE_*` vars) and deletes user accounts that are stale. We consider stale accounts that are:
  - Older than 72 hours and;
  - `has_subscription = True` and;
  - Not test / staff / superuser accounts

## Why?

Stale user accounts still hold their chosen usernames, prevent new users from getting them. We should be cleaning them up regularly.

## How to test (steps)

- To make testing easier, navigate to `src/thunderbird_accounts/authentication/tasks.py` and update the `cutoff` variable to be just `timezone.now()`. Also, update the `User.filter` to be just `created_at__lt=cutoff`.
- Navigate to localhost:8087, login using `admin@example.org` / `admin` (or create a new account through the FTUE but that would be a little more annoying as it currently requires click-ops in Keycloak to enable it)
- Bash into the celery docker container and run `uv run celery -A thunderbird_accounts call thunderbird_accounts.authentication.tasks.purge_incomplete_signups`
- Check the celery docker container logs and see that the deleted account count is 1, like this log example:

```
[2026-06-03 17:43:06,914: INFO/ForkPoolWorker-8] Task thunderbird_accounts.authentication.tasks.purge_incomplete_signups[ccdaaba7-d4b8-4a31-bb37-eb29ecf03ca0] succeeded in 0.1860894580022432s: {'task_status': 'completed', 'deleted': 1, 'errors': 0, 'skipped': 0}
```

- Reload the page and see that you are logged out and can't log in again

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/964


src/thunderbird_accounts/authentication/tests/test_tasks.py
```diff
@@ -0,0 +1,139 @@
...
+        user.refresh_from_db()
+        return user
+
+    def test_includes_stale_user_without_subscription(self):
+        user = self._create_user('stale-no-sub')
+
+        self.assertIn(user, get_stale_incomplete_signup_users())
+
+    def test_includes_stale_user_awaiting_payment_without_subscription(self):
+        user = self._create_user('stale-awaiting', is_awaiting_payment_verification=True)
```
markstos, line 65: > is_awaiting_payment_verification=True

I don't believe purging users in this state is correct and recommend excluding them from the purge.

Users get into the `is_awaiting_payment_verification` state when  the user has reached a Paddle transaction status of paid or completed, and the app is waiting for the subscription webhook to create/activate  the local subscription. Our UI tells people that the payment verification can take "up to 48 hours". 

We should probably have automation for this category after 72-hours as well, but not by purging them. They have tried to pay and there's a problem. A better resolution for this group is see if we can automate a re-check with Paddle in case we somehow missed a webhook, or nudge the user to tell them their payment didn't go through after 72 hours and suggest they try to pay again, which might involve cancelling the initial transaction with Paddle so they can try again without fear of getting double-billed. I'm not a payments expert though we might want to ask Paddle about their best practices here.

davinotdavid: Oh great catch! Thanks! Indeed we shouldn't include users awaiting for payment verification, updated the code.

Regarding the resolution for the group who are in payment limbo, I believe that's out of scope for this PR / issue but definitely something that we should take a look at. [We have some mitigation plans in case we miss a webhook though](https://github.com/thunderbird/thunderbird-accounts/issues/838#issuecomment-4461093186)!

MelissaAutumn:
I'll add this discussion for the thundermail meeting, but I think we should push the timeout to at least one week. If not one week then the time between our invite waves.

davinotdavid:
Moving this PR to draft as scope changed. The idea is to explore the possibility of having a tag in Mailchimp for building email automations before we actually purge the accounts.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -1,22 +1,27 @@
...
+
+logger = logging.getLogger(__name__)
 
 logger = logging.getLogger(__name__)
 
 
-def get_abandoned_cart_users_for_mailchimp_tag() -> QuerySet[User]:
-    """Incomplete sign-ups old enough to receive the abandoned_cart Mailchimp tag."""
-    cutoff = timezone.now() - timedelta(hours=settings.ABANDONED_CART_TAG_HOURS)
+def get_stale_incomplete_signup_users(cutoff_hours: int = settings.ABANDONED_CART_TAG_HOURS) -> QuerySet[User]:
```
MelissaAutumn: Since it's now shared, we should remove the 1hr default as I forsee a typo causing a big headache.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -100,3 +105,61 @@ def tag_abandoned_cart_in_mailchimp(self):
...
+    """
+    deleted = 0
+    errors = 0
+    skipped = 0
+
+    for user in get_stale_incomplete_signup_users(cutoff_hours=settings.INCOMPLETE_SIGNUP_PURGE_HOURS).iterator():
+        try:
+            with transaction.atomic():
+                user = User.objects.select_for_update().get(pk=user.pk)
+                if user.subscription_set.exists():
```
MelissaAutumn, line 127: Can you add a comment along the lines of "Test the user against some of the conditions now that we have a locked fresh copy." (Maybe make it betterer than what I wrote)

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -100,3 +105,61 @@ def tag_abandoned_cart_in_mailchimp(self):
...
+
+                if user.is_awaiting_payment_verification:
+                    skipped += 1
+                    logger.info(
+                        'purge_incomplete_signups: skipped %s because payment verification is pending',
+                        user.username,
+                    )
+                    continue
+
+                purge_errors = delete_user_data(user)
```
MelissaAutumn, line 140: Do we have a off-boarding email? If not we should probably have one that is sent out at the same time of this being "Your data has been removed due to inactivity" or something. 

Otherwise we might get support messages being like "I signed up and now my account doesn't work" even though they haven't paid.

.env.example
```diff
@@ -138,3 +138,8 @@ ABANDONED_CART_TAG_HOURS=1
 ABANDONED_CART_MAILCHIMP_TAG=abandoned_cart
 
 FLOWER_UNAUTHENTICATED_API=True
+
+# Used to purge incomplete signups from the database after a certain number of hours.
+INCOMPLETE_SIGNUP_PURGE_HOURS=168
```
MelissaAutumn: Can you add a comment that this is 7 days

MelissaAutumn: Also can you add pulumi defaults?

src/thunderbird_accounts/settings.py
```diff
@@ -535,6 +535,8 @@ def oidc_logout(request):
 
 CELERY_BEAT_SCHEDULE = {}
 
+INCOMPLETE_SIGNUP_PURGE_HOURS = int(os.getenv('INCOMPLETE_SIGNUP_PURGE_HOURS', '72'))
```
MelissaAutumn: This is different fom env

MelissaAutumn (changes requested):
Few comments

.env.example
davinotdavid: Thanks, added! I've forgotten to add the abandoned cart pulumi defaults to `config.prod.yaml` so I've added those now too. 🙇 

Also, I didn't add the abandoned cart pulumi defaults to `stage` as Mailchimp doesn't run there.

src/thunderbird_accounts/settings.py
davinotdavid: Oh good catch, thanks! Updated here and added the correct value to the pulumi config as well.

src/thunderbird_accounts/authentication/tasks.py
davinotdavid: AFAIK we don't have it but I believe this is by design as the abandoned cart mailchimp automation was done so that marketing can reach out before the accounts get purged.

I believe it was mentioned that the copy in that email would explicitly say that the account would be deleted in 7 days but since I believe this automation is not yet in place, it would be nice to confirm. cc @inattee

MelissaAutumn: We might want to hold this PR until that's in--place. I think it's a pretty poor UX even if they haven't paid us any money yet.

davinotdavid: There was some chatter in Matrix about this with Marketing and @kewisch and the summary was:

- Initially, we are not sending a off-boarding email
- We will observe the conversion of the `abandoned_cart` first
- In the future, re-assess if we need an off-boarding email

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -100,3 +104,64 @@ def tag_abandoned_cart_in_mailchimp(self):
...
+                if user.subscription_set.exists():
+                    skipped += 1
+                    logger.info('purge_incomplete_signups: skipped %s because a subscription exists', user.username)
+                    continue
+
+                if user.is_awaiting_payment_verification:
+                    skipped += 1
+                    logger.info(
+                        'purge_incomplete_signups: skipped %s because payment verification is pending',
+                        user.username,
```
MelissaAutumn:
```suggestion
                        user.uuid,
```

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -100,3 +104,64 @@ def tag_abandoned_cart_in_mailchimp(self):
...
+    for user in get_stale_incomplete_signup_users(cutoff_hours=settings.INCOMPLETE_SIGNUP_PURGE_HOURS).iterator():
+        try:
+            with transaction.atomic():
+                user = User.objects.select_for_update().get(pk=user.pk)
+
+                # Now that we have a locked fresh copy of the user, we can safely check for subscriptions
+                # and payment verification so we avoid race conditions
+                if user.subscription_set.exists():
+                    skipped += 1
+                    logger.info('purge_incomplete_signups: skipped %s because a subscription exists', user.username)
```
MelissaAutumn:
```suggestion
                    logger.info('purge_incomplete_signups: skipped %s because a subscription exists', user.uuid)
```

MelissaAutumn (approved):
let's give it a try on stage

PR #916, opened 2026-06-03
Consolidate is_address_taken for sign-up and check alt domains. (Fixes #915)

### LINKED ISSUE #915: Prevent signing up with a in-use tb.pro alias
If someone tries to sign up with hello and `hello@thundermail.com` is available but `hello@tb.pro` is not then we should fail.

MelissaAutumn:
Fixes #915 

Since we mirror the username against our shared aliases this will prevent a case where `hello@thundermail.com` is available but someone has already taken `hello@tb.pro`.


davinotdavid (approved):
lgtm, we can do some further manual testing in staging

## QA Logs

- Reviewed code diff
- Tested manually locally

PR #919, opened 2026-06-03
Re-verifying a custom domain should no longer blank the stalwart_id (Fixes #918)

### LINKED ISSUE #690: unable to setup custom domain if used in different tb pro account
(NOTE: I searched https://accounts.tb.pro/admin/mail/domain/ for <domain1> and couldn't find it)

### Step to reproduce

In Zendesk ticket [4567](https://tbpro.zendesk.com/agent/tickets/4567), Malini removed  \<domain1\> from \<tbproacount1\> and then tried to add \<domain1\> to another tbproaccount let's call it \<tbproaccount2\>. and it failed!

### Expected result

* The custom domain can be added.

### Actual  result

* "This domain is already configured. Please reach out to support in case you believe this is a mistake."
 gz#4567

### LINKED ISSUE #918: Domain objects do not have their stalwart ids set
Verified domains do not have their stalwart ids set. This prevents a domain from properly being removed on stalwart's end. 

This used to work (as indicated by some domains do indeed having their id) but no longer. Could be stalwart regression.

MelissaAutumn:
Fixes #918 
Fixes #690 

This should fix the rest of the domains problems, we might have to run "Fix missing stalwart ids" on some existing domains though.

I believe that Stalwart is returning a 200 if a domain already exists which is weird? So if a domain is verified and does another round of verification create_domain returns None, and then we blindly set `stalwart_id = None` :I 

So this fixes the issue by checking if the domain already exists or if it's verified. If the domain is missing on stalwart ends a reverification will actually fix it for the user, and if the domain isn't verified but has a stalwart_id somehow we won't blank it out. 

I've updated the tests with the new test cases, we might want to look into #152 sooner than later hah.

### DIFF
--- a/src/thunderbird_accounts/mail/views.py
+++ b/src/thunderbird_accounts/mail/views.py
@@ -28,7 +28,8 @@ from thunderbird_accounts.mail.exceptions import (
     AccessTokenNotFound,
     AccountNotFoundError,
     DomainAlreadyExistsError,
-    DomainNotFoundError, EmailNotValidError,
+    DomainNotFoundError,
+    EmailNotValidError,
 )
 from thunderbird_accounts.mail.utils import (
     check_stale_dns_records,
@@ -253,7 +254,7 @@ def verify_custom_domain(request: HttpRequest):
         else:
             is_verified = True
             critical_errors = []
-            warnings = ['Custom domain DNS verification disabled. Automatically verified domain.']
+            warnings = [_('Custom domain DNS verification disabled. Automatically verified domain.')]
 
         domain.last_verification_attempt = now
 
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/mail/views.py
```diff
@@ -274,22 +275,38 @@ def verify_custom_domain(request: HttpRequest):
...
+                # Fetch or create a domain on Stalwart's end, and retrieve the domain_id
+                if stalwart_resp:
+                    domain_id = stalwart_resp.get('id')
+                else:
+                    domain_id = stalwart_client.create_domain(domain_name)
+
+                # Ensure missing DKIM selectors without replacing keys created at domain-add time.
+                stalwart_client.ensure_dkim(domain_name)
+
+                mail_tasks.publish_hosted_dkim_dns_records.delay(domain_name)
```
MelissaAutumn, line 299: @markstos I messaged you about this on matrix, but here's the context.

Is this okay to run potentially a million times?

markstos: It looks like it could retry up to 24 times for a given case, and can only be triggered by authenticated users and is safe in terms of both Stalwart and Cloudflare APIs as far as I'm aware. Given that our tasks have some backoff and rate-limiting, I don't anticipate hitting Cloudflare's limits of 200/second requests or 12000/5-minutes.

markstos (approved):
## QA Log

- Reviewed diff
- Noted test coverage
- Researched implications of repeated "publish DKIM record calls."

PR #922, opened 2026-06-03
Add more comprehensive reserved word lists with the help of open sour…

Sancus:
This adds reserved word lists from three sources:

- https://github.com/forwardemail/reserved-email-addresses-list
- https://github.com/shouldbee/reserved-usernames
- https://github.com/miketromba/reserved-slugs

Certain things are excluded from the last one due to being excessive for our use case or not our business imo(profanity), they are `('country-codes', 'languages', 'profanity')`.

These sources are vendored here and an update script is provided for updating them in the future should it be necessary. I'm not too sure that that's the right place for the update script, and technically it could be a Django management command, but it seems a bit silly.

## Reserved Word Files

- `generated-words.json` - Pulled from upstream sources via update.py and checked in.
- `exact-words.json` - Exact word matches only, no wildcard.
- `affix-words.json` - Matches anything with the word in prefix or suffix. So admin would match 'mozadmin' and 'adminfx' and so on.

I moved a significant amount of logic out of `checker.py` into these word lists, but some of the regexes we had before don't cleanly map so I left those ones in `checker.py` alone. As a result of the restructure, the reserved word checker becomes a package.

## Notes
This includes a bunch of unicode homographs, and while we don't currently allow unicode, it doesn't hurt anything and it might save us a footgun if we do allow unicode in the future. However, I can remove it if desired as well.

This list is meant to be broad, and removing things from it would be acceptable, but I think it's better to reserve first and ask questions later if need be.


src/thunderbird_accounts/authentication/reserved/checker.py
```diff
@@ -0,0 +1,127 @@
...
+    affix = _normalized(generated['affix']) | _normalized(_read_json(_AFFIX_WORDS_FILE))
+    return frozenset(exact), frozenset(affix)
+
+
+# RESERVED_LOCAL_PARTS is matched by exact comparison only;
+# AFFIX_RESERVED_LOCAL_PARTS is matched by exact comparison, prefix, or suffix.
+RESERVED_LOCAL_PARTS, AFFIX_RESERVED_LOCAL_PARTS = _load_words()
+
+
+def is_reserved(test_string: str) -> bool:
```
MelissaAutumn, line 139:
```suggestion
@cache
def is_reserved(test_string: str) -> bool:
```

We'll need to import cache, but this is a good candidate for func caching.

src/thunderbird_accounts/authentication/reserved/update.py
```diff
@@ -0,0 +1,167 @@
...
+    # An affix entry already matches its own exact form, so keep the sections disjoint.
+    exact -= affix
+
+    _write_words(exact, affix)
+    print(f'{GENERATED_WORDS_FILE}: {len(exact)} exact + {len(affix)} affix entries')
+    print(f'(reserved-slugs excluded categories: {", ".join(RESERVED_SLUGS_EXCLUDED)})')
+    print('Done. Bump the README "Retrieved" dates and re-run the authentication tests.')
+
+
+if __name__ == '__main__':
```
MelissaAutumn: This should probably be a django management command

src/thunderbird_accounts/authentication/reserved/checker.py
```diff
@@ -0,0 +1,127 @@
...
+    names = {_normalize_local_part(word) for word in words}
+    names.discard('')
+    return names
+
+
+def _load_words() -> tuple[frozenset[str], frozenset[str]]:
+    generated = _read_json(_GENERATED_WORDS_FILE)
+    exact = _normalized(generated['exact']) | _normalized(_read_json(_EXACT_WORDS_FILE))
+    affix = _normalized(generated['affix']) | _normalized(_read_json(_AFFIX_WORDS_FILE))
+    return frozenset(exact), frozenset(affix)
```
MelissaAutumn, line 110: Can we wrap a try except here? If load json file errors out this will probably cause a import error on this file which would effectively break the app.

MelissaAutumn (changes requested):
Doesn't have to be in this ticket (since this feels like a rabbithole), but you can bypass a lot of this with `-`. Like `-admin-` or `main-admin`. 

Might be worth splitting words based on common separator characters and running exact, startswith, endswith checks on those.

src/thunderbird_accounts/authentication/reserved/update.py
MelissaAutumn: (I know management commands are a bit silly, but it at least keeps things in a pattern)

Sancus:
> Doesn't have to be in this ticket (since this feels like a rabbithole), but you can bypass a lot of this with `-`. Like `-admin-` or `main-admin`.
> 
> Might be worth splitting words based on common separator characters and running exact, startswith, endswith checks on those.

Right. That's kind of inherent to prefix/suffix matching.  Maybe I just do wildcard matching? It's only 61 terms right now (plus alternate spellings/homographs).

MelissaAutumn:
Would wildcard matching detect the `me` in `meagan`? If so I think we can leave that for future pondering and just roll with this.

Sancus:
> Would wildcard matching detect the `me` in `meagan`? If so I think we can leave that for future pondering and just roll with this.

If 'me' is on that list, yes it would. It's not right now. But yes, in general short terms are problematic on a wildcard match. Right now the "affix" list is:

```
abuse                administracion       administration       administrator
admin                billing              boletins             bounce
bounces              conserie             conserje             contabilidad
contact              contacto             contato              copyright
customer             developer            do-not-reply         do-not-respond
donotreply           donotrespond         email                example
help                 hostmaster           imap                 legal
license              mailer-daemon        mailer.daemon        mailerdaemon
mzla-test            naoresponda          no-reply             no-replys
no_reply             noc                  noreply              noreplys
policy               postmaster           privacy              root
security             smtp                 soporte              ssl
superuser            suporte              support              sys.administrator
sysadmin             sysadministrator     system-administrator system.administrator
systemadministrator  terms                tls                  vendas
webmaster
```

MelissaAutumn:
rip `rootbeer at thundermail dot com`

src/thunderbird_accounts/authentication/reserved/update.py
Sancus: It does mean you need the whole app working to update these files, as opposed to just a couple python libraries. But I suppose anyone updating these files will likely have it working anyway so not a big deal. I'll change that.

Sancus:
> rip `rootbeer at thundermail dot com`

OK, so I went with making the affix filter "separator aware", so you can't avoid the prefix/suffix matching by just sticking separators in. I've added some tests to this effect. This DOES make it significantly broader, with more false positives, but at least it's still less oppressive than full wildcard matches on some of these terms.

Management command added, function cache, and try/catch block added.

PR #926, opened 2026-06-04
Refactor: extract DNS functions from util.py

markstos:
## What changed?

Extracts DNS functions from `util.py` into `dns.py`

This PR is purely a refactor. There are no logic changes.

## Why?

Util should be a namespace of last resort as it as doesn't communicate
much useful about what it contains. In this case, over half of the
util namespace had become DNS-related functions, a category which is
likely to grow over time.

For clarity and file-size management, DNS functions have been moved
into their own namespace.

## Limitations and Notes

The existing `dkim.py` file is left standalone to minimize churn and
because it's alreaady sizable on its own. One DKIM function remains
as in the new `dns.py` file as it's a more natural fit there.

One other small refactor was made to avoid nested "mail.dns.dns" namespace:

Instead, we use `dns_rdatatype` and `dns_resolver` instead:

```
-import dns.rdatatype
-import dns.resolver
+import dns.rdatatype as dns_rdatatype
+import dns.resolver as dns_resolver
```


MelissaAutumn (approved):
Looks good but I think you forgot to include test_dns.py. 

Okay to merge once that's included and tests are passing.

PR #929, opened 2026-06-04
Lowercase alias and domain names, and switch is_address_taken to iexact. (Fixes #927)

### LINKED ISSUE #927: Lowercase email aliases on save
Sentry: https://thunderbird.sentry.io/issues/7527178088/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

We should be saving aliases as lowercase to reduce case sensitivity issues. (Optionally we can also use address__iexact lol.)

MelissaAutumn:
Fixes #927 

This might need some deeper thought than I've put in it so far, but I think this captures all the footguns. 

is_address_taken is used in sign-up flow, and the aliases/custom domain ones should cover those user inputs on the dashboard. I've added a single test case that should cover is_address_taken. 

Nit away!

### DIFF
--- a/src/thunderbird_accounts/mail/tests/test_views.py
+++ b/src/thunderbird_accounts/mail/tests/test_views.py
@@ -797,14 +797,34 @@ class AddEmailAliasTestCase(TestCase):
         """Ensure that creating an email alias that is already in-use will error out."""
         email_alias_url = reverse('add_email_alias')
 
-        Email(address=f'admin@{settings.PRIMARY_EMAIL_DOMAIN}').save()
+        Email(address=f'cool_dude_99@{settings.PRIMARY_EMAIL_DOMAIN}').save()
 
         with patch('thunderbird_accounts.mail.views.MailClient', Mock()) as mock:
             instance = Mock()
             mock.save_email_addresses = instance
             response = self.client.post(
                 email_alias_url,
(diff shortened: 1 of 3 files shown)


markstos (commented):
What about about putting the case-insensitive comparison inside `is_address_taken`?  Is there a case where we have callers would want that to be case-sensitive? 

If we can make this change, not only would it slim up current code, but it reduces the risk of adding a future caller which fails to add the case-insensitive comparison. 

## QA Log

- Read diff
- Learned about DJango's __iexact syntax.
- Reviewed call sites for `is_address_taken`
- Researched whether Stalwart stores emails case-insensitive or not. Inconclusive, but it does lowercase emails in a number of cases for the same practical reasons we are considering it here. https://github.com/stalwartlabs/stalwart/issues/1916

MelissaAutumn:
> What about about putting the case-insensitive comparison inside `is_address_taken`? Is there a case where we have callers would want that to be case-sensitive?

I did add the case-insensitive check to is_address_taken, but there's some other spots where existing non-lowercased data might fail a lookup like on domain/alias creation/deletion. 

I'm not entirely happy spamming `.lower()` everywhere, maybe a middleware or a function decorator would be nicer in the future though.

markstos (approved):
## QA Log

- Reviewed diff
- Confirmed that DRY'ing the code to move the case check into inner function isn't viable here.

PR #930, opened 2026-06-04
Don't raise an error if we pop a session key that doesn't exist on paddle complete (Fixes #925)

### LINKED ISSUE #925: KeyError: 'paddle_txid' - on paddle transaction complete
Sentry: https://thunderbird.sentry.io/issues/7462241217/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

Our paddle transaction complete function pops session ids, however for whatever reason those session ids don't exist. In the above case I can confirm the user is subscribed successfully. So the webhook is working as intended this is just a bad UX.

Wrt to the function and key error, it's actually just popping the session keys to eject them from session (and also check if we're in a paypal flow.) We should set a default value of `None`, which will avoid raising any key errors but also ship them to the success url. 

It'll be a little awkward if this occurs during a Paypal popup transaction but there's not much we can do here.

MelissaAutumn:
Fixes #925 

I can only guess that this is caused by blocking our functional session cookie. (Note: We store session data in the database as is django's default method, but the session cookie contains the id to retrieve the session data. We do not store session data in the cookie itself!)

Important to note that nothing about this bug affects the payment transaction. The user still gets an account once we get the webhook from paddle. This is a UX issue and while still a little awkward if they're paying with paypal, this should fix the 500 error.

### DIFF
--- a/src/thunderbird_accounts/subscription/views.py
+++ b/src/thunderbird_accounts/subscription/views.py
@@ -122,11 +122,12 @@ def paddle_transaction_complete(request: HttpRequest, paddle: Client):
     in our db via webhooks.
     """
     user = request.user
-    transaction_id = request.session.pop(SESSION_PADDLE_TRANSACTION_ID)
-    payment_type = request.session.pop(SESSION_PADDLE_PAYMENT_TYPE)
+    transaction_id = request.session.pop(SESSION_PADDLE_TRANSACTION_ID, None)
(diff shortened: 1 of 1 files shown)


markstos (approved):
## QA Log

- Reviewed issue, description, diff

davinotdavid (approved):
## QA Log

- Reviewed diff

PR #932, opened 2026-06-04
Disable server side cursors (Fixes #931)

### LINKED ISSUE #931: InvalidCursorName: DB Pooling causing db cursor errors
Sentry: https://thunderbird.sentry.io/issues/7461177063/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

This can be triggered easily with the admin panel, and it SEEMS to coincide with neon's pg pooling.

MelissaAutumn:
Fixes #931 

According to django docs (https://docs.djangoproject.com/en/5.2/ref/databases/#transaction-pooling-server-side-cursors) it seems like django will hold onto the connection in order to speed things up. However this doesn't work if we're punted off to another db connection mid-call. (my understanding)

This is currently happening explicitly in the admin panel when you open a bunch of records in new tabs, but this could affect regular end-users. 

Best to fix this with the nuclear option and then look at how we can resolve this wrt pooling later.

### DIFF
--- a/src/thunderbird_accounts/settings.py
+++ b/src/thunderbird_accounts/settings.py
@@ -223,7 +223,7 @@ WSGI_APPLICATION = 'thunderbird_accounts.wsgi.application'
 # Database
 # https://docs.djangoproject.com/en/5.1/ref/settings/#databases
 AVAILABLE_DATABASES = {
-    'dev': {
+    'main': {
         'default': {
             'ENGINE': 'django.db.backends.postgresql',
             'NAME': os.getenv('DATABASE_NAME'),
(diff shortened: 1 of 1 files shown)


Sancus:
https://neon.com/docs/guides/django

Neon's Django guide recommends 
```
    'DISABLE_SERVER_SIDE_CURSORS': True,
    'CONN_HEALTH_CHECKS': True,
```

with the additional note: 
>From Django 4.1, you can use a higher CONN_MAX_AGE setting in combination with the [CONN_HEALTH_CHECKS](https://docs.djangoproject.com/en/4.1/ref/settings/#conn-health-checks) setting to enable connection reuse while preventing errors that might occur due to closed connections.

MelissaAutumn:
`CONN_MAX_AGE` seems to default to `0` in 5.2, but I've added `CONN_HEALTH_CHECKS` just to cover bases here.

markstos (approved):
## QA Log

- Confirmed the "dev" to "main" change is improves clarity and has no other other impact
- Confirmed the configuration changes match what is recommended by Neon for Django. https://neon.com/docs/guides/django

PR #936, opened 2026-06-04
Adjust fix_archives_folder to pass if it already exists. (Fixes #935)

### LINKED ISSUE #935: archives folder fix
Sentry: https://thunderbird.sentry.io/issues/7088369914/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

I want this error nuked from orbit!

MelissaAutumn:
Fixes #935 

I'm not sure why, but the mailbox query on some mailboxes seem to return an empty list. However the set returns an error that we can identify as "this folder already exists with this role." So if we get that we can mark the user as having an archives folder.

(For context: Stalwart was having issues making archives folders for a while, so this was a quick fix.)

### DIFF
--- a/src/thunderbird_accounts/mail/utils.py
+++ b/src/thunderbird_accounts/mail/utils.py
@@ -480,7 +480,7 @@ def fix_archives_folder(access_token, account: Account) -> bool:
                     [
                         'Mailbox/query',
                         {
-                            'accountId': account_id,
+                            'accountId': str(account_id),
                             'filter': {'role': 'archive'},
                         },
                         '0',
@@ -494,16 +494,16 @@ def fix_archives_folder(access_token, account: Account) -> bool:
             # If they don't create a new inbox with the role 'archive', named 'Archives'
             # (set in settings.STALWART_ARCHIVES_FOLDER_NAME) which is subscribed by default
             temp_id = str(uuid.uuid4())
-            inbox_res = client.make_jmap_call(
+            set_res = client.make_jmap_call(
                 {
                     'using': ['urn:ietf:params:jmap:core', 'urn:ietf:params:jmap:mail'],
                     'methodCalls': [
                         [
                             'Mailbox/set',
                             {
-                                'accountId': account_id,
+                                'accountId': str(account_id),
                                 'create': {
-                                    temp_id: {
+                                    str(temp_id): {
                                         'name': settings.STALWART_ARCHIVES_FOLDER_NAME,
                                         'role': 'archive',
                                         'isSubscribed': True,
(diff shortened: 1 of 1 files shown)


src/thunderbird_accounts/mail/utils.py
```diff
@@ -480,7 +480,7 @@ def fix_archives_folder(access_token, account: Account) -> bool:
                     [
                         'Mailbox/query',
                         {
-                            'accountId': account_id,
+                            'accountId': str(account_id),
```
MelissaAutumn, line 483: At some point Stalwart moved from using numeric ids to random characters. So this is an attempt to cover that.

src/thunderbird_accounts/mail/utils.py
MelissaAutumn, line 492: The query returns a response like:
```python
{
            'methodResponses': [
                [
                    'Mailbox/query',
                    {'accountId': 'c', 'queryState': 'sai', 'canCalculateChanges': True, 'position': 0, 'ids': ['h']},
                    '0',
                ]
            ],
            'sessionState': '3e25b2a0',
}
```

src/thunderbird_accounts/mail/utils.py
markstos: Thanks for the explanation.

src/thunderbird_accounts/mail/utils.py
```diff
@@ -544,25 +544,29 @@ def fix_archives_folder(access_token, account: Account) -> bool:
...
             if not return_response:
                 return_response = method_response.get('notCreated', {})
                 # The only way we're getting out of here without an error, is if the folder already exists
                 desc = return_response.get(temp_id, {}).get('description')
+                
                 if desc and desc in already_exists_descriptions:
-                    raise Exception(f'Failed to create archive folder: {desc}')
+                    pass # If it already exists then we can actually mark it as done and move on
+                else:
+                    sentry_sdk.set_extra('desc', desc)
```
markstos: `set_extra` has been deprecated. The recommend replacement  is `set_context` which has a similar (or identical?) function signature:

https://docs.sentry.io/platforms/python/enriching-events/context/

Out of scope: There are a few other places in the code base where we still use `set_extra`.

markstos (changes requested):
The code looks good, but could use test coverage for the case. 

There's also a nit with the recommended Sentry method to use.

src/thunderbird_accounts/mail/utils.py
MelissaAutumn: Thanks for the pointer! I'll fix these ones and created https://github.com/thunderbird/thunderbird-accounts/issues/946 to deal with the rest.

markstos (approved):
## QA Log

- Follow-up Sentry refactor looks good.

PR #938, opened 2026-06-04
Fix alias count showing the incorrect count when filtering results (Fixes #937)

### LINKED ISSUE #937: Fix email alias count in dj admin
Seems a wee bit bugged.

MelissaAutumn:
The alias count now works correctly when filtering results. 

I found https://books.agiliq.com/projects/django-admin-cookbook/en/latest/optimize_queries.html which was quite helpful.

Fixes #937


kewisch (approved):
Good catch, thank you. I figured it must be something around distinctness, but wasn't sure why the user we found this with would have around 220 of 225 non-distinct email addresses.

PR #940, opened 2026-06-04
Switch how we default zoneinfo/locale to their default values. (Fixes #939)

### LINKED ISSUE #939: zoneinfo can be submitted with a key but no value
Sentry: https://thunderbird.sentry.io/issues/7227985128/events/7e8ad32c22ba4cc2a086e080b32b756f/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=next-event

If zoneinfo is submitted without a value the dict.get() default value does not get pulled in, and thus zoneinfo will remain None. 

This is bad as keycloak wants a value.

MelissaAutumn:
Fixes #939 

If someone's browser doesn't want to return a timezone then we'll get `zoneinfo: None`. Because the key is there the data.get('zoneinfo')'s default value will not be pulled and zoneinfo will continue to be None. However we need zoneinfo as a non null value. So we flip zoneinfo (and locale) to use `or $val`. 

In order to better test for this I added the positional argument thingy to import_user, so now if you want to specify timezone it has to be via kwargs. This allows us to test the value in the mock call.

### DIFF
--- a/src/thunderbird_accounts/authentication/admin/forms.py
+++ b/src/thunderbird_accounts/authentication/admin/forms.py
@@ -98,7 +98,7 @@ class CustomNewUserForm(CustomUserFormBase):
             keycloak_pkid = keycloak.import_user(
                 self.cleaned_data.get('username'),
                 self.cleaned_data.get('recovery_email'),
-                self.cleaned_data.get('timezone'),
+                timezone=self.cleaned_data.get('timezone'),
                 name=name,
             )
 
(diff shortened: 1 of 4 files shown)


markstos (approved):
## QA Log

- Reviewed issue, PR description, diff
- Reviewed Python's positional vs keyword handling
- Considered if a default timezone should be set in `import_user()`. The approach here is good because there are cases to cover outside of `import_user()`.

PR #945, opened 2026-06-05
Pre-fill contact form with recovery email for users without active subscription

### LINKED ISSUE #923: Contact form improvements for mid-signup users - use recovery email
When someone is mid-signup on Thundermail and fills in the contact form, the contact form fills in their @thundermail.com address.  At this point, their account is not active. 

This means that they do not get our auto-replies and we can not reply to them without looking up their recovery address.

**Suggested solution:** If the user does not have a Plan attached to their account, auto-fill their recovery email instead of their @thundermail.com email.

Example issue: https://tbpro.zendesk.com/agent/tickets/6037

davinotdavid:
## What changed?

- `recovery_email` is now being passed in the `_page` context for the Vue app so that we can pre-fill it in the contact form in case the logged in user does not have an active subscription yet
- Added E2E tests (+ new .env var `ACCTS_OIDC_RECOVERY_EMAIL` and secret `E2E_ACCTS_OIDC_RECOVERY_EMAIL`)
- Sneaking in a tiny fix for the contact form UI in wider viewports

## Why?

For users that don't have an active subscription yet, we can't contact them back through their @thundermail.com address so we should prefill the contact form with the recovery email instead in those cases.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/923

## Screenshots

https://github.com/user-attachments/assets/8c06245a-0eaf-4d31-ae60-26476be1d385

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -126,7 +126,7 @@ Then edit your local `.env` file and provide the following values:
 ACCTS_HUB_URL=<URL-for-TB-Accts-Hub>
 ACCTS_OIDC_EMAIL=<existing-tbpro-user-email>
 ACCTS_OIDC_PWORD=<exisiting-tbro-password>
-ACCTS_
+ACCTS_OIDC_RECOVERY_EMAIL=<existing-tbpro-user-recovery-email>
 PRIMARY_THUNDERMAIL_EMAIL=<primary-thundermail-email-address>
 ```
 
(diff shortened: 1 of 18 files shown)


test/e2e/README.md
```diff
@@ -125,6 +126,7 @@ Then edit your local `.env` file and provide the following values:
 ACCTS_HUB_URL=<URL-for-TB-Accts-Hub>
 ACCTS_OIDC_EMAIL=<existing-tbpro-user-email>
 ACCTS_OIDC_PWORD=<exisiting-tbro-password>
+ACCTS_
```
MelissaAutumn: 👀

MelissaAutumn (changes requested):
Looks good just a small mistake on readme.

test/e2e/README.md
davinotdavid: Holy macaroni, weekend needed 😅 thanks!

MelissaAutumn (approved):
Code works on my local machine, I haven't ran the e2ee tests locally as they seem fine on CI.

PR #947, opened 2026-06-05
Consider missing DKIM DNS records an error (#942)

### LINKED ISSUE #942: Make missing DKIM CNAMEs an error instead of a warning
If they don't set it up deliverability will suffer, which isn't good for the user and will also likely result in higher support volumes. As long as they are updating their DNS records, there's no good reason to not add the DKIM records while they are at it.

markstos:
## What changed?

Missing DKIM DNS records are now considered an error.

## Why?

If they don't set it up deliverability will suffer, which isn't good for the user and will also likely result in higher support volumes. As long as they are updating their DNS records, there's no good reason to not add the DKIM records while they are at it.

## Applicable Issues

Closes #942

## QA Log

- Auto tests
- Tested locally with missing DKIM errors, observed critical error treatment.

## Screenshots

image


davinotdavid (approved):
## QA Log

- Checked diff
- Ran tests
- Can verify that it now shows dkim records as critical:

Image

--

Note (to self?): this UI will change very soon https://github.com/thunderbird/thunderbird-accounts/pull/955#issuecomment-4696180163

PR #949, opened 2026-06-05
Update DMARC validation to pass any record that starts with V=DMARC1 (#941)

### LINKED ISSUE #941: Allow any kind of DMARC rule as long as it is syntactically correct

markstos:
## What changed?

Before, the logic was doing an exact string match, which is not what we want here.

Also, update tests for _compare_semantic_txt to not use DMARC in examples
since we no longer pass DMARC records through that method.

## Why?

We want to support all DMARC policies.

## Limitations and Notes

As seen in screenshot, the current design displays the *example* row in validation.
Since a row that looks *different* but is *valid DMARC* may pass, there is no conflict,
thus now error is displayed.

We can consider changing this UX during the upcoming UX overhaul.

Also, at this time we not trying to fully syntactically verify all possible DMARC syntax details. Our design goals are to:

1. Provide an example if they have none or don't know what to do
2. Support other valid DMARC syntax.

## Applicable Issues

 * Closes #941

# QA Log

- Automated test coverage
- Manual testing (see screenshot)

## Screenshots

image


src/thunderbird_accounts/mail/dns.py
```diff
@@ -139,6 +139,15 @@ def _compare_dkim_txt(expected_content: str, live_values: list[str]) -> tuple[DN
     return DNSRecordStatus.CONFLICT, dkim_values
 
 
+def _verify_dmarc(live_values: list[str]) -> tuple[DNSRecordStatus, list[str]]:
+    dmarc_values = [value for value in live_values if value.startswith('v=DMARC1')]
```
kewisch: Would recommend we make this a bit more elaborate. This would allow `v=DMARC123` as well. We should check if it follows the tag-value format from https://datatracker.ietf.org/doc/html/rfc7489#section-6.3 and the linked https://www.rfc-editor.org/info/rfc6376/#section-3.2 , and that all required keys (`p` being one of the valid receiver policies, and `v` being DMARC1`.

kewisch (changes requested):

src/thunderbird_accounts/mail/dns.py
markstos: Commit has been pushed to address this.

kewisch (approved):

kewisch:
This one needs rebasing before we merge, there are some unrelated changes.

kewisch (changes requested):
Just moving to request changes so we don't miss the rebase

MelissaAutumn:
I've rebased for @markstos, but this commit log looks really weird (even before the rebase.)

MelissaAutumn:
@markstos can you create a new branch and cherry pick the feature commits there? Just ping me for a quick review to make sure nothing broke.

markstos:
Closing in favor of #967, a clean re-base.

PR #951, opened 2026-06-09
Simplify and wrap user plan activation in a db lock (Maybe fixes #838)

### LINKED ISSUE #838: Payment completed, Stalwart not updated
Lisa Wess is escalating this issue:

Ticket ID: 5631
Ticket subject: Subscription
Browser: Safari 26.4
OS: Mac OS X 10.15.7

Ticket includes PII, please see internal notes for links to Stalwart/DJ Admin and Paddle.

**Issue**

Customer is seeing, "Your payment is being processed.

This step can take up to 48 hours to complete."

Paddle shows payment complete and subscription in place. DJ Admin shows payment pending and no plan.

**Question**

1. Can I safely update the plan in DJ Admin, or are there other steps needed?
2. What went wrong?

gz#5631
gz#6119

gz#6125

MelissaAutumn:
Fixes #838 

So we were experiencing a race condition where a slower internet connection would resolve the paddle transaction callback url which sets is_awaiting_payment_verification = True within a second (one case was 80ms) as the Paddle webhook which activates their plan (sets plan_id for relationship) and sets is_awaiting_payment_verification = False.

In cases where the callback route completed last the user would have their plan activate and then it would wipe out any user model changes. So they would have a thundermail address created but no way to get the information to use it, and would be locked out of the dashboard.

The solution here is to wrap the user model update from the task in a db lock. This is just a postgres row lock, and  **SHOULD** pause any other queries that have to do with this model. I don't see any evidence we have to start wrapping everything in a db lock. 

We additionally now wrap the critical user update on paddle complete view with a db lock as well, but specify that if there **IS** a lock to skip the rest of the atomic transaction. This along with checking for the existence of a plan relationship will hopefully remove the race condition we were experiencing.

I think the other issue here besides the lack of a db lock for this critical update is that we were updating with a stale object. So a user object would be updated in celery, but after we fetched it on a web route. So when we updated the web route user, the changes would overwrite the more important updates done in celery. We can also work-around this a little by using update_fields=['is_awaiting_payment_verification'] but I'm hopeful that these changes are enough.

Tests use sqlite which doesn't support db row locking, I'm also not sure it's the right tool to test race conditions. 

Am I missing anything obvious? If so, please nit!

Useful readings:
https://docs.djangoproject.com/en/6.0/ref/models/querysets/#select-for-update

### DIFF
--- a/src/thunderbird_accounts/subscription/tasks.py
+++ b/src/thunderbird_accounts/subscription/tasks.py
@@ -313,6 +313,12 @@ def paddle_subscription_event(self, event_data: dict, occurred_at: datetime.date
             },
         )
 
+    # Log some problem cases that would be our fault. 
+    if len(subscription_items) == 0:
+        logging.error(f'Subscription contains no items. This user ({user.uuid}) will not have any features enabled!')
+    elif len(subscription_items) > 1:
+        logging.error('Subscription contains more than one item. This is not supported!')
+
     for item in subscription_items:
         # Note: These are reoccurring items only
         # Slurp up the subscription items, these are only created if all three items are new
@@ -353,8 +359,8 @@ def paddle_subscription_event(self, event_data: dict, occurred_at: datetime.date
             # Update quota
             plan = product_obj.plan
             if not plan:
-                logging.warning(f'Product {product.get("id")} has no plan attached!')
-            else:
+                logging.error(f'Product {product.get("id")} has no plan attached!')
+            elif status == Subscription.StatusValues.ACTIVE.value:
                 activate_subscription_features(user, plan)
 
         SubscriptionItem.objects.update_or_create(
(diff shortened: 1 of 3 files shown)


davinotdavid (approved):
This seems reasonable to me. I say we give it a try! lgtm!

markstos (commented):
Great job spotting the race condition.

The approach here makes sense to me.  I have experience deploying SELECT for UPDATE to resolve race conditions and this is a good approach.

Other attempts to update the same row (which should be rare) will block between the SELECT the UPDATE while other SELECT queries attempting to access the locked row will immediately get the old value returned until the transaction completes.

## QA Log

- Reviewed Python with docs
- Reviewed issue, PR description, diff
- Reviewed `atomic()` docs and `skip_locked` behavior.

Issue #896, opened 2026-06-02
Tag accounts that did not complete sign up process for MC automation


malini:
We have some users who have gone through the signup process enough to claim a @thundermail.com account name, but did not pay, so they are not using the service. We should autoremove these accounts within say 72 hours of creation.

--

EDIT: After discussions in Matrix, the idea of this task is to tag stale accounts for Mailchimp automations so that we can reach out before purging the account for good.

MelissaAutumn:
We have celery crons running now. We can run with on a daily occurance where an account less than 72 hours without a subscription or is_awaiting_payment_verification flag is purged. 

It's important to do this only for accounts without a subscription, this does not include accounts that have a lapsed subscription.

davinotdavid:
@kewisch since we are still discussing around the timing / emails around purging stale accounts, may I suggest us to split this ticket into two:

- Tag accounts that did not complete sign up process (Flight 3, already completed per https://github.com/thunderbird/thunderbird-accounts/pull/952)
- Purge stale accounts (Flight 4, currently blocked by https://github.com/thunderbird/thunderbird-accounts/pull/914)

Also, we should synchronize with the Marketing team (cc @inattee) on the timing of the `abandoned_cart` emails before purging the accounts.

kewisch:
Agree on this as discussed on Matrix.

PR #952, opened 2026-06-09
Create task to tag stale accounts for MC automations

davinotdavid:
## What changed?

- Refactored the existing `add_subscriber_to_mailchimp_list` task to extract reusable methods for Mailchimp tagging.
- Added a `tag_abandoned_cart_in_mailchimp` task that runs every hour which tags stale user accounts in Mailchimp with an `abandoned_cart` tag.

## Why?

It is important for the Marketing team to have a chance to run automations on users that might have dropped during the cart and before they actually get purged.

## Limitations and Notes

This should be merged in before the [purge accounts PR](https://github.com/thunderbird/thunderbird-accounts/pull/914) is merged since we would want users to be tagged before they get auto-purged.

## Applicable Issues

Related with https://github.com/thunderbird/thunderbird-accounts/issues/896

## Screenshots

image

### DIFF
--- a/src/thunderbird_accounts/authentication/tasks.py
+++ b/src/thunderbird_accounts/authentication/tasks.py
@@ -54,24 +54,21 @@ def tag_abandoned_cart_in_mailchimp(self):
             if user.subscription_set.exists():
                 skipped += 1
                 logger.info(
-                    'tag_abandoned_cart_in_mailchimp: skipped %s because a subscription exists',
-                    user.username,
+                    f'tag_abandoned_cart_in_mailchimp: skipped {user.uuid} because a subscription exists',
                 )
                 continue
 
             if user.is_awaiting_payment_verification:
                 skipped += 1
                 logger.info(
-                    'tag_abandoned_cart_in_mailchimp: skipped %s because payment verification is pending',
-                    user.username,
+                    f'tag_abandoned_cart_in_mailchimp: skipped {user.uuid} because payment verification is pending',
                 )
                 continue
 
             if not user.recovery_email:
                 skipped += 1
                 logger.info(
-                    'tag_abandoned_cart_in_mailchimp: skipped %s because recovery_email is blank',
-                    user.username,
+                    f'tag_abandoned_cart_in_mailchimp: skipped {user.uuid} because recovery_email is blank',
                 )
                 continue
 
@@ -85,14 +82,12 @@ def tag_abandoned_cart_in_mailchimp(self):
         except TaskFailed as ex:
             errors += 1
             logger.warning(
-                'tag_abandoned_cart_in_mailchimp: mailchimp error for %s: %s',
-                user.username,
-                ex.reason,
+                f'tag_abandoned_cart_in_mailchimp: mailchimp error for {user.uuid}: {ex.reason}',
             )
             continue
         except Exception as ex:
             errors += 1
-            logger.exception('tag_abandoned_cart_in_mailchimp: failed to tag %s: %s', user.username, ex.reason)
+            logger.exception(f'tag_abandoned_cart_in_mailchimp: failed to tag {user.uuid}: {ex.reason}')
             continue
         else:
             tagged += 1
--- a/src/thunderbird_accounts/authentication/tasks.py
+++ b/src/thunderbird_accounts/authentication/tasks.py
@@ -1,0 +1,107 @@
+import logging
+from datetime import timedelta
+
+from celery import shared_task
+from django.conf import settings
+from django.db.models import Exists, OuterRef, QuerySet
+from django.utils import timezone
+
+from thunderbird_accounts.authentication.models import User
+from thunderbird_accounts.celery.exceptions import TaskFailed
+from thunderbird_accounts.subscription.models import Subscription
+from thunderbird_accounts.subscription.tasks import add_or_tag_mailchimp_member
+
+logger = logging.getLogger(__name__)
+
+
+def get_abandoned_cart_users_for_mailchimp_tag() -> QuerySet[User]:
+    """Incomplete sign-ups old enough to receive the abandoned_cart Mailchimp tag."""
+    cutoff = timezone.now() - timedelta(hours=settings.ABANDONED_CART_TAG_HOURS)
+    has_subscription = Subscription.objects.filter(user_id=OuterRef('pk'))
+    return (
+        User.objects.filter(
+            created_at__lt=cutoff,
+            is_test_account=False,
+            is_staff=False,
+            is_superuser=False,
+            is_awaiting_payment_verification=False,
+        )
+        .annotate(has_subscription=Exists(has_subscription))
+        .filter(has_subscription=False)
+    )
+
+
+@shared_task(bind=True)
+def tag_abandoned_cart_in_mailchimp(self):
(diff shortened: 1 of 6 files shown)


src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+    mailchimp_language_map = settings.ACCOUNTS_TO_MAILCHIMP_LANGUAGES
+    mailchimp_tag = settings.ABANDONED_CART_MAILCHIMP_TAG
+
+    for user in get_abandoned_cart_users_for_mailchimp_tag().iterator():
+        try:
+            if user.subscription_set.exists():
+                skipped += 1
+                logger.info(
+                    'tag_abandoned_cart_in_mailchimp: skipped %s because a subscription exists',
+                    user.username,
```
MelissaAutumn: Can we swap these to user.uuid, alao you can use fstringa to inline the variable.

src/thunderbird_accounts/subscription/tasks.py
```diff
@@ -491,14 +491,112 @@ def update_thundermail_quota(self, plan_uuid):
     }
 
 
+def _mailchimp_basic_auth() -> str:
+    return base64.b64encode(f'this_does_not_support_bearer_auth:{settings.MAILCHIMP_API_KEY}'.encode()).decode()
```
MelissaAutumn, line 496: I'm glad we used that as the username here, because I forgot and was about to ask why aren't we using bearer auth haha.

src/thunderbird_accounts/subscription/tasks.py
```diff
@@ -491,14 +491,112 @@ def update_thundermail_quota(self, plan_uuid):
...
+
+
+def _mailchimp_api_query(
+    method: str, api_endpoint: str, basic_auth: str, _json: dict | None = None
```
MelissaAutumn: Nit: json is a stringified format, maybe data is the better param name here

src/thunderbird_accounts/subscription/tasks.py
```diff
@@ -491,14 +491,112 @@ def update_thundermail_quota(self, plan_uuid):
...
+        # in case the next request fails.
+
+        # Error details reference: https://mailchimp.com/developer/marketing/docs/errors/#error-glossary
+        try:
+            error_details = ex.response.json()
+        except (JSONDecodeError, AttributeError):
+            error_details = {}
+
+        # Send some extra information to sentry
+        sentry_sdk.set_extra('tag_error_details', error_details)
```
MelissaAutumn: can you swap this over to set_context? set extra is deprecated. set context takes a dict as the second arg.

src/thunderbird_accounts/settings.py
```diff
@@ -600,6 +601,15 @@ def oidc_logout(request):
 MAILCHIMP_LIST_ID = os.getenv('MAILCHIMP_LIST_ID')
 USE_MAILCHIMP = bool(MAILCHIMP_API_KEY)  # If we don't have an api key disable mailchimp
 
+ABANDONED_CART_TAG_HOURS = int(os.getenv('ABANDONED_CART_TAG_HOURS', '1'))
```
MelissaAutumn, line 609: Can we set the pulumi var for 1 week? Or did we settle on a different value?

MelissaAutumn (changes requested):
Thanks! I've got some nits though.

src/thunderbird_accounts/settings.py
davinotdavid: Ah this one is for the abandoned cart tag, not for the purge! @inattee says: `Best practice for cart recovery email is to usually send them fairly close (1-4 hours, max 24 hours), however I would say we can start by passing the tag to Mailchimp within 1 hour of abandonment`.

[The purge in the other PR](https://github.com/thunderbird/thunderbird-accounts/pull/914) will happen in 7 days though!

davinotdavid:
@MelissaAutumn thanks for the review! Addressed the nits and rebased to solve conflicts.

src/thunderbird_accounts/authentication/tests/test_tasks.py
```diff
@@ -0,0 +1,159 @@
...
+    tag_abandoned_cart_in_mailchimp,
+)
+from thunderbird_accounts.celery.exceptions import TaskFailed
+from thunderbird_accounts.subscription.models import Subscription
+
+
+class GetAbandonedCartUsersForMailchimpTagTestCase(TestCase):
+    def setUp(self):
+        self.subdomain = settings.PRIMARY_EMAIL_DOMAIN
+        self.eligible_created_at = timezone.now() - timedelta(hours=settings.ABANDONED_CART_TAG_HOURS + 1)
```
MelissaAutumn, line 25: You can also use freezegun for freezing datetimes in a specific place. Might be useful for doing a DST check.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -0,0 +1,102 @@
...
+logger = logging.getLogger(__name__)
+
+
+def get_abandoned_cart_users_for_mailchimp_tag() -> QuerySet[User]:
+    """Incomplete sign-ups old enough to receive the abandoned_cart Mailchimp tag."""
+    cutoff = timezone.now() - timedelta(hours=settings.ABANDONED_CART_TAG_HOURS)
+    has_subscription = Subscription.objects.filter(user_id=OuterRef('pk'))
+    return (
+        User.objects.filter(
+            created_at__lt=cutoff,
```
MelissaAutumn, line 23: The legal accept record might be more accurate, but I think this should be fine.

src/thunderbird_accounts/subscription/tasks.py
```diff
@@ -492,14 +492,112 @@ def update_thundermail_quota(self, plan_uuid):
     }
 
 
+def _mailchimp_basic_auth() -> str:
+    return base64.b64encode(f'this_does_not_support_bearer_auth:{settings.MAILCHIMP_API_KEY}'.encode()).decode()
+
+
+def _mailchimp_api_query(
```
MelissaAutumn, line 499: If we do further work with Mailchimp we should split it into a client (see clients.py in some of the other django apps for an example.)

MelissaAutumn (approved):
This was supposed to be an approval...

src/thunderbird_accounts/authentication/tasks.py
davinotdavid: Hmm we could use the legal accept record yes but that's a good question on whether or not we want to capture those who dropped before accepting the terms 🤔 I'd say we can go broader for now and include those. Nice to keep in mind in case there's an ask to not do that though, thanks!

src/thunderbird_accounts/authentication/tests/test_tasks.py
davinotdavid: Ah yes, I've added a new test for a DST check + using freezegun now, thanks!

src/thunderbird_accounts/subscription/tasks.py
davinotdavid: Good point, I've cut a ticket for it here https://github.com/thunderbird/thunderbird-accounts/issues/956. Thanks!

PR #953, opened 2026-06-09
Refresh the user instance after we perform the db row lock during sub activation.

MelissaAutumn:
One tiny regression, not harmful just kinda gives folks unlimited mail quota (by skipping the part that sets it...)

I've added a test (sure could use #152 :sweat_smile:) that would fail before this change.


davinotdavid (approved):
Ah, sorry for not catching this in the previous review. Without refreshing the user reference, https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/mail/utils.py#L89-L90 would fail to assign the mail quota.

Nice!

MelissaAutumn:
No worries, I also missed it and I wrote the code lol! Thanks!

Issue #944, opened 2026-06-05
Implement re-design of Custom Domain flow


davinotdavid:
We recently received new designs for the Custom Domain flow to better improve the UX of it.

Zeplin: https://zpl.io/xnKorZW

davinotdavid:
@kewisch after syncing with @solangevalverde on this, she mentioned that the proposed Zeplin changes are an intermediate state to solve our current needs but further exploration would be necessary for a more complete solve for this.

After checking the flows in other mail providers, there are opportunities for better / clearer steps during the custom domain setup but that would be a much bigger change that would require Design time for exploration (depending on the priority, might require sync with @laurelterlesky as well).

kewisch:
Make sense. Ultimately I'm interested in improving clarity and resolving the items in #920, which includes [this comment](https://github.com/thunderbird/thunderbird-accounts/issues/920#issuecomment-4625817090) that would be solved by the UX changes. I'm happy to follow UX' guidance on which state to implement when.

davinotdavid:
@kewisch as requested in Matrix, here's the list of DNS Records we provide and their previous/proposed status:

| Record  | Before | Proposed |
| -- | -- | -- |
| MX | **Critical** | Warning |
| SRV | Warning  | Warning |
| TXT | Warning  | Warning |
| CNAME (DKIM) | **Critical**  | Warning |

Note that the idea from the Design team (cc @solangevalverde) was to normalize all as warnings so that ideally the user would fix them all instead of only fixing what is critical as that would increase the chances of Thundermail working as expected.

Also, the actions that should be taken per record vary with the introduction of a badge and an info icon with a tooltip per conflicting line.

kewisch:
@solangevalverde Regarding the levels, I'm not convinced we should keep them at the same level. Here are two use cases:
* User does not want to make use of our caldav/carddav server and would like another host to win the battle for autoconfig
* Some folks may not want to have a TLSRPT or mta-sts entry, to avoid the reports that they can't really take action on

I'm not so much worried about those who are just setting up their domain (they'll eventually fix all the warnings if something isn't working), but more so about the power users that would like a custom setup and our checks might force them to give it up.

davinotdavid:
@kewisch Chatted with @solangevalverde today and here's the takeaways from our chat:

[...]

PR #955, opened 2026-06-11
Custom domain UI revamp

davinotdavid:
## What changed?

- Refactored the `CustomDomainForm` component into more focused / granular components for better future maintainability:
  - `CustomDomainAddStep` and `CustomDomainVerifyStep` only take care of their respective steps
  -  `DnsRecordsTable` only cares about the table itself
  -  `utils.ts` has all the non-component-specific functions used throughout the components
- The `DnsRecordsTable` now uses grid and subgrid for better / easier layouts
- Removed the `critical` status handling in favor of keeping them all as `warning`, per Design request
- Added a header showcasing which domain is the one you are seeing the records of
- Added copy buttons when there are values to be copied 
- Added `Conflicting Records` / `All Records` tabs to filter through the table
- Added info tooltips for conflicting records with personalized tooltips per record type / action

## Why?

Even though we recently updated the Custom Domain UI, it was a temporary measure to unblock the Flight 2. There were still more design updates that the Design team wanted to do that we didn't have time to address so now that we've got a more polished version of it, we should implement it!

## Limitations and Notes

- We are missing the support article to be added in the MX record tooltips. There's already a [Notion ticket requesting it](https://app.notion.com/p/mzthunderbird/New-Article-TB-Pro-Domains-having-more-than-one-MX-record-3742df5d45ae806584afdfff3fe88cf4?source=copy_link&assetsVersion=23.13.20260612.1035) and it has to be updated soon. Currently this is hard-coded to point to the support website homepage.

- There's no place yet to showcase error in case the clipboard is not available for copying the button. We are only doing a console.error for now but this will probably change in the near future. There's a TODO in the code for that!

- This design is a step towards the north star of custom domains but not the definitive version yet. The Design team is currently evaluating if/when they will get time to do further work around this.

## Applicable Issues

Related with https://github.com/thunderbird/thunderbird-accounts/issues/944

## Screenshots

<details>
<summary>View DNS records only (no verification yet)</summary>
image
</details>

<details>
<summary>Failed (various conflicting records, all records tab)</summary>
image
</details>

<details>
<summary>Failed (various conflicting records, conflicting records tab)</summary>
image
</details>

<details>
<summary>Verified (check marks on the left, no tabs are displayed and no action column)</summary>
image
</details>

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -236,6 +236,7 @@
           "recordConflictWarning": "Conflict detected. Replace {currentValue} with the value in the row above.",
           "mxRecordConflictWarning": "This record conflicts. Remove conflicting MX records or change the priority.",
           "spfRecordConflictWarning": "This record conflicts. The current SPF record does not contain {includeValue}.",
+          "spfRecordIncludeValue": "the expected include value",
           "recordMissingWarning": "Record is missing.",
           "staleDnsRecordsTitle": "Remove these stale records",
           "staleDnsRecords": "Please remove the {type} record for {name} as it points to {existing_values}.",
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -253,6 +253,53 @@
           "domainAlreadyExists": "Domain already exists",
           "domainAlreadyConfigured": "This domain is already configured. Please {link} in case you believe this is a mistake.",
           "reachOutToSupport": "reach out to support",
+          "recordActionInfo": {
+            "MX": {
+              "remove": {
+                "remove": "Remove",
+                "content": "{bold} this record from your DNS register site to avoid having email sent to multiple servers. {link}",
+                "link": "Learn more on our support site."
+              },
+              "add": {
+                "add": "Add",
+                "content": "{bold} this record to your DNS register site to have your email sent to Thunderbird Pro servers. {link}",
+                "link": "Learn more on our support site."
+              },
+              "edit": {
+                "edit": "Edit",
+                "content": "{bold} this record from your DNS register site to avoid having email sent to multiple servers. {link}",
+                "link": "Learn more on our support site."
+              }
+            },
+            "SRV": {
+              "remove": {
+                "remove": "Remove",
+                "content": "{bold} this record from your DNS register site to avoid compatibility issues."
+              },
+              "add": {
+                "add": "Add",
+                "content": "{bold} this record to your DNS register site to avoid compatibility issues."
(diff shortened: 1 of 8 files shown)


markstos (changes requested):
Overall, it looks like the spec. Nice!

- The Edit/Remove/Add labels look like they might be buttons rather than tags, but the 
  "i" tooltip next to them may be sufficient to clarify.

- For the SRV "Edit" case, it's not clear from the tooltip what edit is required. This was
  clearer before when we showed both the ideal and conflicting rows.

- When testing with the "noshmind" domain, I got an "Edit" error for
  "_caldavs._tcp.noshmind.com.", but when I compare that to the recommended DNS record, I don't
  see what the difference is. This appears to be a possible backend bug, as the backend is
  reporting a conflict. <Thinks Harder>. OK, I believe I understand this issue, because it's also
  happening with the MX record. When thre's a conflicting row, like an extra MX record at the
  same priority or a extra, conflicting SRV record, then we are displaying "Edit" next to *correct*
  rows that don't need an edit, they only need the conflicting record removed. In terms of UX,
  I believe the correct rows should NOT have the "Edit" label next to them.

- Perhaps we should clarify what "All Records" means. We are not fetching all the DNS records for their zone.
  (And for technical records, we can't). A better name is "Records to Add"

- I thought we were going to do "Yellow for Warning (Optional)" and Red for Required (like before)
  But when tested with superduperfakefakefake.com, even the missing MX record was marked in
  yellow, not red.

- See also the related issue about considering the missing DKIM CNAMES as errors that we are trying
  to land: https://github.com/thunderbird/thunderbird-accounts/pull/947 That other PR will likely
  land first, so maybe there is nothing to do here.

solangevalverde:
> Perhaps we should clarify what "All Records" means. We are not fetching all the DNS records for their zone.
(And for technical records, we can't). A better name is "Records to Add"

As [I mentioned on the zeplin file](https://zpl.io/PrLopEm), the design has decided for the time being we will avoid differentiating warning and conflict as even non critical conflicts can in the long run create conflict for users. If due to landing mergers and potential deadlines, #947 can land first and then this ticket should be edited to fix the multi-color conflict rows.

Incoming changes from design: copy from the tooltips as well as copy on the tags so they don't seem like actionable objects but a task needed to be done without the tag being interacted with. I will comment here once Zeplin has been updated with the copy change.

davinotdavid:
Moving to draft per latest conversations with Design about this: https://github.com/thunderbird/thunderbird-accounts/issues/944#issuecomment-4792956269

PR #965, opened 2026-06-15
Upgrade the container to trixie, and upgrade Vite6 -> Vite8 (Fixes #963)

### LINKED ISSUE #963: Upgrade to Vite 8
Due to https://github.com/thunderbird/thunderbird-accounts/pull/962 we should really stay on top of updates. 

Probably a relatively easy upgrade but I haven't really looked through the code yet.
https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md

MelissaAutumn:
Fixes #963

Also bump version to v1.14.1.

This should resolve https://github.com/thunderbird/thunderbird-accounts/security/dependabot/121 (and also close https://github.com/thunderbird/thunderbird-accounts/pull/962)

We upgrade the container to trixie since bookworm only has node 18, and vite 7 dropped node 18 support. Locally I use node 22, trixie has node 20. 

Otherwise it all just works as intended on my machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "thunderbird-accounts",
-  "version": "1.14.0",
+  "version": "1.14.1",
   "description": "> [!IMPORTANT]   > Thunderbird Accounts is still in active development and is not ready for any active use.",
   "private": true,
   "directories": {
@@ -39,7 +39,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.33.0",
-    "@vitejs/plugin-vue": "^5.2.3",
+    "@vitejs/plugin-vue": "^6.0.7",
     "@vue/eslint-config-typescript": "^14.6.0",
     "eslint": "^9.33.0",
     "eslint-import-resolver-typescript": "^4.4.4",
@@ -49,7 +49,7 @@
     "jiti": "^2.5.1",
     "prettier": "3.6.2",
     "typescript-eslint": "^8.39.1",
-    "vite": "^6.2.7",
-    "vitest": "^4.1.6"
+    "vite": "^8.0.16",
+    "vitest": "^4.1.8"
   }
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13-bookworm
+FROM python:3.13-trixie
 #
 # Development docker file
 #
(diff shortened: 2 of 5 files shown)


davinotdavid (approved):
lgtm! tested locally so it works on 2 machines now

package.json
```diff
@@ -49,7 +49,7 @@
     "jiti": "^2.5.1",
     "prettier": "3.6.2",
     "typescript-eslint": "^8.39.1",
-    "vite": "^6.2.7",
-    "vitest": "^4.1.6"
+    "vite": "^8.0.16",
```
davinotdavid, line 52: Hm, I merged a [dependabot PR in Appointment](https://github.com/thunderbird/appointment/pull/1705) that didn't seem to have solved the vulnerability / esbuild peer dependency as it still installed the vulnerable esbuild version.

I've added an `overrides` here https://github.com/thunderbird/appointment/pull/1707/changes#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R63-R64

MelissaAutumn: Vite 8 doesn't use esbuild iirc. That's why this fix was recommended here. 

It's possible appointment has another dep that uses it, but this PR clears up my npm audit check. 

Will double check though.

davinotdavid: Hmm they still have it as a peer dependency though? This is the `8.0.4` changelog https://github.com/vitejs/vite/blob/v8.0.16/packages/vite/CHANGELOG.md#804-2026-04-06 that seemed to be the culprit in the Appointment case and this is the PR that did it https://github.com/vitejs/vite/pull/22155/changes

MelissaAutumn: Ah, but esbuild appears to be optional. Either way our package-lock doesn't pull esbuild in, instead it pulls rolldown which is the new build _thingy_. (https://vite.dev/blog/announcing-vite8#the-problem)

I'm not sure why Appointment pulls in esbuild as it also pulls in rolldown (https://github.com/thunderbird/appointment/blob/main/frontend/package-lock.json#L8953)

davinotdavid: Oh you're right, I missed that it is optional, thanks! Seems like Appointment was pulling it by mistake as a fresh `npm i` after clearing everything doesn't seem to be pulling it anymore.

PR #967, opened 2026-06-15
Update DMARC validation to pass any record that starts with V=DMARC1 (Take 2) (#941)

markstos:
Supercedes #949 with a clean rebase.

No additional changes, only cherry-picking.

(I also tried rebasing #949 and somehow that didn't clean up the commit
history either).

### DIFF
--- a/src/thunderbird_accounts/mail/dns.py
+++ b/src/thunderbird_accounts/mail/dns.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from typing import Optional
 
 import dns.rdatatype as dns_rdatatype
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/mail/dns.py
```diff
@@ -139,6 +146,58 @@ def _compare_dkim_txt(expected_content: str, live_values: list[str]) -> tuple[DN
     return DNSRecordStatus.CONFLICT, dkim_values
 
 
+def _parse_txt_tag_value_list(content: str) -> Optional[list[tuple[str, str]]]:
+    parts = content.strip().split(';')
+    # Remove empty segments.
+    if parts and not parts[-1].strip():
+        parts.pop()
```
MelissaAutumn, line 153: Maybe like a `parts = list(filter(None, parts))`? This works if we only want to remove the last item though.

MelissaAutumn (approved):
Thanks, one suggestion but @kewisch already reviewed this so it should be good to go.

src/thunderbird_accounts/mail/dns.py
markstos: I prefer the compactness and simplicity of your proposal but upon checking, it would cause some invalid cases to be considered valid, which is an overriding consideration.

RFC 6376 Section 3.2 defines the grammar we are parsing:

https://www.rfc-editor.org/info/rfc6376/#section-3.2

In that, a double-semicolon should be an error, but the simple one-liner would transform it into valid code.

```
>>> content = 'v=DMARC;; p=none'
>>> parts = content.strip().split(';');
>>> list(filter(None, parts));
['v=DMARC', ' p=none']
```

Leaving the code as-is in more in-line with the goal of this issue to do stricter DMARC parsing.

PR #968, opened 2026-06-15
Perform in-depth validation on autodiscover subdomains (#943)

### LINKED ISSUE #943: Figure out how to handle wildcard entries like *.example.com , which also includes autodiscover.example.com
Wildcard DNS can cause autoconfig.customer.com to resolve. There's no clear way for the backend to be sure whether the resolution happened directly or by wildcard.

For additional context: The `autodiscover` DNS record is primarily implemented by Outlook and we don't have current plans to support it.

## The workaround we need to communicate to users

Because this is an edge case, it's probably best to communicate with a simple error message in the UI with a link to a KB article. The full solution we need to communicate is this:

If the `autodiscover` subdomain is resolving but doesn't actually exist, this can happen because because you have wildcard DNS enabled for subdomains. In this case, adding a  TXT "tombstone" record can be added to block wildcard resolution. Add this record:

Type: TXT
Host: autodiscover.yourdomain.com
Value: "no autodiscover service"

markstos:
## What changed?

Here in the event that an `autodiscover` subdomain is present, we perform additional HTTP-based checks to see if it's actually an Exchange server that would be problematic.

Because the user controls the domain that we are making HTTP and DNS requests do, some care is taken to avoid server-side request-forgery vulns. (SSRF).

The detection approach is based some on what Thunderbird Desktop does.

## Why?

Wildcard DNS is common, which can currently result in a false report of a problematic `autodiscover` subdomain.

## QA Log

- Automated testing 

## Limitations and Notes

As described in #943, a linked KB article is recommended to clarify the edge case a "tombstone" DNS TXT record is a necessary fix IF they 1. Use wildcard DNS
2. Their `autodiscover` subdomain manages to sufficiently appear to us to an Exchange server
without actually being one.

Though, maybe that case is so rare we shouldn't bother with it.

## Applicable Issues

Closes #943

### DIFF
--- a/src/thunderbird_accounts/mail/autodiscover_probe.py
+++ b/src/thunderbird_accounts/mail/autodiscover_probe.py
@@ -135,7 +135,18 @@ def _looks_like_exchange(response: requests.Response, body: bytes) -> bool:
 
 
 def exchange_autodiscover_endpoint_exists(hostname: str, domain_name: str) -> bool:
-    """Probe a host for a functional Exchange Autodiscover endpoint."""
+    """
+    Probe a host to see if it looks like an Exchange Autodiscover endpoint.
+
+    Resolve the DNS name to confirm it's public IP, not private.
+    To avoid a race condition with resolving the domain a second time,
+    we re-use the IP in the connection string and send the DNS name in the host header.
+
+    After allowing AUTODISCOVER_PROBE_MAX_REDIRECTS, we check the headers for an
+    Exchange-like header.
+
+    Finally, we'll check to see if the response body looks like Exchange.
+    """
     body = _autodiscover_probe_body(domain_name).encode()
     session = requests.Session()
     session.trust_env = False
--- a/src/thunderbird_accounts/mail/autodiscover_probe.py
+++ b/src/thunderbird_accounts/mail/autodiscover_probe.py
@@ -23,6 +23,7 @@ AUTODISCOVER_PROBE_HEADERS = {
     'Content-Type': 'text/xml; charset=utf-8',
 }
 AUTODISCOVER_EXCHANGE_HEADER_NAMES = {
+    'x-aspnet-version',
     'x-beserver',
     'x-calculatedbetarget',
     'x-diaginfo',
--- a/src/thunderbird_accounts/mail/autodiscover_probe.py
+++ b/src/thunderbird_accounts/mail/autodiscover_probe.py
@@ -3,12 +3,16 @@ Code for probing if a custom domain's autodiscover subdomain belongs to an Excha
 
 If so, that's a conflict to flag to the user.
 """
+
+from dataclasses import dataclass
 import ipaddress
 import logging
 import socket
 from urllib.parse import urljoin, urlparse
 
 import requests
+from requests_toolbelt.adapters import host_header_ssl
+
 AUTODISCOVER_PROBE_PATH = '/autodiscover/autodiscover.xml'
 AUTODISCOVER_PROBE_MAX_REDIRECTS = 3
 AUTODISCOVER_PROBE_MAX_BYTES = 65536
@@ -29,46 +33,74 @@ AUTODISCOVER_EXCHANGE_HEADER_NAMES = {
 AUTODISCOVER_EXCHANGE_TERMS = ('autodiscover', 'exchange', 'office365', 'outlook')
 
 
+@dataclass(frozen=True)
+class _AutodiscoverRequestTarget:
+    request_url: str
+    hostname: str
+
+
 def _autodiscover_probe_body(domain_name: str) -> str:
     """Build the minimal Exchange Autodiscover XML request body."""
     email_address = f'autodiscover-probe@{domain_name}'
-    return f'''<?xml version="1.0" encoding="utf-8"?>
+    return f"""<?xml version="1.0" encoding="utf-8"?>
 <Autodiscover xmlns="http://schemas.microsoft.com/exchange/autodiscover/outlook/requestschema/2006">
   <Request>
     <EMailAddress>{email_address}</EMailAddress>
     <AcceptableResponseSchema>http://schemas.microsoft.com/exchange/autodiscover/outlook/responseschema/2006a</AcceptableResponseSchema>
   </Request>
-</Autodiscover>'''
+</Autodiscover>"""
 
 
-def _is_safe_autodiscover_url(url: str) -> bool:
+def _format_ip_for_url(address: ipaddress.IPv4Address | ipaddress.IPv6Address) -> str:
+    if address.version == 6:
+        return f'[{address}]'
+    return str(address)
+
+
+def _safe_autodiscover_request_targets(url: str) -> list[_AutodiscoverRequestTarget]:
     """
     Because we are making HTTP + DNS requests to user-controlled domain,
-    Make various security checks for the actual HTTP request.
+    resolve and validate hostnames before connecting to the vetted IP address.
+    We connect via HTTP to an IP and send the the Host header to avoid a race condition
+    between the initial DNS resolution to check the validate the IP address and
+    the second DNS resolution to send the HTTP probe, as the DNS resolution could
+    potentially change during that window.
     """
     try:
         parsed = urlparse(url)
         if parsed.scheme not in {'http', 'https'} or parsed.port not in {None, 80, 443}:
-            return False
+            return []
         if not parsed.hostname or parsed.username or parsed.password:
-            return False
+            return []
         try:
             ipaddress.ip_address(parsed.hostname)
-            return False
+            return []
         except ValueError:
             pass
 
         port = parsed.port or (443 if parsed.scheme == 'https' else 80)
-        # SSRF check: resolve customer-controlled hostnames before requesting them so they cannot
-        # send backend HTTP traffic to private, loopback, link-local, or metadata addresses.
+        addresses = []
         for address_info in socket.getaddrinfo(parsed.hostname, port, type=socket.SOCK_STREAM):
             address = ipaddress.ip_address(address_info[4][0])
             if not address.is_global:
-                return False
+                return []
+            addresses.append(address)
     except (OSError, ValueError):
-        return False
+        return []
 
-    return True
+    path = parsed.path or '/'
+    if parsed.query:
+        path = f'{path}?{parsed.query}'
+
+    targets = []
+    seen_addresses = set()
+    for address in addresses:
+        if address in seen_addresses:
+            continue
+        seen_addresses.add(address)
+        request_url = f'{parsed.scheme}://{_format_ip_for_url(address)}:{port}{path}'
+        targets.append(_AutodiscoverRequestTarget(request_url=request_url, hostname=parsed.hostname))
+    return targets
 
 
 def _body_mentions_autodiscover(body: bytes) -> bool:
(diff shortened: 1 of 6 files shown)


mzeier (commented):
- **SSRF guard is bypassable (TOCTOU).** `_is_safe_autodiscover_url` resolves the host and checks `is_global`, but `requests.post` then re-resolves on its own, so the IP we validated isn't the one we connect to. Since the customer owns the DNS for `autodiscover.<domain>`, a low-TTL record can pass validation as public and resolve to `127.0.0.1`/`10.x`/`169.254.169.254` at request time. Should pin the connection to the validated IP (connect by IP, send hostname in `Host`, re-check after each redirect).

mzeier (commented):
- **`200` match is just "body contains `autodiscover`".** We POST to `/autodiscover/autodiscover.xml`, so any soft-404 page that echoes the path matches.

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -0,0 +1,141 @@
...
+        return _has_exchange_autodiscover_evidence(response, body)
+    return False
+
+
+def exchange_autodiscover_endpoint_exists(hostname: str, domain_name: str) -> bool:
+    """Probe a host for a functional Exchange Autodiscover endpoint."""
+    body = _autodiscover_probe_body(domain_name).encode()
+
+    for scheme in ('https', 'http'):
+        url = f'{scheme}://{hostname}{AUTODISCOVER_PROBE_PATH}'
```
eleanxr, line 156: Thunderbird also looks for `http[s]://autodiscover.<hostname>/autodiscover/autodiscover.xml`.

See https://searchfox.org/comm-central/rev/72b8ba0761b3881d926be53f77fbf75d5a9316d5/mail/components/accountcreation/modules/ExchangeAutoDiscover.sys.mjs#231

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -0,0 +1,141 @@
...
+    # outlook.com needs this exact string, with space and lower case "utf".
+    'Content-Type': 'text/xml; charset=utf-8',
+}
+AUTODISCOVER_EXCHANGE_HEADER_NAMES = {
+    'x-beserver',
+    'x-calculatedbetarget',
+    'x-diaginfo',
+    'x-feserver',
+    'x-msdiagnostics',
+    'x-owa-version',
```
eleanxr, line 33: In our exchange on premises test account with autodiscover, I'm also seeing x-aspnet-version as something that looks like a MS specific header. I'm not seeing x-owa-version or x-msdiagnostics, but those definitely seem like a plausible exchange headers.

I also see x-powered-by: ASP.NET in our test accounts Exchange autodiscover endpoint.

mzeier:
**Reproducing #943 with a wildcard A record**

The DNS side is already enough to trigger it — a wildcard `*.noshmind.com` A record means `autodiscover.noshmind.com` resolves even though no autodiscover record was ever created. The reason it may look like "nothing happens" is that `check_stale_dns_records` only runs inside the custom-domain verification flow (`views.py:266`), so calling it directly is the fastest way to see the two branches diverge.

**Steps**

1. Confirm the wildcard is live (a random label resolves to the same IP as `autodiscover`):
   ```sh
   dig +short A autodiscover.noshmind.com
   dig +short A foobar123.noshmind.com   # same IP => wildcard
   ```

2. On **main**, in a Django shell:
   ```python
   from thunderbird_accounts.mail.dns import check_stale_dns_records
   check_stale_dns_records('noshmind.com')
   ```
   Expected (false positive):
   ```python
   [{'code': 'autodiscoverCnameUnexpected', 'type': 'A',
     'name': 'autodiscover.noshmind.com', 'existing_values': ['103.168.73.37']}]
   ```

3. On the **PR branch (#968)**, run the same call:
   ```python
   check_stale_dns_records('noshmind.com')
   ```
   Expected:
   ```python
   []
   ```
   The PR probes `https://autodiscover.noshmind.com/autodiscover/autodiscover.xml`, sees it doesn't respond like an Exchange Autodiscover endpoint, and suppresses the record.

davinotdavid (commented):
Overall, code LGTM but other folks would be much better at reviewing this as I lack the subject matter expertise.

I did run this locally (thanks @mzeier !) and can verify that it solves the original problem stated in the issue:

Before (main)

Image

After (this branch), no autodiscover records are displayed

Image

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -0,0 +1,141 @@
...
+                break
+
+            try:
+                response = requests.post(
+                    url,
+                    data=body,
+                    headers=AUTODISCOVER_PROBE_HEADERS,
+                    timeout=AUTODISCOVER_PROBE_TIMEOUT,
+                    allow_redirects=False,
+                    stream=True,
```
MelissaAutumn: I don't know exchange, any benefit from streaming it vs accepting it as-is and decoding the content?

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -0,0 +1,141 @@
...
+    if response.status_code == 401:
+        return True
+    if 200 <= response.status_code < 300:
+        return _body_mentions_autodiscover(body)
+    if response.status_code in {400, 403, 405}:
+        return _has_exchange_autodiscover_evidence(response, body)
+    return False
+
+
+def exchange_autodiscover_endpoint_exists(hostname: str, domain_name: str) -> bool:
```
MelissaAutumn, line 137: Can you expand on what's checked in the function description? If I'm using this function I'd like to see what it's going to do.

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -0,0 +1,141 @@
...
+    'x-calculatedbetarget',
+    'x-diaginfo',
+    'x-feserver',
+    'x-msdiagnostics',
+    'x-owa-version',
+}
+AUTODISCOVER_EXCHANGE_TERMS = ('autodiscover', 'exchange', 'office365', 'outlook')
+
+
+def _autodiscover_probe_body(domain_name: str) -> str:
```
MelissaAutumn, line 44: nit: This can be cached with functools.

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -0,0 +1,141 @@
...
+            address = ipaddress.ip_address(address_info[4][0])
+            if not address.is_global:
+                return False
+    except (OSError, ValueError):
+        return False
+
+    return True
+
+
+def _body_mentions_autodiscover(body: bytes) -> bool:
```
MelissaAutumn, line 108: nit: this can be cached with functools.

Also any reason why body is not decoded at this point? Does exchange return raw bytes or are we just parsing partially loaded xml?

MelissaAutumn (changes requested):
In the future I guess since we do want to get this out soon, I'd like to see this be stuffed into a class. I think a server probe interface would be a useful thing to have where it ingests a url, maybe optionally some credentials and probes a server which is handled by whatever impl you're using. So in this case the IsExchangeProbe / ExchangeAutodiscoverProbe would have all of this code implemented in it, but with one or two public functions to call.

src/thunderbird_accounts/mail/autodiscover_probe.py
MelissaAutumn: (This is my main blocker btw, nits are nitpicks.)

markstos:
> * **SSRF guard is bypassable (TOCTOU).** 

Good catch. I've changed the approach to make the time-of-use connection use the same IP validated at Time-of-Check, closing the gap. Commit pushed.

markstos:
> * **`200` match is just "body contains `autodiscover`".** We POST to `/autodiscover/autodiscover.xml`, so any soft-404 page that echoes the path matches.

During coding, AI also recommended to me to do formal XML parsing of body instead of a simple substring match. The simple string match was implemented instead because adding XML parsing feels rather far off into the weeds from the goal of validating if an `autodiscover` subdomain is going to be a problematic. To trigger the case where substring-check vs XML-parsing made a difference, the user would need:

1. Wildcard DNS
2. `autodiscover` ends up resolving to domain listening over HTTP. 
3. Web server is responding at `/autodiscover/autodiscover.xml` with a 200 response code even though it's not an Exchange server.
4. Body contains `autodiscover` because it also has "soft 404" behavior that echoes back.

If a user manages to trigger this rare case, the plan is link them to a KB article which explains they can add a specific TXT record which will resolve our validation, and may well help fend off other scanners that also believe they have an Exchange server.

src/thunderbird_accounts/mail/autodiscover_probe.py
```diff
@@ -103,24 +135,34 @@ def _looks_like_exchange(response: requests.Response, body: bytes) -> bool:
 def exchange_autodiscover_endpoint_exists(hostname: str, domain_name: str) -> bool:
     """Probe a host for a functional Exchange Autodiscover endpoint."""
     body = _autodiscover_probe_body(domain_name).encode()
+    session = requests.Session()
+    session.trust_env = False
+    session.mount('https://', host_header_ssl.HostHeaderSSLAdapter())
```
MelissaAutumn, line 153: Hmm honestly for a dep that hasn't been updated since 2023 I'd maybe just pull this class directly into accounts instead of adding the dep.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: This was the file I looked at during development, and we also probing the same URL structure.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: Thanks for the hints. I've added `x-aspnet-version` to my list.  

Since we we are currently only checking key names and not values, I'm passing on the `x-powered-by` header for now, but may revisit that later if we need deeper checks.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: The benefits of `stream=True` here are both performance and defensive. Working with `AUTODISCOVER_PROBE_MAX_BYTES`, we only download the first 64 bytes and look for the `autodiscover` string within that. That's potentially faster for the user if the server returned a larger body document. Also, it's a defensive measure against malicious server returning a huge document that we might attempt to parse and load completely into memory. 

I'm pushing a commit to a comment for `AUTODISCOVER_PROBE_MAX_BYTES` to hint at this behavior as well as `stream=True`, since it's less common pattern.

markstos: I also more clearly expressed 65536 as 64 * 1024  ... 64 bytes.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: I've pushed a commit with docs to clarify the behavior here.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: In benchmarking of the similar caching added in `mail/dkim.py`, the speed-up created was fractions of a millisecond, while the caching caused permanent but small memory growth for every cache entry stored. In the spirit of "don't optimize, benchmark",  the downside of the memory growth would seem to outweigh the minuscule performance benefit.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: See "benchmark before optimizing" comment above regarding caching.

> Also any reason why body is not decoded at this point? Does exchange return raw bytes or are we just parsing partially loaded xml?

Decoding earlier would require a charset decoding or guessing, which isn't necessary here. We are looking for `autodiscover` as a bytestring, which has the same representation in UTF-8/HTML/XML or plain text.

Returning bytes is not an Exchange thing but more of an HTTP protocol thing. We are explicitly requesting only the first 64 bytes from the server, so dealing with the bytestream was a natural fit here.

src/thunderbird_accounts/mail/autodiscover_probe.py
markstos: I'm aligned with minimizing deps to reduce supply-chain attack surface area, and the function we are using is only about 15 lines of code, so it's a good candidate for vendoring. This introduces shipping some Apache-2 licensed code alongside our MPL-2.0 code. From a first pass, the licenses are compatible and that can be done, but I'd like more experienced opinion on handling license mixing before I do that. Deferrable suggestion?

markstos:
> In the future I guess since we do want to get this out soon, I'd like to see this be stuffed into a class. I think a server probe interface would be a useful thing to have where it ingests a url, maybe optionally some credentials and probes a server which is handled by whatever impl you're using. So in this case the IsExchangeProbe / ExchangeAutodiscoverProbe would have all of this code implemented in it, but with one or two public functions to call.

@MelissaAutumn I agree this might be a good follow-up. Let's keep this in mind for the next time we write a probe. Until we have the second case, the abstraction may be YAGNI.

src/thunderbird_accounts/mail/autodiscover_probe.py
kewisch: It appears this combination is specifically allowed, given it is used as the example: https://www.mozilla.org/en-US/MPL/2.0/permissive-code-into-mpl/

MelissaAutumn (approved):

src/thunderbird_accounts/mail/autodiscover_probe.py
MelissaAutumn: Ah derp, I didn't notice the license. I'm okay with including the dep.

markstos: OK, let's accept the dep for now to defer the cross-license issue, but it does appear that we can embed an Apache-2 file with appropriate header later in the project if we decide to go that route.

PR #975, opened 2026-06-17
HOTFIX: Redirect users from /sign-up?email= to /login if account exists

### LINKED ISSUE #976: Users in `abandoned_cart` tag have a CTA of /sign-up instead of /login

davinotdavid:
## What changed?

When navigating to `/sign-up?email=`, if the user already has an account, redirect them to `/login` instead with the email pre-filled.

## Why?

The `abandoned_cart` email automation has a link to `/sign-up?email=` which is the sign up flow. These users already have accounts so they would be misled into thinking they need to recreate it or they have to manually change it to the login flow.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/976

### DIFF
--- a/src/thunderbird_accounts/core/views.py
+++ b/src/thunderbird_accounts/core/views.py
@@ -75,7 +75,7 @@ def home(request: HttpRequest):
         if email:
             existing_user = get_user_by_contact_email(email)
             if existing_user:
-                return HttpResponseRedirect(create_login_hint_url(email))
+                return HttpResponseRedirect(create_login_hint_url(existing_user.recovery_email))
 
     if request.user.is_authenticated and request.user.has_active_subscription:
         try:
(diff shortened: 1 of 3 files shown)


src/thunderbird_accounts/core/views.py
```diff
@@ -67,6 +68,15 @@ def home(request: HttpRequest):
...
         return HttpResponseRedirect(reverse('logout'))
 
+    # abandoned_cart tagged users have a subscribe CTA button that points to /sign-up?email=email
+    # however, we should instead redirect them to login if they already have an account with that email
+    if not request.user.is_authenticated and request.path == '/sign-up':
+        email = request.GET.get('email')
+        if email:
+            existing_user = get_user_by_contact_email(email)
+            if existing_user:
+                return HttpResponseRedirect(create_login_hint_url(existing_user.recovery_email))
```
Sancus: recovery_email may be typically present currently, but it can be null so this would 500 in that case

Sancus (changes requested):
I'm not sure if recovery email is the right login hint here, but if it is, we need to make sure the user has one before this flow can work at all.

src/thunderbird_accounts/core/views.py
davinotdavid: fair enough, to solve for both comments I've added `email` as fallback since we would only enter this branch if there's `email`

Sancus (approved):

PR #978, opened 2026-06-17
ci(release): fail fast when v<version> tag already exists in ECR

aatchison:
## Summary

Today the `deploy-production` workflow's "Retag docker image" step failed with an opaque error when publishing a release whose `deployment.json` carried a version that had already been pushed to ECR:

> tag invalid: The image tag 'v1.14.2' already exists in the 'thunderbird/accounts' repository and cannot be overwritten because the tag is immutable.

(See [run 27718887661](https://github.com/thunderbird/thunderbird-accounts/actions/runs/27718887661/job/81999993386).)

### Root cause

- `merge.yml` generates `deployment.json` at stage-build time using the *current* `pyproject.toml` version, and uploads it to a draft release.
- `release.yml` reads that `version` field and pushes `v<version>` on publish.
- ECR has tag immutability on this repo.
- When a draft is built before `pyproject.toml` is bumped and is later published as if it were the next version, the `v<version>` push collides with the already-published tag.

The release-draft title in the GitHub UI is purely cosmetic — editing it to say "v1.14.3" does **not** change `deployment.json`.

### What this PR does

Adds a pre-flight `aws ecr describe-images` check before the retag step. If the `v<version>` tag already exists, the workflow exits early with a clear annotated error pointing at the actual remediation:

> Tag v1.14.2 already exists in ECR repository thunderbird/accounts and the repo has tag immutability enabled.
>
> This draft was almost certainly built before pyproject.toml was bumped to v1.14.2.
>
> Publish the newer draft built from the version-bump commit (its deployment.json will carry the new version) instead of this one.

Failure mode goes from "1-minute docker push, opaque ECR error" to "instant skip with explanation."

## Test plan

- [ ] Lint passes on `release.yml`
- [ ] On the next legitimate publish (e.g., r-0437 → v1.14.3), the new check passes through (the tag does not yet exist) and the workflow proceeds normally
- [ ] If we ever re-publish a colliding draft, the annotation shows in the GHA UI with the remediation text

## Notes

- Requires `ecr:DescribeImages` on the deploy role. The role already has `ecr:GetAuthorizationToken` + push permissions; `DescribeImages` is normally bundled with those (e.g., via `AmazonEC2ContainerRegistryPowerUser`). If the role doesn't have it, this step will fail with an AccessDenied that's easy to read and fix.
- A longer-term hardening would be in `merge.yml`: only create a draft release when `pyproject.toml`'s version has actually changed since the last published release. Happy to do that in a follow-up if folks want.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,26 @@ jobs:
         with:
           mask-password: "true"
 
+      # Fail fast with a clear message if the v<version> tag already exists in ECR.
+      # ECR has tag immutability on this repo, so the later docker push would fail
+      # with an opaque "tag invalid" error. This happens when a draft release is
+      # published whose deployment.json was generated before pyproject.toml was
+      # bumped — the draft carries the previous version and collides with the
+      # already-published tag.
+      - name: Check version tag is not already published
+        run: |
+          SEMVER="v${{ steps.deployment-data.outputs.semver }}"
+          SOURCE_TAG="${{ steps.deployment-data.outputs.ecr-tag }}"
+          REPO=$(echo "$SOURCE_TAG" | cut -d/ -f2- | cut -d: -f1)
+          if aws ecr describe-images --repository-name "$REPO" \
+               --image-ids imageTag="$SEMVER" \
+               --region eu-central-1 >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


aatchison:
Closing per discussion — root cause was a self-inflicted publish of a draft whose deployment.json predated the version bump, not a workflow defect. We'll revisit hardening if it recurs.

Issue #973, opened 2026-06-16
Sign-Up flow interpret email from token


MelissaAutumn:
We'll be interpreting their recovery email automatically from the token listed in the url. This will most likely be their email or a base64 encoded version of their email.

Additionally with this step, we no longer need to show the recovery email step. So let's remove that and the sign-up link on the login page.

Changes:
* [x] Add invite only warning to login page
* [x] Remove "Subscribe" link on login page
* [x] Remove recovery email step
* [x] Interpret recovery email from sign-up / email token

PR #980, opened 2026-06-17
Sign-Up Flow UI Updates

### LINKED ISSUE #972: Sign-Up flow changes
More of a mega-issue for the following changes:
* [ ] Add confirm plan step on sign-up flow
* [ ] Copy changes on "Check your Email" step
* [ ] Copy changes on "Email verified" step
* [ ] Copy changes on generic AIA is done screen (Your account is ready)

davinotdavid:
## What changed?

- Updated `SignUpLayout` with:
  - A few CSS tweaks but also the ability to have a secondary / alternative action button for the `Not Interested` button in the `StepConfirmPlan`
  - Main button changed from `primary-button` to `brand-button` according to the Design files
- Added `StepConfirmPlan` with dynamic data from Paddle with localized pricing, plan name and description.
- Updated the "Check Email Step" copy and button
- Updated the "Verify Email Step" copy and button
- Updated the step after confirming email / final step copy and button
- Small update to the error screen to have a `brand-button` "back to application"

## Why?

There might be a gap in time rom the time that users come to the waitlist to the actual time they sign up so users might need to be reminded of the pricing as a first step before they proceed on creating their account. 

## Limitations and Notes

- I'd recommend us to rename the previously existing `Step1Username`, `Step2Password` and `Step3Verify` to not have the step number in the file / folder names so that we can reorder them at will. This has not been done in this PR though to avoid unnecessary big diffs!
- ~The pricing plan is static at the moment and not localized. The only place that we can do pricing localization is during the Paddle checkout.~

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/972

## Screenshots

Confirm Plan Step (New)
image

Check Email Step (Updated copy)
image

Verify Email (Updated copy and action button)
image

After confirming email / Final step (Updated copy and action button)
image


assets/app/vue/defines.ts
```diff
@@ -10,4 +10,19 @@ export const STATUS_PAGE_URL = 'https://status.tb.pro/';
...
+  'description': 'Our initial offering, ideal for most users',
+  'price': '6',
+  'period': 'per month,<br>paid annually',
+  'cta_label': 'Join Waitlist',
+  'mail_storage': '30', // GB
+  'send_storage': '60', // GB
+  'num_domains': '3',
+  'num_inboxes': '1',
+  'num_email_addresses': '15',
+}
```
MelissaAutumn: I was going to have this data be placed on the template so we would never have to update this, but honestly this'll work for now.

davinotdavid: Ah that'd be better yes, moved to the component level. Thanks!

assets/app/vue/components/PlanCard.vue
```diff
@@ -0,0 +1,147 @@
+<script setup lang="ts">
+// TODO: Fetch the active plan from the backend / Paddle (?)
```
MelissaAutumn: Oh I didn't notice we merged in https://github.com/thunderbird/thunderbird-accounts/pull/477

We should probably pull that down from django then, huh. Can we swap the actual text to use the localization values from the subscription component? (If they don't match, just add the new ones.) And then can we embed (in template) or pull down (via api) the paddle price id and use https://developer.paddle.com/paddle-js/methods/paddle-pricepreview/ to calculate a localized price? 

Seems like the easiest lift imo. You can use LoadingSkeleton for the price portion of that component with an initial value of 6. 

We can leave the plan info as-is. Since we're probably not going to change that for a while.

If you think this is too much of a lift, we can make this a fast follow.

assets/app/vue/components/PlanCard.vue
```diff
@@ -0,0 +1,147 @@
...
+</script>
+
+<template>
+  <div class="plan-card">
+    <div class="left">
+      <h3>{{ TBPRO_DEFAULT_PLAN.name }}</h3>
+      <p>{{ TBPRO_DEFAULT_PLAN.description }}</p>
+      <h4>
+        <span class="price">${{ TBPRO_DEFAULT_PLAN.price }}</span>
+        <span v-html="TBPRO_DEFAULT_PLAN.period" />
```
MelissaAutumn: Let's avoid use of v-html where possible.

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
```diff
@@ -5,6 +5,7 @@ import { captureError, captureStep } from '../api';
 
 export const enum SIGN_UP_STEPS {
   INVALID = 0,
+  CONFIRM_PLAN = 5,
```
MelissaAutumn, line 8: And that's why I always use 10s for enum values. 😄

assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue
```diff
@@ -0,0 +1,52 @@
...
+  name: 'StepConfirmPlan',
+};
+</script>
+
+<template>
+  <sign-up-layout
+    step-id="step-confirm-plan"
+    :title="$t('views.mail.views.signUp.stepConfirmPlan.title')"
+    :subtitle="$t('views.mail.views.signUp.stepConfirmPlan.subtitle')"
+    :submit-disabled="false"
```
MelissaAutumn: If we pull data from Paddle we should use the disabled states.

MelissaAutumn (approved):
Looks good after addressing comments.

Main blocker here is the localized text, if you can swing in the localized prices from paddle.js that'd be rad. If not let's make that a fast follow.

MelissaAutumn:
I agree about removing the Step numbers from the pages. I'll address that in #973

assets/app/vue/components/PlanCard.vue
davinotdavid: Oh interesting! Thanks for the guidance here! I gave it a try here and it seems to be working well. The localized pricing, plan name and description is being pulled in:

image

Code-wise, it looks a little odd now that we have some stuff in `env.d.ts` that are pertaining to the `CheckoutStep` and some for the `StepConfirmPlan` that look very similar.

assets/app/vue/components/PlanCard.vue
davinotdavid: Agreed, removed!

davinotdavid:
@MelissaAutumn alright! Thanks again for the review and guidance!

Last commit adds pulling the plan info from the backend and the localized pricing from Paddle! Is this what you were thinking?

MelissaAutumn:
Looks good now! Had to remove some extra slots (you only need those if you're adding another layer of templating, not if you're going to use it. 

Good to merge!

PR #984, opened 2026-06-18
Gate sign up flow

### LINKED ISSUE #973: Sign-Up flow interpret email from token

MelissaAutumn:
Fixes #973 

This should resolve the following:
*    Add invite only warning to login page
*    Remove "Subscribe" link on login page
*    Remove recovery email step
*    Interpret recovery email from sign-up / email token

I'll re-check this PR in the morning as I'm quite tired, but this checks out on my end. I did not end up renaming the numbered step components for ease of review. We can fast follow that.


assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
```diff
@@ -104,7 +102,7 @@ export const useSignUpFlowStore = defineStore('signUpFlow', () => {
       if (type.indexOf('invalidPassword') === 0) {
         step.value = SIGN_UP_STEPS.PASSWORD;
       } else if (type === 'status-409') { // User already exists error (this comes from keycloak)
-        step.value = SIGN_UP_STEPS.VERIFY;
+        step.value = SIGN_UP_STEPS.USERNAME;
```
MelissaAutumn: This should be impossible unless there's a user in keycloak that's not in accounts btw.

assets/app/vue/views/SignUpView/views/Step2Password/index.vue
```diff
@@ -26,7 +26,14 @@ const onSubmit = async () => {
     return;
   }
 
-  nextStep();
+  const response = await submit();
+  if (response === true) {
+    await nextStep();
```
MelissaAutumn, line 31: This will submit Step.DONE to posthog.

assets/app/vue/router.ts
```diff
@@ -43,7 +44,44 @@ const routes: RouteRecordRaw[] = window._page?.isErrorPage ? [
...
+              'Content-Type': 'application/json',
+              'X-CSRFToken': window._page?.csrfToken,
+            },
+          });
+
+          const data = await response.json();
+          if (data?.allowed) {
+            return true; // continue along with the request
+          } else if (data?.detail) { // rate limited
+            throw new Error(data?.detail);
```
MelissaAutumn: Not ideal, but they'll have to refresh 10 times in a single minute to hit this.

src/thunderbird_accounts/authentication/tests/test_api.py
```diff
@@ -214,3 +214,52 @@ def test_import_errors_propagate(self, mock_import_user: MagicMock):
...
+
+
+@override_settings(USE_ALLOW_LIST=True)
+class CanISignUpTestcase(APITestCase):
+    def setUp(self):
+        self.client = RequestClient()
+
+        self.wait_list = [
+            '348b8787-ddf0-4ab9-931e-db388f8770c0@example.com',
+            '153ff958-dabb-4afc-94ab-3b1b3b5ff051@example.com',
```
MelissaAutumn, line 227: Gonna start using uuids in tests (via python's uuid.uuid4() lol) to avoid name collisions.

src/thunderbird_accounts/core/views.py
```diff
@@ -68,16 +67,6 @@ def home(request: HttpRequest):
...
-    # abandoned_cart tagged users have a subscribe CTA button that points to /sign-up?email=email
-    # however, we should instead redirect them to login if they already have an account with that email
-    if not request.user.is_authenticated and request.path == '/sign-up':
-        email = request.GET.get('email')
-        if email:
-            existing_user = get_user_by_contact_email(email)
-            if existing_user:
-                login_hint = existing_user.recovery_email or email
-                return HttpResponseRedirect(create_login_hint_url(login_hint))
-
```
davinotdavid, line 81: Thanks for removing this! 🙏

assets/app/vue/router.ts
```diff
@@ -43,7 +44,44 @@ const routes: RouteRecordRaw[] = window._page?.isErrorPage ? [
...
+          }
+
+          failQueryParam = `?email=${email}`;
+        } catch (e) {
+          console.error("Failed to fetch allowed status for the sign up page. Error: ", e);
+        }
+      }
+
+      // Fail state, send them to the wait list.
+      window.location.href = `${TBPRO_WAIT_LIST}${failQueryParam}`;
```
davinotdavid, line 91: Since we've removed the ~great~ `abandoned_cart` logic in the `home` method, we would still need to verify whether or not we should redirect the user to the login page instead of the waitlist.

Currently, if we have a user who has been through the waitlist + sign up flow but haven't paid, they will receive and email with a CTA of `/sign-up?email=user@recoveryemail.com` and in this current logic they would be redirected to the waitlist instead of being redirected to `login`.

Even though this is not a nice flow it was the hotfix implemented for the users who already received the email with said CTA :(

davinotdavid (changes requested):
The flows proposed in the original issue work great, thanks! 🎉 Overall code LGTM with a blocker:

I have some concerns on the `/sign-up?email=` bit that was hotfixed before and would not work anymore after merging this.

Other than that, I see that [you've commented in Zeplin](https://zpl.io/jpkXKq4) on the position of the notice bar but I don't see that being a blocker for this IMHO. I've also noticed that we don't have the `Keep me signed in` in the designs anymore and I've asked a clarifying question there as well on the fixed `@thundermail.com` suffix since we still allow (and need, for the time being) to be able to log in with the recovery email.

Current Zeplin mockup for reference:
Image

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
davinotdavid: This would be great to append in the comment in the code imho!

assets/app/vue/router.ts
MelissaAutumn: Ah right, I didn't actually have the entire context about the abandoned cart email having the wrong link but now I do. I will look into this regression.

assets/app/vue/views/SignUpView/stores/signUpFlowStore.ts
MelissaAutumn: Gooood point.

MelissaAutumn:
Okay switched it over to a 3 value enum. It should work like before, but I want to move off this asap. 

I also fixed up how rate limiting on that page is handled. I don't know if `/chill` is maybe a little too jokey, but it's not hard to tweak, and I'm expecting no one to be refreshing the page at least 10 times to see it.

assets/app/vue/router.ts
```diff
@@ -43,7 +45,53 @@ const routes: RouteRecordRaw[] = window._page?.isErrorPage ? [
...
+            }),
+            headers: {
+              'Content-Type': 'application/json',
+              'X-CSRFToken': window._page?.csrfToken,
+            },
+          });
+
+          const data = await response.json();
+          console.log("Guard response:", data);
+          if (data?.go_to === CAN_I_SIGN_UP_RESPONSES.SIGN_UP) { // continue along with the request
```
MelissaAutumn, line 70: This sucks 😃

MelissaAutumn:
e2e tests are failing, but I think that's due to to the logic change. Will fix in a minute.

davinotdavid (approved):
Nice! Works well during my testing here 🚀 

Sorry about the unfortunate logic that we need to keep for the abandoned cart bit but hopefully we will change it soon™.

After fixing the E2E I think we are good to merge and test the whole flow on stage!

MelissaAutumn:
@davinotdavid ignore e2e test fails for now. I've split that off into #986

PR #985, opened 2026-06-19
WIP: Zendesk Sidebar using OAuth

markstos:
## Description

What's here tries to implement a Zendesk Sidebar App to display the account details of a related Thundermail customer.

Authentication is handled by following Zendesk's standard OAuth pattern.
That includes Zendesk supplying us a public key and well as the installing
admin going through an OAuth flow during installation.

To avoid Iframe/CORS challenges, Zendesk proxies requests through themselves and
then on to our. On our end, we set a head indicating we allow Zendesk to display
our site within an iframe.

## Status

In the first draft that Philipp worked on, it demonstrated the app displaying all the related user details, using a custom authentication flow where ever user logged in individually (nice for auditing access). After running into CORS challenges with that, I updated the code to try Zendesk's prescribed OAuth flow.

The current status of the code is that the content under `zendesk/` can be installed
as a Zendesk app according to to the instructions in the [./zendesk/README.md](./zendesk/README.md).

The initial connection to our /zendesk/sidebar/ succeeds, but then loading the /zendesk/sidebar/content fails. Zendesk shows a generic error in the browser:

```
{"errors":[{"code":"ConnectionNotFound","title":"connection not found"}]}
```

This errors appears to be caused by the Zendesk proxy failing to connect us. Our logs don't show the second connection actually go all the way to us, and the `ConnectionNotFoundError` doesn't appear in our code base. Perhaps some value in our config is resulting in a wrong or incomplete connection string?

## Testing

For testing, it's useful to be able to a version of code running on our laptop as a public web server so that Zendesk can connect to that an iFrame, as Zendesk's proxy can't connect to our "localhost". For this, `tailscale funnel` is a great fit. I used a personal Tailscale account for this, but I hear that we have a corporate Tailscale account being rolled out, which could be a great fit here.

* For testing locally with `tailscale funnel`, `PUBLIC_BASE_URL` needs to be set to the with your Tailscale host name. `tailscale funnel 8087` is used for forward that domain to the local thunderbird-accounts service running with `sudo docker compose up -V`

* To make a valid `zendesk/manifest.json`, we need to fill in a valid `oauth.client_id` and `oauth.client_secret`. For staging, these come from our Staging Keycloak instance, in the `tppro-dev` realm, under the `zendesk-sidebar-stage` client. That serves as the client_id and the client secret can be found there as well. Note the list of clients are paginated, and it may be on the second page.

* When installing a Zendesk "Sidebar App", we'll be prompted for a "Sidebar URL" and a "API Host". Examples are as follows, assuming a local Tailscale setup:
  * `sidebar_url`: Like `https://your-host.random-words.ts.net/zendesk/sidebar/`
  * `api_host`: Like `your-host.random-words.ts.net`

The initial Zendesk OAuth flow after installing the app into our Zendesk staging area needs to done with a user that exists in the Keycloak tbpro-dev realm.

If you want to start over with a new Zendesk app, you'll need to discover the "app_id"and "installation_id" for it. These can be done with `curl` or `xh` and basic auth.
I prefer `xh` for the prettier output. It works like `httpie` can be installed with `brew`:

```
xhs --auth '{{user}}/token:{{api_token}}' {{subdomain}}.zendesk.com/api/v2/apps/installations.json
```

### ZENDESK_SUBDOMAIN

* Setting this is essential for testing.

### ZENDESK_APP_AUDIENCE

```
# Substitute the value of our Zendesk staging subdomain here
# The ID here is the "installation id" of the Thundermail Zendesk app.
ZENDESK_APP_AUDIENCE="https://{{subdomain}}.zendesk.com/api/v2/apps/installations/52711738346259.json"
```

### ZENDESK_APP_PUBLIC_KEY

```
# This is Zendesk's public key for our app, which does not need to be secret
# It is supposed to have literal \n chars embedded hhere
ZENDESK_APP_PUBLIC_KEY="-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyOMz5h1fiTTVDuE+t8i6\nunMB0jaNav3UFuEjuDZAHMr2bFMpHrv5V0v/OwGzksDbiA/YeeBFUMzigIPYx3L/\n0+9bQYybcqdu1EeACDSVg+B/d1kd0ouUr40MWPzwsQNZYTyrI/cdTbtxw+JOrypi\nlobLASXVVnzhlt4etXzKd6oSWQDDr22Tm61rlwnZYepylXvU8DDKUSHyF91zxHZz\n44GksysIkyiJgbyqjW5OCAHoDlg1ET3cCE/cfoMTi6OOZ7zyqjG7XYPMje5m2yHO\np2iL9d9OmK9Vkf9OHRTFpdwmY/7qKvyvA3J+r4qGG6esZEEqcq16+ZOYY98lslvC\nLwIDAQAB\n-----END PUBLIC KEY-----"
```

## Questions

It's not clear whether Zendesk's intended model is that the admin authorize once or that each agent goes through the OAuth flow. Zendesk's docs are not clear on this point.

## TODO

`manifest.json` contains some details hardcoded to a particular environment, but we need some different defaults for staging and production. How manage those differences is TBD. We would either maintain at least two completely separate per-env files or use some kind of templating system or file generation to generate manifest.json from some other source of truth, like the Pulumi config system.

The general remaining plan is to:

* Test locally, with Tailscale or similar
* Validate that the security approach is satisfactory for us
* Test on staging
* Deploy to production.

### Official Zendesk docs

 * [Adding Third-Party OAuth to a Support App](https://developer.zendesk.com/documentation/apps/build-an-app/adding-third-party-oauth-to-a-support-app/)
 * [Building a Server-Side App / Part 1 / Core Concepts](https://developer.zendesk.com/documentation/apps/build-an-app/building-a-server-side-app/part-1-core-concepts/)
 * [Sidebar Apps for Support](https://developer.zendesk.com/documentation/apps/app-design-guidelines/support/sidebar-apps-support/)
 * [Example of Adding third-party OAuth Support into a Zendesk Sidebar app](https://developer.zendesk.com/documentation/apps/build-an-app/adding-third-party-oauth-to-a-support-app/)
 * [manifest.json OAuth docs](https://developer.zendesk.com/documentation/apps/app-developer-guide/manifest/#oauth)


kewisch:
Closing this as we're taking a slightly different approach. Thanks for preparing this!

PR #990, opened 2026-06-22
Add Django Admin action to back-fill recovery_email from Keycloak

### LINKED ISSUE #836: Backfill recovery_email from Keycloak into Django Admin
We had users joining before the introduction of the recovery_email field in the User model in Django so we need an admin action / command to backfil those!

davinotdavid:
## What changed?

Added a new Django Admin action to back-fill `recovery_email` for Users.

## Why?

We've introduced the `recovery_email` field in the User model after having onboarded several emails so we need to back-fill this field. Since we have this info in Keycloak, we should source it from there.

## How to test

Thankfully, our `admin@example.org` user that is created automagically when spinning up a local environment from scratch doesn't have a recovery email set so we can simply do a `docker compose down -v` and a `docker compose up -d --build` and then:

- Navigate to `localhost:8087/admin/authentication/user/`
- See that the recovery email field for the admin user is empty
- Select the admin user and run the `Backfill recovery email from Keycloak` action
- See that the recovery email field is filled!

## Applicable Issues

Related with https://github.com/thunderbird/thunderbird-accounts/issues/821
Closes https://github.com/thunderbird/thunderbird-accounts/issues/836

## Screenshots

image

### DIFF
--- a/src/thunderbird_accounts/authentication/admin/actions.py
+++ b/src/thunderbird_accounts/authentication/admin/actions.py
@@ -221,8 +221,8 @@ def admin_backfill_recovery_email(modeladmin, request, queryset):
     skipped = 0
     failed = 0
 
-    for user in queryset.filter(Q(recovery_email__isnull=True) | Q(recovery_email='')):
-        if not user.oidc_id:
+    for user in queryset.filter(Q(recovery_email__isnull=True) | Q(recovery_email='')).exclude(oidc_id__isnull=True):
+
             skipped += 1
             continue
 
--- a/src/thunderbird_accounts/authentication/admin/actions.py
+++ b/src/thunderbird_accounts/authentication/admin/actions.py
@@ -216,6 +216,14 @@ def admin_sync_plan_to_keycloak(modeladmin, request, queryset):
 
 @admin.action(description=_('Backfill recovery email from Keycloak'))
 def admin_backfill_recovery_email(modeladmin, request, queryset):
+"""Backfill's recovery emails from user with ``null`` recovery emails in this system and a valid one in Keycloak.
+
+Displays the following messages on the admin panel after the action is performed:
+* Updated: The number of user's that have been successfully updated with their recovery email.
+* Skipped: The number of user's who were skipped because they don't have a valid recovery email in Keycloak.
+* Failed: The number of user's that exist in this system but do not exist in Keycloak.
+
+"""
     keycloak = KeycloakClient()
     updated = 0
     skipped = 0
--- a/src/thunderbird_accounts/authentication/admin/actions.py
+++ b/src/thunderbird_accounts/authentication/admin/actions.py
@@ -4,9 +4,11 @@ import logging
 
 from django.conf import settings
 from django.contrib import messages, admin
+from django.db.models import Q
 from django.utils.translation import gettext_lazy as _, ngettext
 
-from thunderbird_accounts.authentication.exceptions import UpdateUserPlanInfoError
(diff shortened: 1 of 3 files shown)


src/thunderbird_accounts/authentication/admin/actions.py
```diff
@@ -212,6 +214,75 @@ def admin_sync_plan_to_keycloak(modeladmin, request, queryset):
...
 
+@admin.action(description=_('Backfill recovery email from Keycloak'))
+def admin_backfill_recovery_email(modeladmin, request, queryset):
+    keycloak = KeycloakClient()
+    updated = 0
+    skipped = 0
+    failed = 0
+
+    for user in queryset.filter(Q(recovery_email__isnull=True) | Q(recovery_email='')):
+        if not user.oidc_id:
```
MelissaAutumn:
```suggestion
    for user in queryset.filter(Q(recovery_email__isnull=True) | Q(recovery_email='')).exclude(oidc_id__isnull=True):

```
I think this should work, pls test though.

src/thunderbird_accounts/authentication/admin/actions.py
```diff
@@ -212,6 +214,75 @@ def admin_sync_plan_to_keycloak(modeladmin, request, queryset):
         )
 
 
+@admin.action(description=_('Backfill recovery email from Keycloak'))
+def admin_backfill_recovery_email(modeladmin, request, queryset):
```
MelissaAutumn, line 218:
```suggestion
def admin_backfill_recovery_email(modeladmin, request, queryset):
"""Backfill's recovery emails from user with ``null`` recovery emails in this system and a valid one in Keycloak.

Displays the following messages on the admin panel after the action is performed:
* Updated: The number of user's that have been successfully updated with their recovery email.
* Skipped: The number of user's who were skipped because they don't have a valid recovery email in Keycloak.
* Failed: The number of user's that exist in this system but do not exist in Keycloak.

"""
```

Thought it might be handy to include what the counters mean somewhere. Code seems the easiest place for them. Feel free to touch it up though!

src/thunderbird_accounts/authentication/tests/test_admin.py
```diff
@@ -463,3 +464,40 @@ def test_success_without_stalwart_account(self, mock_requests: MagicMock, mock_d
 
         mock_delete_principal.assert_not_called()
 
+
+@patch.object(CustomUserAdmin, 'message_user')
+@patch('thunderbird_accounts.authentication.admin.actions.KeycloakClient')
+class AdminBackfillRecoveryEmailActionTest(TestCase):
```
MelissaAutumn, line 470: Can we add a test with multiple users, and make sure only one updates?

MelissaAutumn (changes requested):
Looks good to me, just have a few suggestions here.

src/thunderbird_accounts/authentication/admin/actions.py
davinotdavid: It does work, nice! Thanks!

src/thunderbird_accounts/authentication/tests/test_admin.py
davinotdavid: Yep! Added, good case to add, thanks!

src/thunderbird_accounts/authentication/admin/actions.py
davinotdavid: Thanks! That's already quite good in my opinion so just applied your suggestion!

MelissaAutumn (approved):

PR #992, opened 2026-06-22
Update pulumi images

MelissaAutumn:
Just some image updates after a keycloak prod deploy.


MelissaAutumn:
(these are already deployed btw)

ryanjjung (approved):

PR #995, opened 2026-06-22
MFA Keycloak Changes

Sancus:
keycloak-only changes for MFA.

This includes flow config, the new JAR, and theme changes only. Basically just the stuff that needs to be deployed to keycloak.

Once this is merged, the MFA PR needs to be rebased.

### DIFF
--- a/keycloak/themes/tbpro/assets/vue/views/LoginRecoveryAuthnCodeView/index.vue
+++ b/keycloak/themes/tbpro/assets/vue/views/LoginRecoveryAuthnCodeView/index.vue
@@ -1,15 +1,14 @@
 <script setup>
-import { TextInput, PrimaryButton, SelectInput, LinkButton } from "@thunderbirdops/services-ui";
-import { computed, ref, useTemplateRef } from "vue";
+import { TextInput, PrimaryButton, LinkButton } from "@thunderbirdops/services-ui";
+import { computed, useTemplateRef } from "vue";
 import MessageBar from '@kc/vue/components/MessageBar.vue';
 import CancelForm from '@kc/vue/components/CancelForm.vue';
 
-const errors = window._page.currentView?.errors;
 const formAction = window._page.currentView?.formAction;
-const userOtpCredentials = window._page.currentView?.userOtpCredentials;
-const selectedOtpCredential = window._page.currentView?.selectedOtpCredential;
+const recoveryCodePrompt = window._page.currentView?.recoveryCodePrompt;
+const errors = window._page.currentView?.errors;
 const showTryAnotherWay = window._page.currentView?.showTryAnotherWay === 'true';
-const OtpCredentialModel = ref(selectedOtpCredential);
+
 const loginForm = useTemplateRef('login-form');
 const tryAnotherWayForm = useTemplateRef('try-another-way-form');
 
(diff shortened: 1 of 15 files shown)


keycloak/themes/tbpro/assets/vue/views/LoginRecoveryAuthnCodeView/index.vue
```diff
@@ -0,0 +1,80 @@
...
+import MessageBar from '@kc/vue/components/MessageBar.vue';
+import CancelForm from '@kc/vue/components/CancelForm.vue';
+
+const formAction = window._page.currentView?.formAction;
+const recoveryCodePrompt = window._page.currentView?.recoveryCodePrompt;
+const errors = window._page.currentView?.errors;
+const showTryAnotherWay = window._page.currentView?.showTryAnotherWay === 'true';
+
+const loginForm = useTemplateRef('login-form');
+const tryAnotherWayForm = useTemplateRef('try-another-way-form');
```
MelissaAutumn, line 13: I really need to setup typescript on this vue app.

keycloak/themes/tbpro/assets/vue/views/LoginRecoveryAuthnCodeView/index.vue
```diff
@@ -0,0 +1,80 @@
...
+<script>
+export default {
+  name: 'LoginRecoveryAuthnCodeView'
+};
+</script>
+
+<template>
+  <div class="panel">
+    <h2>{{ $t('authRecoveryCodeHeader') }}</h2>
+    <form id="kc-recovery-code-login-form" ref="login-form" method="POST" :action="formAction" @submit.prevent="onSubmit"
```
MelissaAutumn, line 37: I don't think you'll need the submit.prevent if all we're doing is submitting. Although it might be a good idea to keep that, and include a loading.value = true. Then lock the buttons on this form if loading == true.

MelissaAutumn (approved):
Looks pretty standard, I have not poured over the docs for the config values, but I'm going to assume they're reasonable. 

Sorry the pr took so long to review!

keycloak/themes/tbpro/assets/vue/views/LoginRecoveryAuthnCodeView/index.vue
MelissaAutumn: I realize that this is probably a copy and paste from my code, haha. Thanks for dealing with my weird theme.

Sancus: I think that `.prevent` is preventing some forms of native submit right now like hitting Enter? I think we could make the button a true native form submit button if we wanted to, but if we keep it as-is then locking the form during loading isn't a bad idea. If we do thta we should do it in all the submittable forms though, and I'm kind of reclutant to add code for that right now.

I'd rather get this merged and then we can tweak edge cases later if merited.

MelissaAutumn: lgtm

Sancus:
I noticed the "try another way" flow got lost somewhere in the infinite refactors this has had, so I put it back, now you can use recovery codes for both login and step-up instead of just step-up, as was always intended.

I also changed one label to better align with the designs. But this should be good to merge.

Issue #908, opened 2026-06-03
E2E tests fail in GH Actions for forks due to missing env vars


davinotdavid:
## Describe the bug

Currently, whenever someone submits a PR from a fork, the CI will fail tests. We should revise the `run-e2e-tests-local` job here: https://github.com/thunderbird/thunderbird-accounts/blob/main/.github/workflows/validate.yml#L120-L210

This is the error:
`manage.py create_plan_for_e2e_test: error: the following arguments are required: paddle_product_id`

## To Reproduce

Make a PR from a fork and check CI failing

## Expected behavior

CI should pass in PRs from a fork so that contributors (and reviewers) can feel confident about their work.

## Actual behavior

CI fails

## Additional context

There's some paddle env vars that are required for running the e2e tests and GitHub doesn't expose repository/environment secrets to pull_request workflows from forks. This might require investigation!

kewisch:
@rwood-moz could you look into this soon? We need PRs to work from a fork given the contribution model we're aiming for, but obviously don't want to expose a risk of contributors being able to access environments they shouldn't have access to.

rwood-moz:
Some preliminary info courtesy of ChatGPT:

```
Yes. This is expected GitHub Actions security behavior.

For pull_request workflows from forks, GitHub does not pass repository/org/environment secrets to the runner, except 
for the built-in GITHUB_TOKEN, and that token is typically restricted/read-only for fork PRs. GitHub documents this 
explicitly: “Actions secrets are not passed to workflows that are triggered by a pull request from a fork.”

Why: fork PR code is untrusted. If secrets were exposed, a contributor could modify the workflow or test code to 
print/exfiltrate them.

Common ways to handle it:

Split workflows
Run safe CI on pull_request without secrets, and run secret-dependent jobs only on trusted branches or after merge.

Use pull_request_target carefully
This runs in the context of the base repo, so secrets are available, but it is dangerous if you check out and execute 
the fork’s code. Safe pattern: use it only for labeling, commenting, validation, or a controlled “approval gate,” not 
arbitrary test execution.

Manual trusted rerun / approval pattern
Require a maintainer to review the PR, then run a privileged workflow against a reviewed SHA or copied branch.

Use environments with required reviewers
Put sensitive secrets in a GitHub Environment and require approval before jobs can access them.
```

kewisch:
[...]

PR #1001, opened 2026-06-23
Zendesk sidebar - customer endpoint

### LINKED ISSUE #1000: Add customer endpoint for Zendesk Support Sidebar
Add an endpoint the Zendesk ZAF application can use to retrieve customer support information and display in the sidebar. Make sure it is only possible if the user has respective support permissions.

kewisch:
## What changed?

Add an additional API call that allow us to retrieve all relevant customer information at once, for the Zendesk sidebar application.

This was written with AI assist, though I've reviewed and adapted the code manually to make sure it fits. The tests may be a little more AI heavy.

## Limitations and Notes

I'm not 100% sure I got the authentication classes right, and we may want to do some additional checks. @mzeier mentioned we could put this endpoint behind tailscale, but if there are any additional checks we could add let me know.

I also took the opportunity to refactor some support related code into its own module. It may make sense to review each commit separately since one is basically just moving code.

## Applicable Issues

Fixes #1000


kewisch:
Test failures are likely #908 again

davinotdavid (approved):
lgtm! given the demo you gave during the services meeting, I am reviewing this with the assumption that the payloads match the expectations of the sidebar.

### QA Log

- checked diff
- ran tests locally
- tested with cURL with staff account and normal account to check permissions
- tested manually with Zendesk credentials to see if contact form fields still load successfully without auth

Issue #986, opened 2026-06-19
E2E tests fail due to sign up changes


MelissaAutumn:
e2e tests are failing, but I think that's due to to the logic change.

_Originally posted by @MelissaAutumn in https://github.com/thunderbird/thunderbird-accounts/issues/984#issuecomment-4754068499_
            

The fix isn't particularly trivial on the frontend side of things so I'm splitting this up into a fast follow ticket.

MelissaAutumn:
We'll need a method of setting up some values on the allow list. Probably through cli management commands, we can also use envs for the actual email values, and we'll have to document this in the readme.

MelissaAutumn:
For the allow list portion:
* A test account or admin account can provision a new allow list entry
* This new allow list entry contains a flag that will make this user a test account upon sign up and mark their account for deletion after an hour or something

rwood-moz:
I'm going to temporarily disable the failing sign-up E2E tests (in #1015) but we can re-enable them once this issue is resolved.

PR #1004, opened 2026-06-24
Pin dotenv to earlier ver to avoid console adverts

### LINKED ISSUE #999: Remove dotenv from e2e tests
Not sure I trust a package that includes ads in the terminal...We should move away from dotenv to something built-in or trustworthy.

rwood-moz:
## What changed?
Pin the node `dotenv` package to an earlier version that doesn't post adverts to the console log.

## Why?
@MelissaAutumn Noticed that the dotenv package was posting an advert in the console output. Turns out this is something that was added to the `dotenv` package in version 17.x and forward; but earlier versions of the package don't post to the console.

## Applicable Issues
Fixes #999.

## QA Log
Ran the E2E tests locally and confirmed that the dotenv output line (including the advert) no longer appears in the console.

### DIFF
--- a/test/e2e/package-lock.json
+++ b/test/e2e/package-lock.json
@@ -13,7 +13,7 @@
         "@playwright/test": "1.59.1",
         "@types/node": "^24.10.1",
         "browserstack-node-sdk": "1.52.3",
-        "dotenv": "^17.2.1",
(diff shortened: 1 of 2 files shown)


rwood-moz:
Note the failing PR E2E check is from an unrelated issue (#986).

PR #1006, opened 2026-06-25
Fix all-other-session logout option in MFA flow. (Fixes #1005)

### LINKED ISSUE #1005: "sign out from other devices" in OTP setup nukes the current session too
It should only nuke all other sessions, not your current one.

Sancus:
## Summary
This patch fixes nuking of the current session and also switches to an endpoint that is restricted to the user's token for session logout, which is a net positive for security. Also added tests for this option.

## Manual Testing

1.  https://mfa-accounts.thunderbird.dev uses your credentials from stage-thundermail.com for auth and is updated for these changes. 
2. Go through the MFA setup flow, check "sign out all other devices" when verifying your authenticator app. 
3. Note that you do NOT immediately receive a session expired error and can continue normally.

## Notes
The way we set Keycloak URL in settings is weird, it should be a base domain endpoint and then add whatever paths(realm, interface, etc) as necessary, but instead we start off with a full URL including path.

Fixing this creates a bunch of extraneous changes though so I decided it was out of scope for this patch. Thus we have to derive the URL for the accounts REST API in a weird way.

Fixes #1005

### DIFF
--- a/src/thunderbird_accounts/authentication/clients.py
+++ b/src/thunderbird_accounts/authentication/clients.py
@@ -133,10 +133,6 @@ class KeycloakClient:
         self.request(f'users/{oidc_id}/credentials/{credential_id}', RequestMethods.DELETE)
         return True
 
-    def logout_user_sessions(self, oidc_id: str) -> bool:
-        self.request(f'users/{oidc_id}/logout', RequestMethods.POST)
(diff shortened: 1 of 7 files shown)


MelissaAutumn (approved):
lgtm

PR #959, opened 2026-06-12
[WIP] Stalwart v0.16 Upgrade

MelissaAutumn:
Draft PR for visibility.

PR #1018, opened 2026-06-25
Add django-waffle so we can start feature flagging. :waffle: (Fixes #1017)

### LINKED ISSUE #1017: Add waffle
🧇 feature flags 🧇

MelissaAutumn:
Fixes #1017 

This PR simply adds django-waffle so we can start developing against it. I thought about moving the mfa feature flag behind waffle, but didn't want to scope creep this 3 second PR. 

It's mostly pulled out of my #959 changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ dependencies = [
     "cloudflare>=5.2.0",
     # CVE pins
     "idna>=3.15", # CVE-2026-45409 fix
+    "django-waffle>=5.0.0",
 ]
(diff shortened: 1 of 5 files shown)


MelissaAutumn:
Oh forgot the documentation: https://waffle.readthedocs.io/en/stable/about/why-waffle.html

It's a feature flag system for django. In the admin panel we can add flags, switches, or samples. We then add code referencing those models and the library handles / admin panel settings handle the rest.

Sancus (approved):

PR #1019, opened 2026-06-25
Skip recovery codes require an auth app if we cannot remove them.

MelissaAutumn:
The repo env isn't setup, so skipping this for now.


Sancus (approved):

rwood-moz:
Thanks @MelissaAutumn :)

PR #1022, opened 2026-06-25
Use waffle for MFA feature flag instead of localStorage. (Fixes #1020)

### LINKED ISSUE #1020: Place MFA behind a waffle flag
We'll have to set it up on stage/prod, but we should move the client-side flag to a waffle flag. 

Waffle flags should be accessible via window.waffle, I went ahead and created an initial flag on stage labelled `multi-factor-authentication`. 

https://accounts-stage.tb.pro/admin/waffle/flag/1/change/

It can be accessed in js via `window.waffle.FLAGS['multi-factor-authentication']` or via `waffle.flag_is_active('multi-factor-authentication')`. 

https://waffle.readthedocs.io/en/stable/usage/javascript.html

Sancus:


Sancus:
https://mfa-accounts.thunderbird.dev is updated and contains this patch

MelissaAutumn (approved):
wfm, thanks!

PR #264, opened 2025-09-05
[WIP] Hook up the User admin panel forms with Keycloak (and Stalwart)

MelissaAutumn:
This is a lot of code to hook up Add, Change, and Delete operations with Keycloak (and delete with Stalwart too.) 

It needs some clean up and some automated testing but it works locally. This should partially clear up #262

PR #1023, opened 2026-06-25
Fix custom-domain outbound mail not being DKIM-signed (signature id must match Stalwart signing expression)

mzeier:
## Problem

Outbound mail from **verified custom domains** in Thundermail is sent **without a DKIM signature**, so it fails DKIM and breaks DMARC alignment at receiving providers. The customer's DNS (CNAMEs -> `dkim.thunderhosted.com`, published public keys) is correct; the gap is server-side signature selection.

Reported via thunderbird/platform-infrastructure#591 (customer `theokell.com`).

## Root cause

`MailClient.create_dkim()` created Stalwart DKIM signatures with `'id': None`, so Stalwart auto-assigned an id that does **not** match its outbound signing expression:

```
if is_local_domain('*', sender_domain)
  then ['rsa-' + sender_domain, 'ed25519-' + sender_domain] else false
```

The expression selects signatures **by id** (`rsa-<domain>` / `ed25519-<domain>`). With a non-matching id, Stalwart finds no signature for the sender domain and sends the message unsigned. This affects **every** custom domain; the primary domain works only because its signatures were seeded with matching ids during server setup.

## Fix

Build the signature id as `'<prefix>-<normalized-domain>'` (e.g. `rsa-theokell.com`) via a new `STALWART_DKIM_ALGO_ID_PREFIXES` map, so created signatures match the signing expression. Domain is normalized (lowercase, no trailing dot) to match `sender_domain` at signing time.

- `settings.py`: add `STALWART_DKIM_ALGO_ID_PREFIXES = {'Rsa': 'rsa', 'Ed25519': 'ed25519'}`
- `mail/clients.py`: set `id` from `<prefix>-<normalized_domain>`
- `mail/tests/test_clients.py`: update expected payload ids

## Backfill required (not in this PR)

Signatures already created for existing custom domains have non-matching ids and will remain unsigned until recreated with correct ids (delete + recreate, or rename the `signature.<id>.*` keys). Recreation regenerates the keypair but is **customer-transparent**: customer CNAMEs point at our `dkim.thunderhosted.com` zone and `publish_hosted_dkim_dns_records` re-pushes the new public keys. Tracking in platform-infrastructure#591.

## Testing

- Updated `TestMailClientCreateDkim.test_success` to assert the constructed ids.
- Verified id construction + domain normalization in isolation (`_normalize_domain` is pure): `theokell.com`, uppercase, and trailing-dot inputs all yield `rsa-theokell.com` / `ed25519-theokell.com`.
- Full Django suite runs in the project's docker stack (CI); not run locally (local venv lacks `paddle_billing`).

## Post-merge verification

1. Add/re-provision a custom domain; confirm `signature.rsa-<domain>` and `signature.ed25519-<domain>` exist in Stalwart with the right selectors (tm1/tm2).
2. Send from `@<domain>` -> message has `DKIM-Signature` (`d=<domain>`, `s=tm1`/`tm2`); DKIM `pass`, DMARC aligned (learndmarc.com / mail-tester).

### DIFF
--- a/src/thunderbird_accounts/mail/clients.py
+++ b/src/thunderbird_accounts/mail/clients.py
@@ -257,11 +257,17 @@ class MailClient:
         Response objects may used for testing.
         Throws exception if request fails.
         """
+        from thunderbird_accounts.mail.dkim import _normalize_domain
+
         response_data = []
         dkim_algorithms = settings.STALWART_DKIM_ALGOS if algorithms is None else algorithms
+        normalized_domain = _normalize_domain(domain)
         for algorithm in dkim_algorithms:
+            # The signature id must match Stalwart's signing expression ('<prefix>-<domain>',
+            # e.g. rsa-example.com); otherwise outbound mail for this domain is sent unsigned.
+            id_prefix = settings.STALWART_DKIM_ALGO_ID_PREFIXES.get(algorithm)
             data = {
-                'id': None,
+                'id': f'{id_prefix}-{normalized_domain}' if id_prefix else None,
                 'algorithm': algorithm,
                 'domain': domain,
                 'selector': settings.STALWART_DKIM_ALGO_SELECTORS.get(algorithm),
--- a/src/thunderbird_accounts/mail/tests/test_clients.py
+++ b/src/thunderbird_accounts/mail/tests/test_clients.py
@@ -236,14 +236,14 @@ class TestMailClientCreateDkim(TestCase):
         self.assertEqual(
             [
                 {
-                    'id': None,
+                    'id': f'ed25519-{self.domain}',
                     'algorithm': 'Ed25519',
                     'domain': self.domain,
                     'selector': 'tm2',
                     'stage': DkimSignatureStage.PENDING.value,
                 },
                 {
-                    'id': None,
+                    'id': f'rsa-{self.domain}',
                     'algorithm': 'Rsa',
                     'domain': self.domain,
                     'selector': 'tm1',
--- a/src/thunderbird_accounts/settings.py
+++ b/src/thunderbird_accounts/settings.py
@@ -443,6 +443,14 @@ STALWART_DKIM_ALGO_SELECTORS = {
     'Rsa': 'tm1',
     'Ed25519': 'tm2',
 }
+# Stalwart's outbound signing expression selects signatures *by id*, e.g.
+#   if is_local_domain('*', sender_domain) then ['rsa-' + sender_domain, 'ed25519-' + sender_domain] else false
+# so each signature's id must be '<prefix>-<domain>' (e.g. rsa-example.com) or Stalwart finds
+# no matching signature for the sender domain and sends the message unsigned.
+STALWART_DKIM_ALGO_ID_PREFIXES = {
+    'Rsa': 'rsa',
+    'Ed25519': 'ed25519',
+}
 # Stalwart 0.15 does not expose the admin JMAP DKIM signature lifecycle API.
 STALWART_DKIM_STAGE_MANAGEMENT_ENABLED = os.getenv('STALWART_DKIM_STAGE_MANAGEMENT_ENABLED', '').lower() == 'true'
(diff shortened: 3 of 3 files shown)


src/thunderbird_accounts/mail/clients.py
```diff
@@ -257,11 +257,17 @@ def create_dkim(self, domain, stage: DkimSignatureStage = DkimSignatureStage.PEN
...
         response_data = []
         dkim_algorithms = settings.STALWART_DKIM_ALGOS if algorithms is None else algorithms
+        normalized_domain = _normalize_domain(domain)
         for algorithm in dkim_algorithms:
+            # The signature id must match Stalwart's signing expression ('<prefix>-<domain>',
+            # e.g. rsa-example.com); otherwise outbound mail for this domain is sent unsigned.
+            id_prefix = settings.STALWART_DKIM_ALGO_ID_PREFIXES.get(algorithm)
             data = {
-                'id': None,
+                'id': f'{id_prefix}-{normalized_domain}' if id_prefix else None,
```
MelissaAutumn, line 270: So @mzeier does this actually work? I remember there was some issues with v0.15 not actually saving something related to dkim.

MelissaAutumn (changes requested):

src/thunderbird_accounts/mail/clients.py
mzeier: Fair challenge, and I don't want to assume — here's what I dug up and what I think we should do before merging.

**On the `id: None` history:** it's not a deliberate v0.15 workaround. It's been `None` since the original Stalwart hookup (#264), so "let Stalwart assign the id" was just the default, not a known-good decision.

**On whether supplying an id works:** I couldn't prove it from source either way. v0.16 reorganized DKIM into a registry model — the old `crates/http/src/management/` is gone, and the modern path is a `Domain` registry object with a `selector_template` where Stalwart auto-generates *and auto-rotates* signatures and manages DNS itself (see `tests/src/automation/dkim.rs` @ v0.16.1). We're still using the **legacy** `POST /api/dkim` endpoint with explicit `tm1`/`tm2` selectors, and the automation tests don't exercise that legacy path with a client-supplied id. 

**So this needs an empirical check on staging before merge.** Concretely:
1. `POST /api/dkim` with `id = "rsa-<domain>"` for a test domain.
2. `GET` it back (list signatures) and confirm the id/name actually persisted as `rsa-<domain>` and wasn't overridden/auto-generated.
3. Send a test message from `@<domain>` and confirm a `DKIM-Signature` header appears (`d=<domain>`, `s=tm1`/`tm2`).

If step 2 shows Stalwart ignores the supplied id, this PR won't fix it and we pivot: discover what id Stalwart actually assigns and align the server-side signing expression to it instead (the `['rsa-' + sender_domain, 'ed25519-' + sender_domain]` expression lives in Stalwart config, not this repo) — or move to the newer registry/`selector_template` model entirely.

I'll hold this as a draft until we've run that check.

markstos (commented):
I reviewed this by first checking the Stalwart 0.15 docs  to see how this `id` parameter is used.

https://stalw.art/docs/0.15/api/management/endpoints/#create-dkim-signature

It is essentially undocumented, so I checked out the source code and reviewed the code.

In there, I found that when `id` is `none`, it defaults to the same values proposed in this PR, meaning this proposed change would have no impact.

Instead, the root cause with 0.15 appears to be that DKIM creation updates persistent config, while outbound signing uses a in-memory registry. 

What appears to be the fix is to make a `GET` call to the `/reload` management API, which syncs the persistent and in-memory state.

This method is lightly documented here:
https://stalw.art/docs/0.15/api/management/endpoints/#reload-settings

I tested the theory of the alternate fix locally with my own Stalwart instance, by making a DKIM update, then doing a pre-test of sending, a reload, and a post-test of sending. The test confirmed that the DKIM signature did not work until the reload. 

The hot fix is to restart our Stalwart instances ASAP. I'll work on a PR to automate the reloading and close this PR.

PR #1029, opened 2026-06-25
Put stale user's in the "Users to Purge" group instead of nuking them

### LINKED ISSUE #1028: Flag accounts for purge
Temp measure until we're confident on the task.

MelissaAutumn:
Fixes #1028 

I get or create a new group titled "Users to Purge" and users that would have been removed to that group instead. After the normal loop, we check to see if anyone in the group need to be removed (because they signed up.)

### DIFF
--- a/src/thunderbird_accounts/authentication/admin/models.py
+++ b/src/thunderbird_accounts/authentication/admin/models.py
@@ -97,6 +97,7 @@ class CustomUserAdmin(UserAdmin):
         'is_active',
         'is_awaiting_payment_verification',
         'plan',
+        'groups',
     ]
 
     form = CustomUserChangeForm
--- a/src/thunderbird_accounts/authentication/tasks.py
+++ b/src/thunderbird_accounts/authentication/tasks.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.models import Group
 import logging
 from datetime import timedelta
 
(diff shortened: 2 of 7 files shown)


src/thunderbird_accounts/authentication/admin/models.py
```diff
@@ -97,6 +97,7 @@ def shorten_plan(self, user: User) -> str | None:
         'is_active',
         'is_awaiting_payment_verification',
         'plan',
+        'groups',
     ]
```
MelissaAutumn, line 101: So we can filter users by group in the admin panel

src/thunderbird_accounts/authentication/tests/test_tasks.py
```diff
@@ -219,6 +220,96 @@ def test_no_ops_when_mailchimp_disabled(self, mock_client_cls):
         self.assertEqual(result['task_status'], 'skipped')
         self.assertEqual(result['tagged'], 0)
 
+# TEMP: Until #1028 is no longer needed
```
MelissaAutumn, line 223: Didn't want to nuke all of the og tests juuust yet.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -117,6 +119,17 @@ def purge_incomplete_signups(self):
...
+    # Fetch or create the "Users to purge" group
+    group, _ = Group.objects.get_or_create(name='Users to Purge')
+    if not group:
+        result = {
+            'task_status': 'failed',
+            'deleted': 0,
+            'errors': 0,
+            'skipped': 0,
+        }
+        return result
```
davinotdavid, line 131: According to the `get_or_create` there + tests added in this PR, I believe this should never happen, right?

davinotdavid (approved):
Thanks! LGTM

## QA Log

- Checked diff
- Ran tests locally
- Ran it manually locally with an eligible account and triggered the task manually

Image

kewisch (approved):

src/thunderbird_accounts/authentication/tasks.py
MelissaAutumn: Should never happen, but just in case let's not break things.

PR #1031, opened 2026-06-29
MFA UI FIxes for #1007, #1011, #1012

### LINKED ISSUE #1007: Hide "Loading multifactor authentication methods" unless it takes longer.
The "Loading multifactor authentication methods" message shouldn't be shown by default. If necessary we can show it if loading takes longer than N seconds, but so far it has been pretty instant.

### LINKED ISSUE #1011: Don't show dropdown for OTP methods if there is just one.
If there is just one authentication method, then we shouldn't be showing the dropdown. We already have the "Try another way" line as well, so maybe we default to the most secure mechanism and hide the rest behind "Try another way".

Image

### LINKED ISSUE #1012: Duplicate messaging when the authenticator code is wrong
There is duplicate messaging that the authenticator code is wrong.

Image

Sancus:
Preview is updated: https://mfa-accounts.thunderbird.dev

## Summary

Three commits:

### Skeleton MFA method state while loading (Fixes #1007)
Instead of text, we add a `hasLoaded` flag and render badge/button-sized shimmer skeletons until the first load resolves. This looks cleaner and eliminates any need to delay a loading message.

### Don't duplicate per-field errors in the global banner (Fixes #1012)
The theme template now honors Keycloak's `displayMessage` flag, so pages that surface a field-level error (e.g. `login-otp`) only show that error and not an info banner.

### Hide the OTP credential dropdown when there's only one (Fixes #1011)
The OTP login page always showed a device-selection dropdown, even for users with a single authenticator (nothing to choose). The selector is now gated on `userOtpCredentials.length > 1`, matching stock Keycloak's `userOtpCredentials?size gt 1` guard.

## Test plan

Verified on the preview environment (`https://mfa-accounts.thunderbird.dev`) signed in as an MFA-enrolled account.

**#1011 — OTP dropdown**
- [x] Sign in with a user that has a single authenticator → on the OTP page, confirm **no** device dropdown is shown.

**#1012 — duplicate error**
- [x] On the OTP page, enter an incorrect code → confirm a **single** inline error under the field and **no** duplicate global banner at the top.
image

**#1007 — loading state**
- [x] Enter a valid code and open **Manage MFA** → confirm methods load without a "Not set"/"Set up" flash and without a "Loading…" text flash.
- [x] Throttle/delay the `…/auth/mfa/methods/` request → confirm the badge and action render as shimmer skeletons (titles visible) and swap to the real state once the request resolves.

image

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -387,7 +387,6 @@
       "multiFactorAuthenticationDescription": "Multi-factor authentication (MFA) adds an extra layer of security by requiring more than a password to sign in. We recommend setting up at least two methods to avoid lockout. You can change or remove MFA methods anytime on this screen.",
       "multiFactorAuthenticationLearnMore": "Learn more about MFA",
       "authenticationMethods": "Authentication Methods",
-      "loading": "Loading multi-factor authentication methods...",
       "confirmRemoveAuthenticatorApp": "Remove authenticator app?",
       "confirmRemoveAuthenticatorAppDescription": "Your recovery codes will also be removed, and you will need to set up a new authenticator app before multi-factor authentication is enabled again.",
       "confirmRemoveRecoveryCodes": "Remove recovery codes?",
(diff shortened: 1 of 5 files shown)


Sancus:
Please rebase merge this when approved, each commit msg already has a reference to the issue and to this PR.

davinotdavid (approved):
Lgtm!

### QA Log

- Checked code diff
- Manually tested cases

Hide the OTP credential dropdown when there's only one
Image

Don't duplicate per-field errors in the global banner
Image

Skeleton MFA method state while loading
Image

PR #1041, opened 2026-06-29
End to End sign-up flow fixes

### LINKED ISSUE #986: E2E tests fail due to sign up changes

MelissaAutumn:
Fixes #986

There was a bit of backend work needed to make the tests work again. I've added an endpoint that's only accessible via test or superuser accounts that can create a test allow list entry.

This allow list entry is marked as a test/temp entry, and when a user is created from it that user is also marked as test/temp. Additionally I've added a cron that runs minutely to remove these accounts/entries that are over 5 minutes old. 

Originally I was trying to do this user cleanup on the test-side of things, but I found it to be extremely unreliable. (If a test crashes, it wouldn't clean up.) So the backend method seems to work great. 

As far as tests go, they're basic. We now sign-up to the point where we get a "verify account" email, and everything seems to be working smoothly. I did remove the mobile specific test because I couldn't tell what was different about it. 

Note: Since admin is the e2e test user locally this works without any effort (since admin@example.org is a superuser.) On stage/prod we'll have to give the permission to the e2e test accounts in order for these tests to be functional.


src/thunderbird_accounts/authentication/admin/models.py
```diff
@@ -177,3 +177,26 @@ def has_view_permission(self, request, obj=None):
...
+    def has_view_permission(self, request, obj=None):
+        """Only show view permissinos to users with superuser and staff flags"""
+        if not request.user:
+            return False
+        if not request.user.is_superuser or not request.user.is_staff:
+            return False
+        return True
+
+    def get_queryset(self, request):
+        return super().get_queryset(request).prefetch_related('content_type')
```
MelissaAutumn, line 202: This allows us to view permissions in the admin panel. It's a default django feature but they don't show the list in the admin panel by default.

src/thunderbird_accounts/authentication/api.py
```diff
@@ -203,6 +238,11 @@ def sign_up(request: Request):
             status=403,
         )
 
+    try:
+        validate_email(username, generic_email_error)
+    except EmailNotValidError:
+        return Response({'error': generic_email_error, 'type': 'bad-email'}, status=400)
+
```
MelissaAutumn, line 245: ...while sign up does this check on keycloak's end, let's save some effort by adding it in here too.

src/thunderbird_accounts/support/views.py
```diff
@@ -14,6 +14,13 @@
 @cache_page(60 * 15)
 def contact_fields(request: HttpRequest):
     """Get ticket fields from Zendesk API and filter based on Zendesk Admin."""
+
+    # Prevent incorrectly setup zendesk settings from crashing this route
+    if not settings.ZENDESK_SUBDOMAIN:
+        return JsonResponse(
+            {'success': False, 'error': _('Failed to fetch ticket fields')}, status=500
+        )
```
MelissaAutumn: This was noisy in the test

src/thunderbird_accounts/settings.py
```diff
@@ -612,20 +614,26 @@ def oidc_logout(request):
...
+        'schedule': crontab(
+            hour=int(os.getenv('INCOMPLETE_SIGNUP_PURGE_CRON_HOUR', '3')),
+            minute=int(os.getenv('INCOMPLETE_SIGNUP_PURGE_CRON_MINUTE', '0')),
+        ),
+    },
+    'purge-stale-test-allow-list-entries': {
+        'task': 'thunderbird_accounts.authentication.tasks.purge_stale_test_allow_list_entries',
+        'schedule': crontab(), # Every minute
+    }
+}
```
MelissaAutumn, line 629: Consolidated to make it cleaner

test/e2e/pages/tb-accts-oidc-page.ts
```diff
@@ -27,15 +27,19 @@ export class TBAcctsOIDCPage {
...
-  async signIn() {
-    console.log('signing in to TB Accounts');
-    expect(ACCTS_OIDC_EMAIL, 'getting ACCTS_OIDC_EMAIL env var').toBeTruthy();
-    expect(ACCTS_OIDC_PWORD, 'getting ACCTS_OIDC_PWORD env var').toBeTruthy();
-    await expect(this.emailInput).toBeVisible({ timeout: TIMEOUT_30_SECONDS });
-    await this.emailInput.fill(String(ACCTS_OIDC_EMAIL));
-    await this.page.waitForTimeout(TIMEOUT_2_SECONDS);
-    await this.passwordInput.fill(String(ACCTS_OIDC_PWORD));
-    await this.page.waitForTimeout(TIMEOUT_2_SECONDS);
+  async signIn(username: string | null = null, password: string | null = null) {
```
MelissaAutumn, line 30: I don't think I use this feature, but added a way for us to sign in to a different account.

MelissaAutumn: This will be useful for onboarding/ftue tests.

MelissaAutumn:
e2e tests are failing on contact form, which is not my problem! (well it IS but not in this ticket. :cry: )
https://github.com/thunderbird/thunderbird-accounts/issues/1045

test/e2e/tests/common/sign-up.spec.ts
```diff
@@ -1,74 +1,62 @@
...
-test.skip(true, 'Temporarily disabled due to issue 986');
+//test.skip(true, 'Temporarily disabled due to issue 986');
 
 test.beforeEach(async ({ page }) => {
-  // Remove any authentication cookies before passing it to our page
-  await page.context().clearCookies();
-  // Ensure we're cookieless.
-  expect(await page.context().cookies()).toEqual([])
-
   signUpPage = new TbAcctsSignUpPage(page);
```
rwood-moz: This was one of the differences in the tests/mobile/sign-up spec. Here you'll need to pass in the playwright project name:

` signUpPage = new TbAcctsSignUpPage(page, testInfo.project.name);`

So you'll have to add online 17:

`test.beforeEach(async ({ page }, testInfo) => {`

The reason we need this is because some of the methods (like `signUpPage.submitForm`) [need to check if running on android](https://github.com/thunderbird/thunderbird-accounts/blob/main/test/e2e/pages/tb-accts-signup-page.ts#L102).

test/e2e/tests/common/sign-up.spec.ts
```diff
@@ -1,74 +1,62 @@
...
-  await signUpPage.navigateToPage();
+  // We need to land on a page that we can stay on to retrieve a csrftoken.
+  await page.goto(ACCTS_CONTACT_URL);
 });
 
-test.describe('sign up form on desktop browser', {
+
+test.describe('sign up form on desktop browser meltime', {
   tag: [PLAYWRIGHT_TAG_E2E_SUITE, PLAYWRIGHT_TAG_E2E_PROD_DESKTOP_NIGHTLY],
```
rwood-moz: You'll need to import and add these playwright tags to the list please since the mobile test was removed, and we want this to run on mobile too:

 `PLAYWRIGHT_TAG_E2E_SUITE_MOBILE, PLAYWRIGHT_TAG_E2E_PROD_MOBILE_NIGHTLY`.

test/e2e/tests/common/sign-up.spec.ts
```diff
@@ -1,74 +1,62 @@
...
-
   signUpPage = new TbAcctsSignUpPage(page);
-  await signUpPage.navigateToPage();
+  // We need to land on a page that we can stay on to retrieve a csrftoken.
+  await page.goto(ACCTS_CONTACT_URL);
 });
 
-test.describe('sign up form on desktop browser', {
+
+test.describe('sign up form on desktop browser meltime', {
```
rwood-moz: If this test is being used for both mobile and deskop then the test name should no longer specify `desktop browser`, also did you want to remove `meltime`? :)

rwood-moz (changes requested):

test/e2e/pages/tb-accts-signup-page.ts
```diff
@@ -37,9 +34,59 @@ export class TbAcctsSignUpPage {
...
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRFToken': csrfToken,
+      },
+      data: JSON.stringify({
+        email
+      })
+    });
+    return await response.json();
+  }
```
rwood-moz, line 72: Should this be put in a `try/catch` so we can have a more graceful failure? If the request fails it results in:

```
  5) [firefox] › tests/common/sign-up.spec.ts:88:7 › sign up form on desktop browser meltime › step password fails with password mismatch @e2e-suite @e2e-prod-desktop-nightly 

    SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON

       at ../pages/tb-accts-signup-page.ts:66

      64 |       })
      65 |     });
    > 66 |     return await response.json();
```

rwood-moz (changes requested):

test/e2e/tests/common/sign-up.spec.ts
```diff
@@ -95,13 +87,46 @@ test.describe('sign up form on desktop browser', {
...
+
+    // tb.pro may insert a locale prefix (e.g. /en-US/waitlist/) when redirecting
+    await page.waitForLoadState('domcontentloaded');
+    const waitListOrigin = new URL(TB_PRO_WAIT_LIST_URL).origin;
+    await page.waitForURL(
+      url => url.origin === waitListOrigin
+        && url.pathname.endsWith('/waitlist/')
+        && url.searchParams.get('email') === testEmail,
+    );
+  });
```
rwood-moz, line 128: Hmmm when I run the tests locally this one fails, instead of being directed to the allow list, the `Confirm Plan` screen/step is displayed. Not sure why, when you are specifying that the user not be on the allow list.

```
TimeoutError: page.waitForURL: Timeout 30000ms exceeded.
=========================== logs ===========================
waiting for navigation until "load"
  navigated to "http://localhost:8087/sign-up?email=1291e11540da457789b6b0d52ce62eac%40example.com"
============================================================

  110 |     await page.waitForLoadState('domcontentloaded');
  111 |     const waitListOrigin = new URL(TB_PRO_WAIT_LIST_URL).origin;
> 112 |     await page.waitForURL(
      |                ^
  113 |       url => url.origin === waitListOrigin
  114 |         && url.pathname.endsWith('/waitlist/')
  115 |         && url.searchParams.get('email') === testEmail,
    at /Users/rwood/dev/accounts/test/e2e/tests/common/sign-up.spec.ts:112:16
```

MelissaAutumn:
Thanks for the review and pointing out the mobile change! I think I have an idea to keep them consolidated.

test/e2e/tests/common/sign-up.spec.ts
MelissaAutumn: Ah I think it's just a config issue. I'll add back the test.skip line, and a message of what var you need to flip.

test/e2e/tests/common/sign-up.spec.ts
MelissaAutumn: naw, it's always meltime. (Yes, thanks for catching that it was for grepping.)

rwood-moz: > naw, it's always meltime. (Yes, thanks for catching that it was for grepping.)

I do the same (well not `meltime`) haha

MelissaAutumn:
Due to some difficulty I cannot test android locally (running adb likes to constantly kill and reconnect my usb hub.) So I'd like to see this running on stage.

rwood-moz:
> Due to some difficulty I cannot test android locally (running adb likes to constantly kill and reconnect my usb hub.) So I'd like to see this running on stage.

Thank you for the updates @MelissaAutumn! Awesome sign-up test, and it passes for me locally now. The only way to test this on android is on BrowserStack; however we cannot run tests on a local dev stack on BrowserStack. If this PR just changed the E2E tests we could run the tests on this branch on stage on BrowserStack android, but since it also has product-side changes we cannot do that; we'll have to land this in main and then test it out on BrowserStack android on stage after that.

rwood-moz (approved):
Reviewed the code and ran the E2E tests locally (with USE_ALLOW_LIST=True in backend .env). The sign-up test runs and great and passes. The contact test is failing but that is unrelated (and failing on main too).

rwood-moz:
> Note: Since admin is the e2e test user locally this works without any effort (since [admin@example.org](mailto:admin@example.org) is a superuser.) On stage/prod we'll have to give the permission to the e2e test accounts in order for these tests to be functional.

@MelissaAutumn When you have a chance could you please give the permission required for the e2e test accounts? I tried to run in stage but see the test cannot create the allow list entry. I'll DM you the stage and prod accts, thanks 😄

rwood-moz:
> @MelissaAutumn When you have a chance could you please give the permission required for the e2e test accounts? I tried to run in stage but see the test cannot create the allow list entry. I'll DM you the stage and prod accts, thanks 😄

Update: Mel updated the test accounts on stage, and after another change (#1052) the new sign-up test is running successfully now on stage on [Firefox desktop in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Firefox+Desktop/238?tab=sessions&testListView=spec). However unfortunately the test is failing on Android Chrome, probably something specific to an Android locator. I've filed #1055  to look into that.

PR #1047, opened 2026-06-29
Cover zendesk's request.id retrieval with a try/except.

MelissaAutumn:
Related #1045 

Also fix some test imports.

### DIFF
--- a/src/thunderbird_accounts/support/tests/test_views.py
+++ b/src/thunderbird_accounts/support/tests/test_views.py
@@ -117,7 +117,7 @@ class ZendeskContactSubmitTestCase(TestCase):
         self.client = RequestClient()
 
     @patch('thunderbird_accounts.support.views.ZendeskClient')
-    @patch('thunderbird_accounts.core.utils.parse_user_agent_info')
+    @patch('thunderbird_accounts.support.views.parse_user_agent_info')
     @override_settings(
         ZENDESK_FORM_ID='42',
         ZENDESK_FORM_BROWSER_FIELD_ID='1001',
@@ -278,7 +278,7 @@ class ZendeskContactSubmitTestCase(TestCase):
         instance.update_ticket.assert_not_called()
 
     @patch('thunderbird_accounts.support.views.ZendeskClient')
-    @patch('thunderbird_accounts.core.utils.parse_user_agent_info')
+    @patch('thunderbird_accounts.support.views.parse_user_agent_info')
     @override_settings(
         ZENDESK_FORM_ID='42',
         ZENDESK_FORM_BROWSER_FIELD_ID='1001',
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/support/tests/test_views.py
```diff
@@ -117,7 +117,7 @@ def setUp(self):
         self.client = RequestClient()
 
     @patch('thunderbird_accounts.support.views.ZendeskClient')
-    @patch('thunderbird_accounts.core.utils.parse_user_agent_info')
+    @patch('thunderbird_accounts.support.views.parse_user_agent_info')
```
MelissaAutumn, line 120: If we import something at the top of the file we're basically bringing it into that module (think of it as bringing in a function as a variable/callable.) 

So if we're mocking we need to mock the import not the original source.

src/thunderbird_accounts/support/views.py
```diff
@@ -9,6 +10,8 @@
 
 from thunderbird_accounts.support.zendesk import ZendeskClient
 
+# Add browser and OS information to hidden custom fields
+from thunderbird_accounts.core.utils import parse_user_agent_info
```
MelissaAutumn, line 14: Moved this out of the middle of the function below.

src/thunderbird_accounts/support/views.py
```diff
@@ -168,14 +171,24 @@ def contact_submit(request: HttpRequest):
...
+        sentry_sdk.set_context('exception_data', {
+            'ex': ex,
+            'zendesk_api_response': zendesk_api_response,
+            'response_data': response_data,
+        })
+        sentry_sdk.capture_message(
+            f'Failed to create Zendesk ticket: {zendesk_api_response}',
+            level='error',
+        )
+        return JsonResponse({'success': False}, status=500)
```
MelissaAutumn, line 187: I need more debug info, so this isn't a fix more like a cover and see what happens sort of PR.

src/thunderbird_accounts/support/views.py
```diff
@@ -168,14 +171,24 @@ def contact_submit(request: HttpRequest):
...
 
-    # Add browser and OS information to hidden custom fields
-    from thunderbird_accounts.core.utils import parse_user_agent_info
 
     user_agent_string = request.headers.get('User-Agent')
-    browser_string, os_string = parse_user_agent_info(user_agent_string)
+    browser_string, os_string = parse_user_agent_info(user_agent_string or '')
```
MelissaAutumn, line 191: silencing lsp

davinotdavid (approved):

src/thunderbird_accounts/support/tests/test_views.py
davinotdavid: Oh my bad, that was probably me here! Thanks for catching this 🙏

PR #1054, opened 2026-07-02
Update oidc middleware with two new feature flags that introspect and allow posting

### LINKED ISSUE #380: Add middleware to introspect current access token
With #210 we'll want the ability to revoke logged in sessions. However you can only logout a session with a request object.

We should add a middleware after the mozilla oidc middleware that checks if a user is auth, and if so introspect keycloak. If the call is successful and the token is good, cache the result for 3-5 mins. 

If its not then call djangos logout with the current request.

This will give a 3-5 minute window where a revoked session could be used, but I think thats okay.

### LINKED ISSUE #1053: Allow refresh token to be used if the request is not GET
If we have a refresh token and the access token is currently expired we should be able to request a new access token in POST requests. 

If we don't have a refresh token or that request fails, then the POST (or whichever) request should also fail (On get requests you'll be hit with the redirect to login.)

MelissaAutumn:
Fixes #380 (Add middleware to introspect current access token)
Fixes #1053 (Allow refresh token to be used if the request is not GET)

* `auth-introspect-token-per-request` forces the middleware to introspect the token every request
* `auth-allow-post-reauth` allows the middleware to attempt to reauth on non-get requests

I've updated the middleware acquired from a mozilla-django-oidc PR to allow refreshing tokens on non-GET requests and making the middleware introspect the access/refresh token. These are both gated behind the above feature flags. 

To test add the feature flags listed above via [django admin panel](http://localhost:8087/admin/waffle/flag/) and set them to true. With a short-lived access token you should hopefully experience a better experience without being bumped to the login screen. Additionally if you invalidate your token (via password reset) it should pick up on it.

An easy way to test this is to remove the active session for the currently logged in user on keycloak's admin panel. (http://keycloak:8999/admin/master/console/#/tbpro/users/). On a refresh that user should be punted to the login screen with a "refresh token is not active" debug log entry listed in the console.

This code is kinda too complex imo, suggestions for streamlining while keeping it secure are welcome. I wanted to let us roll these changes back easily if needed, so that's why they're waffled.

### DIFF
--- a/assets/app/vue/composables/useFetch.ts
+++ b/assets/app/vue/composables/useFetch.ts
@@ -1,0 +1,34 @@
+import { createFetch } from "@vueuse/core"
+
+//const { isFetching, error, data } = useMyFetch('users')
+export const useAuthfulFetch = createFetch({
+  options: {
+    async beforeFetch(ctx) {
+      // If we don't explicitly set an accept header, make sure it's set to application/json
+      if (!ctx.options?.headers['Accept']) {
+        ctx.options.headers['Accept'] = 'application/json';
+      }
+      return ctx;
+    },
+    async onFetchError(ctx) {
+      // Try and handle reauth responses (this happens if the refresh token is also invalid.)
+      try {
+        const data = JSON.parse(ctx.data);
+        // If it's a 403 we need to ship them to auth
+        if (ctx.response.status === 403 && data?.type === 'auth_error') {
+          if (data?.refresh_url) {
+            window.location = data?.refresh_url;
+            return ctx;
+          }
+        }
+      } catch {
+        // ...
+      } 
+
+      return ctx;
+    },
+  },
+  fetchOptions: {
+    mode: 'cors',
+  },
+});
--- a/assets/app/vue/composables/useFetch.ts
+++ b/assets/app/vue/composables/useFetch.ts
@@ -1,7 +1,7 @@
 import { createFetch } from "@vueuse/core"
 
 //const { isFetching, error, data } = useMyFetch('users')
-export const useAuthfulFetch = createFetch({
+export const useAuthFetch = createFetch({
   options: {
     async beforeFetch(ctx) {
       // If we don't explicitly set an accept header, make sure it's set to application/json
--- a/assets/app/vue/composables/useFetch.ts
+++ b/assets/app/vue/composables/useFetch.ts
@@ -1,30 +1,32 @@
-import { createFetch } from "@vueuse/core"
+import { createFetch } from '@vueuse/core';
 
-//const { isFetching, error, data } = useMyFetch('users')
-export const useAuthFetch = createFetch({
+interface AuthenticationErrorResponse {
+  error?: string;
+  type?: string;
+  refresh_url?: string;
+}
+
+export class AuthRedirectError extends Error {
+  constructor() {
+    super('Redirecting to authentication');
+    this.name = 'AuthRedirectError';
+  }
+}
+
+export class AuthSessionError extends Error {
+  constructor(message = 'The authentication session has expired') {
+    super(message);
+    this.name = 'AuthSessionError';
+  }
+}
+
+const useAuthFetchRequest = createFetch({
   options: {
     async beforeFetch(ctx) {
-      // If we don't explicitly set an accept header, make sure it's set to application/json
-      if (!ctx.options?.headers['Accept']) {
+      // JSON auth failures include a refresh URL instead of redirecting the fetch request.
+      if (!ctx.options.headers['Accept']) {
         ctx.options.headers['Accept'] = 'application/json';
       }
-      return ctx;
-    },
-    async onFetchError(ctx) {
-      // Try and handle reauth responses (this happens if the refresh token is also invalid.)
-      try {
-        const data = JSON.parse(ctx.data);
-        // If it's a 403 we need to ship them to auth
-        if (ctx.response.status === 403 && data?.type === 'auth_error') {
-          if (data?.refresh_url) {
-            window.location = data?.refresh_url;
-            return ctx;
-          }
-        }
-      } catch {
-        // ...
-      } 
-
       return ctx;
     },
   },
@@ -32,3 +34,33 @@ export const useAuthFetch = createFetch({
     mode: 'cors',
   },
 });
+
+export const useAuthFetch = async (url: string, options: RequestInit = {}) => {
+  const result = await useAuthFetchRequest(url, options);
+  const response = result.response.value;
+  if (!response) {
+    throw new Error(result.error.value || 'Request failed');
+  }
+
+  // VueUse stores HTTP errors in reactive state, so session failures must be handled explicitly.
+  if (response.status !== 403) {
+    return result;
+  }
+
+  const data = (await response
+    .clone()
+    .json()
+    .catch(() => null)) as AuthenticationErrorResponse | null;
+  if (data?.type !== 'auth_error') {
+    // Feature-specific 403 responses, including MFA reauthentication, stay with their callers.
+    return result;
+  }
+
+  if (data.refresh_url) {
+    window.location.assign(data.refresh_url);
+    // Prevent the caller from consuming the 403 after navigation starts.
+    throw new AuthRedirectError();
+  }
+
+  throw new AuthSessionError(data.error);
+};
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
@@ -1,7 +1,7 @@
-import { useAuthfulFetch } from "@/composables/useFetch";
+import { useAuthFetch } from "@/composables/useFetch";
 
 export const addCustomDomain = async (domainName: string) => {
-  const { response } = await useAuthfulFetch(`/custom-domains/add`, {
+  const { response } = await useAuthFetch(`/custom-domains/add`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
@@ -14,12 +14,12 @@ export const addCustomDomain = async (domainName: string) => {
 };
 
 export const getRemoteDNSRecords = async (domainName: string) => {
-  const response = await useAuthfulFetch(`/custom-domains/dns-records?domain-name=${domainName}`);
-  return await response.json();
+  const { response } = await useAuthFetch(`/custom-domains/dns-records?domain-name=${domainName}`);
+  return await response.value.json();
 };
 
 export const verifyDomain = async (domainName: string) => {
-  const { response } = await useAuthfulFetch(`/custom-domains/verify`, {
+  const { response } = await useAuthFetch(`/custom-domains/verify`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
@@ -32,7 +32,7 @@ export const verifyDomain = async (domainName: string) => {
 };
 
 export const removeCustomDomain = async (domainName: string) => {
-  const { response } = await useAuthfulFetch(`/custom-domains/remove`, {
+  const { response } = await useAuthFetch(`/custom-domains/remove`, {
     method: 'DELETE',
     headers: {
       'Content-Type': 'application/json',
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
@@ -1,5 +1,7 @@
+import { useAuthfulFetch } from "@/composables/useFetch";
+
 export const addCustomDomain = async (domainName: string) => {
-  const response = await fetch(`/custom-domains/add`, {
+  const { response } = await useAuthfulFetch(`/custom-domains/add`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
@@ -8,16 +10,16 @@ export const addCustomDomain = async (domainName: string) => {
     body: JSON.stringify({ 'domain-name': domainName }),
   });
 
-  return await response.json();
+  return await response.value.json();
 };
 
 export const getRemoteDNSRecords = async (domainName: string) => {
-  const response = await fetch(`/custom-domains/dns-records?domain-name=${domainName}`);
+  const response = await useAuthfulFetch(`/custom-domains/dns-records?domain-name=${domainName}`);
   return await response.json();
 };
 
 export const verifyDomain = async (domainName: string) => {
-  const response = await fetch(`/custom-domains/verify`, {
+  const { response } = await useAuthfulFetch(`/custom-domains/verify`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
@@ -26,11 +28,11 @@ export const verifyDomain = async (domainName: string) => {
     body: JSON.stringify({ 'domain-name': domainName }),
   });
 
-  return await response.json();
+  return await response.value.json();
 };
 
 export const removeCustomDomain = async (domainName: string) => {
-  const response = await fetch(`/custom-domains/remove`, {
+  const { response } = await useAuthfulFetch(`/custom-domains/remove`, {
     method: 'DELETE',
     headers: {
       'Content-Type': 'application/json',
@@ -38,5 +40,5 @@ export const removeCustomDomain = async (domainName: string) => {
     },
     body: JSON.stringify({ 'domain-name': domainName }),
   });
-  return await response.json();
+  return await response.value.json();
 };
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
@@ -1,4 +1,4 @@
-import { useAuthfulFetch } from "@/composables/useFetch";
+import { useAuthFetch } from "@/composables/useFetch";
 
 type AddCustomDomainResponse = {
   success: boolean;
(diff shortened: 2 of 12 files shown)


src/thunderbird_accounts/settings.py
```diff
@@ -700,3 +698,12 @@ def oidc_logout(request):
...
 TEST_ALLOW_LIST_ENTRIES_STALE_TIME_IN_MINS = 5
+
+# Waffle flag constants
+# These still have to be created on each environment to be True.
+
+# POST OIDC's token introspect route to see if a token is actually active.
+WAFFLE_FLAG_INTROSPECT_TOKEN_PER_REQUEST = 'auth-introspect-token-per-request'
+
+# Allow auth middleware to try and refresh an inactive access token with an active refresh token during non-get requests
+WAFFLE_FLAG_ALLOW_POST_REAUTH = 'auth-allow-post-reauth'
```
davinotdavid, line 719: I think it would be nice to add a section in README for the flags, also I just realized that we don't have an entry for MFA either (which I believe is `multi-factor-authentication`)

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -372,9 +474,10 @@ def finish(self, request, prompt_reauth=True):
         middleware doesn't really want the user in if they don't
         refresh their session.
         """
-        default_response = None
+        # is_redirectable = request.method == 'GET'
+        default_response = HttpResponseForbidden()
         xhr_response_json = {'error': 'the authentication session has expired'}
-        if prompt_reauth:
+        if prompt_reauth:  # and is_redirectable:
```
davinotdavid: Is the `is_redirectable` a left over for debugging here? If so I think we should remove it to reduce confusion

davinotdavid (commented):
Tested locally with the test steps in the PR description and can verify that removing the session in Keycloak and refreshing the page punts the user to the login screen, nice!

However, if I don't refresh the page but continue to try to click around / trigger some requests, they fail the requests instead of punting the user to the login screen. Is that expected?

Not a super realistic scenario but for example if you:

- Add a user to the allow list and go through sign-up / FTUE
- Stay in the ToS page and in another tab / window / browser revoke access
- As the user in ToS page, click on accept or decline and see it failing and not punting the user to login

Repeating the process but in the Paddle checkout page:

- Add a user to the allow list and go through sign-up / FTUE
- Go through ToS and stay in the Paddle checkout page
- In another tab / window / browser revoke access of said user
- As the user in Paddle checkout page, click on the UserAvatar and then Support and then click Thunderbird Pro to go back to the Paddle checkout. See that it fails to load and doesn't punt the user to login

src/thunderbird_accounts/settings.py
MelissaAutumn: Good point, I'll add a section in docs because readme is getting crowded.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: Oops, yep. Thanks!

MelissaAutumn:
Thanks for testing, and yeah we should handle those. Will need to setup useFetch and have a default 302 redirect follower for that. Not too hard, but may end up with a larger change overall.

MelissaAutumn:
Okay it turns out there was a bug where the refresh token was never used unless it's invalid lol. I fixed that flipped condition, so now access tokens should silently refresh correctly.

I've introduced a useFetch (much like in appointment) and sprinkled it around api calls that require auth. It's called useAuthfulFetch but I'm willing to rename it if anyone has a better suggestion. I wanted to convey that this function is for authed api calls only.

If the access token and refresh token are invalid (session invalidation iirc) then the user will be punted to the login screen via the fetch error function. In the future I think we should block page access with a modal and text explaining that they need to sign back in, but that feels like a services-ui pattern and we'll need to borrow some designers I think.

Sancus:
will go through this in the morning

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -35,21 +39,52 @@
...
+        'token': token,
+    }
+
+    req_auth = None
+    if import_from_settings('OIDC_TOKEN_USE_BASIC_AUTH', False):
+        # Basic-auth clients send credentials in the header, not the body.
+        req_auth = HTTPBasicAuth(token_payload['client_id'], token_payload.pop('client_secret'))
+
+    response = requests.post(
+        f'{import_from_settings("OIDC_OP_TOKEN_ENDPOINT")}introspect',
```
Sancus: I think this causes an issue with the lack of trailing slash on the stage config value https://github.com/thunderbird/thunderbird-accounts/blob/8ba5ec154f51f4bbbcc8196979e3dc48b357faef/pulumi/config.stage.yaml#L44

Though that probably should be made consistent...

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -35,21 +39,52 @@
...
+        auth=req_auth,
+        data=token_payload,
+        verify=import_from_settings('OIDC_VERIFY_SSL', True),
+    )
+    response.raise_for_status()
```
Sancus, line 68: I guess we're just passing whatever errors we get from keycloak if any? That seems a little suboptimal but keycloak errors do kinda break everything anyway I think.

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -372,19 +478,17 @@ def finish(self, request, prompt_reauth=True):
         middleware doesn't really want the user in if they don't
         refresh their session.
         """
-        default_response = None
-        xhr_response_json = {'error': 'the authentication session has expired'}
+        default_response = HttpResponseForbidden()
```
Sancus, line 473: This means *any* request that passes through this middleware gets 403 by default, on any sort of timeout or error, right?

assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
```diff
@@ -8,16 +10,16 @@ export const addCustomDomain = async (domainName: string) => {
     body: JSON.stringify({ 'domain-name': domainName }),
   });
 
-  return await response.json();
+  return await response.value.json();
 };
 
 export const getRemoteDNSRecords = async (domainName: string) => {
-  const response = await fetch(`/custom-domains/dns-records?domain-name=${domainName}`);
+  const response = await useAuthfulFetch(`/custom-domains/dns-records?domain-name=${domainName}`);
```
Sancus: This one's from fable:

getRemoteDNSRecords() returns the wrong object, VueUse’s fetch API does not have the same shape as native fetch().

The affected function effectively does:

const response = await useAuthfulFetch(url)
return response.json()

Here, response is the VueUse UseFetchReturn wrapper. Its .json() method configures parsing and returns another fetch wrapper; it is not the native Response.json() method.

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -314,14 +388,39 @@ def is_expired(self, request):
...
+
+        # Only introspect the token if the local exp claim is expired (and the feature flag is enabled.)
+        if not is_expired and self.enable_per_request_introspect:
+            if _is_token_active(access_token):
+                self.set_exit_state(self.EXIT_STATES.ACCESS_TOKEN_IS_ACTIVE)
+                logging.debug('access token introspect is active')
+                return
+
+            # If the refresh token does not exist, or if it's not active then we need a reauth
+            refresh_token_is_active = _is_token_active(refresh_token, TokenHintType.REFRESH_TOKEN)
```
Sancus: What's the purpose of this line? This causes an additional request I think, but if the refresh token is bad/nonexistent, then the refresh request will fail anyway won't it?

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -301,9 +336,48 @@ class OIDCRefreshSession(SessionRefresh):
...
+        This is only used in testing."""
+
+        NOT_REFRESHABLE = 'not-refreshable'
+        NOT_EXPIRED = 'not-expired'
+        ACCESS_TOKEN_IS_ACTIVE = 'access-token-is-active'
+        TOKEN_IS_STORED = 'token-is-stored'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.enable_per_request_introspect = False
```
Sancus: Does this work? Django reuses the middleware instance for more than one request doesn't it? Won't this leak to other requests for other users?

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -301,9 +336,48 @@ class OIDCRefreshSession(SessionRefresh):
...
+        self.allow_post_reauth = False
+        self.exit_state = None
+
+    def set_exit_state(self, state: EXIT_STATES):
```
Sancus: I don't particularly have the mental energy to make a better suggestion, but it is kinda odd to have states that are only for testing in prod code.

assets/app/vue/composables/useFetch.ts
```diff
@@ -0,0 +1,34 @@
+import { createFetch } from "@vueuse/core"
+
+//const { isFetching, error, data } = useMyFetch('users')
+export const useAuthfulFetch = createFetch({
```
Sancus: This stuff seems kinda rough to me in the sense I'm surprised we need bespoke logic for this. But seems like there aren't a lot of better options for our setup.

I asked Fable about this one and it did suggest that this should be a browser-native fetch wrapper instead, so I'll leave that suggestion here. I'm not familiar enough with Vue to say conclusively whether it makes sense as VueUse or not though.

```typescript
/** Thrown after we've started navigating to reauthentication;
 *  swallow this in the global error handler. */
export class AuthRedirectError extends Error {
  constructor() {
    super('Redirecting to authentication');
    this.name = 'AuthRedirectError';
  }
}

interface AuthenticationError {
  type?: string;
  refresh_url?: string;
}

export async function authFetch(
  input: RequestInfo | URL,
  init: RequestInit = {},
): Promise<Response> {
  const headers = new Headers(init.headers);

  if (!headers.has('Accept')) {
    headers.set('Accept', 'application/json');
  }

  const response = await fetch(input, { ...init, headers });

  if (response.status === 403) {
    const body = (await response
      .clone()
      .json()
      .catch(() => null)) as AuthenticationError | null;

    if (body?.type === 'auth_error' && body.refresh_url) {
      window.location.assign(body.refresh_url);
      throw new AuthRedirectError();
    }
  }

  return response;
}
```

Call sites then stay normal:
```typescript
const response = await authFetch(url);
return response.json();
```

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -35,21 +39,52 @@
 OIDC_REFRESH_TOKEN_KEY = 'oidc_refresh_token'
 
 
+class TokenHintType(StrEnum):
+    ACCESS_TOKEN = 'access_token'
+    REFRESH_TOKEN = 'refresh_token'
+
+
+def _is_token_active(token: str, token_hint: TokenHintType = TokenHintType.ACCESS_TOKEN) -> bool:
```
Sancus: This function feels a little redundant with `_request_refreshed_tokens(refresh_token):` ? Like they should probably be one function since they do a lot of the same things?

src/thunderbird_accounts/authentication/middleware.py
```diff
@@ -372,19 +478,17 @@ def finish(self, request, prompt_reauth=True):
...
             refresh_url = self._prepare_reauthorization(request)
             default_response = HttpResponseRedirect(refresh_url)
             xhr_response_json['refresh_url'] = refresh_url
 
-        if request.headers.get('x-requested-with') == 'XMLHttpRequest':
-            xhr_response = JsonResponse(xhr_response_json, status=403)
-            if 'refresh_url' in xhr_response_json:
-                xhr_response['refresh_url'] = xhr_response_json['refresh_url']
-            return xhr_response
+        if (request.headers.get('x-requested-with') == 'XMLHttpRequest' or
```
Sancus: Feels like each of these checks(this and the one below) should actually be a little helper function?

Sancus (changes requested):
This PR isn't really that long but man some of the logic in this broke my brain a little. Maybe I'm getting old or AI-brained. :)

assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts
MelissaAutumn: Thanks I missed a const { response } instead of response. I hate how VueUse's fetch API differs, but it provides an easy way to wrangle global fetch logic.

MelissaAutumn:
> This PR isn't really that long but man some of the logic in this broke my brain a little. Maybe I'm getting old or AI-brained. :)

Most likely my brain melted from active migraines :melting_face:. Thanks for the review, excited to fix the heck out of it.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: It's an unfortunate side effect of django middleware returning none or a response. I've changed the code to stuff it inside a session entry instead of the middleware itself.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: No it doesn't, thanks for checking.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: Yeah, you're right here we can nix this one.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: Only requests that aren't valid in some way. I'm cleaning up the function to remove a lot of confusing branching.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: I don't want to refresh the token every single request, so if the token doesn't appear to be expired we simply see if it's still active. If it's not active (or is expired via claim) then we request a refresh.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: Django actually has a request.accepts('application/json'). The x-requested-with came with the og function :shrug:

src/thunderbird_accounts/authentication/tests/test_middleware.py
```diff
@@ -112,84 +113,89 @@ def test_introspect_does_happen_if_feature_flag_is_set(self, mock_post: MagicMoc
...
+            OIDC_REFRESH_TOKEN_KEY: 'abc456',
         }
+        original_access_token = session.get(OIDC_ACCESS_TOKEN_KEY)
 
         mock_post.side_effect = [
             # Access token check
             SimpleNamespace(raise_for_status=lambda: None, json=lambda: {'active': False}),
-            # Refresh token check
-            SimpleNamespace(raise_for_status=lambda: None, json=lambda: {'active': True}),
+            # Refreshing access token
```
MelissaAutumn, line 134: Not sure how this test worked before as the refresh token would have been okay. Will be adding more tests later.

src/thunderbird_accounts/authentication/middleware.py
MelissaAutumn: I've moved the calling of this function into the giant try/except.

assets/app/vue/composables/useFetch.ts
MelissaAutumn: I think useFetch is fine, we use it in appointment. The issue it wraps fetch calls with vue reactivity which we're not using. It's a little clunky solution, but I'd rather use a well-tested one then make our own wrapper.

markstos (commented):
I think the two concerns could be separable-- the goal to handle logouts more smoothly and dealing with stale access tokens during POSTS.

## Improving logout handling

Regarding improving logout handling-- the native single-logout (SLO) handling of Keycloak seems like a more elegant way to handle this.
In Keycloak, in the Realm we use for Accounts, under Logout settings there is an option to provide a Backchannel logout URL and an option to pass the session ID.

On the Django side, we would need to start tracking the Keycloak session ID (sid) alongside our Django session key so we can match them. With SLO, this closes the 5 minute gap of potential staleness-- "accounts" gets notified as soon as the session is logged out or revoked on Keycloak.

Image

Was SLO considered as alternate approach here?

## Handling expired access tokens during non-GET

This logic update looks useful.

MelissaAutumn:
> I think the two concerns could be separable-- the goal to handle logouts more smoothly and dealing with stale access tokens during POSTS.
> ## Improving logout handling
> 
> Regarding improving logout handling-- the native single-logout (SLO) handling of Keycloak seems like a more elegant way to handle this. In Keycloak, in the Realm we use for Accounts, under Logout settings there is an option to provide a Backchannel logout URL and an option to pass the session ID.
> 
> On the Django side, we would need to start tracking the Keycloak session ID (sid) alongside our Django session key so we can match them. With SLO, this closes the 5 minute gap of potential staleness-- "accounts" gets notified as soon as the session is logged out or revoked on Keycloak.
> Image
> 

Good mention there, this was a stop-gap to fix incorrect session invalidation issues. The back channel is a bigger lift and will be done, I cannot find a ticket so I'll cut it now. Thanks! 

https://github.com/thunderbird/thunderbird-accounts/issues/1158

markstos (approved):
All feedback appears addressed.

Sancus:
There's a couple things I need to look at still, don't merge this yet.

PR #1056, opened 2026-07-03
Add polling time when verifying test has proceeded to the next sign-up step

### LINKED ISSUE #1055: Sign-up E2E test fails on Android Chrome
The new sign-up E2E test (landed in #1041) passes on Firefox desktop but fails on Android Chrome (BrowserStack link[ example failure 1](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Android+Chrome/24?bls_localTimezone=true&testListView=spec), [example failure 2](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Android+Chrome/26?tab=sessions&testListView=spec)). Probably a locator specific to Android or something like that; investigate and update the test.

rwood-moz:
## What changed?
The new sign-up test was intermittently failing when running on Android Chrome. After investigating I noticed that the test is moving forward to the next sign-up step but not giving enough time for the next step to actually appear.

## Why?
When submitting the current sign-up step's form (i.e. clicking the 'setup account' or 'next' button to proceed to the next step) the test was immediately checking that the next step was active, but intermittently this wasn't giving enough time and the screen for the same step was still displayed / the next step's screen wasn't displayed yet, causing the test to fail.

The fix is to add a timeout in the polling that verifies we are on the next sign-up step, so it gives it a chance to proceed. Also adding some logging so we can see in the logs what button is actually being pressed when the test moves on to the next sign-up step, makes debugging easier.

## Applicable Issues
Fixes #1055.

## QA Log
Ran the sign-up test in BrowserStack on [Firefox desktop](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Firefox+Desktop/239?testListView=spec) and several times on [Android Chrome](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Tests+Android+Chrome/31?tab=sessions&testListView=spec) and didn't see the failure anymore.

### DIFF
--- a/test/e2e/pages/tb-accts-signup-page.ts
+++ b/test/e2e/pages/tb-accts-signup-page.ts
@@ -1,5 +1,5 @@
 import { type Page, type Locator, expect } from '@playwright/test';
-import { ACCTS_HUB_URL, ACCTS_SIGN_UP_URL } from '../const/constants';
+import { ACCTS_HUB_URL, ACCTS_SIGN_UP_URL, TIMEOUT_10_SECONDS } from '../const/constants';
 
 export class TbAcctsSignUpPage {
   readonly page: Page;
@@ -105,9 +105,15 @@ export class TbAcctsSignUpPage {
 
     await this.submitForm();
 
-    await expect.poll(async () => {
-      return await this.stepId.inputValue();
-    }).toBe('step-username');
+    await expect.poll(
+      async () => {
+        return await this.stepId.inputValue();
+      },
+      {
+        timeout: TIMEOUT_10_SECONDS,
+        message: 'waiting for the next step (username) to appear',
+      }
+    ).toBe('step-username');
   }
 
   /**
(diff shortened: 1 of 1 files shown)


test/e2e/pages/tb-accts-signup-page.ts
```diff
@@ -130,17 +136,25 @@ export class TbAcctsSignUpPage {
...
+    ).toBe('step-password');
 
     await this.passwordInput?.fill(password);
     await this.passwordConfirmInput?.fill(passwordConfirm);
   }
 
   async submitForm() { 
     // when clicking on android it won't click it unless we force it; but force doesn't work on ios
-    if (this.testPlatform.includes('android')) { 
+    console.log(`clicking '${await this.submitButton.innerText()}' button`);
```
davinotdavid, line 155: Should we remove the console.log here and leave the await?

davinotdavid (approved):
lgtm after comments! thanks!

test/e2e/pages/tb-accts-signup-page.ts
rwood-moz: Thanks!

No I prefer to leave this logging there because this same method is used for every step in the sign-up process, and without this when debugging you don't actually know which button is clicked i.e. `Setup Account` or `Next` etc.

PR #1057, opened 2026-07-03
Refactor notice-bar positioning in Keycloak Vue app

### LINKED ISSUE #1021: Info box missing margin when multiline.
Image

When the info box at the top wraps into two (or three) lines, it takes space away from below instead of leaving a consistent amount of margin.

### LINKED ISSUE #1036: Move keycloak global banner below the logo
I  don't have the zeplin ticket on hand but here's basically what I mean:

Image

This should happen on a base template level iirc, if it's not then we should maybe refactor the template to make it so.

davinotdavid:
## What changed?

- Added `message-bar` component to the `BaseTemplate` and removed all the duplicated message-bars from all the views in the Keycloak vue app
- Updated the padding of the `panel-contents` section to be horizontally even (6rem). Approved by the design team!
- Fixed the logo spanning the whole width as a container causing potentially misclicks
- Fixed the "Forgot Password" link spanning the whole width as a container causing potentially misclicks
- Removed duplicated styles for `.notice-bar` and `.logo-link`
- Removed the `max-height` in the panel so that it can grow according to its content instead of overflowing. This is perhaps a bit controversial as the left side orb graphic grows a bit when a layout shift happens (I know @MelissaAutumn changed this purposely before). IMHO it is less jarring to have it grow rather than overflowing but can put it back if there's a counter argument for it.

> [!NOTE]
> Since this updates _all_ the Keycloak app notices, please peruse the app to help test. I did trigger some of them and they look fine for me though! 

## Why?

- To better approximate the Zeplin designs: https://zpl.io/6JQPMYy. Please note that there's some deviation from the input itself but that's on purpose as we still allow folks to sign up with their recovery email at this moment.
- Reduce duplicated code

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/1036
Fixes https://github.com/thunderbird/thunderbird-accounts/issues/1021

## Screenshots

image

### DIFF
--- a/keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
+++ b/keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
@@ -97,10 +97,6 @@ section {
 
       .message-bar {
         margin-block-end: 1.5rem;
-
-        &:not(:has(:deep(.notice-bar))) {
-          display: none;
-        }
       }
 
       .logo-link {
@@ -112,7 +108,6 @@ section {
         .logo {
           height: 36px;
           width: auto;
-          transition: opacity 0.2s ease;
         }
       }
     }
--- a/keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
+++ b/keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
@@ -1,9 +1,8 @@
 <script setup lang="ts">
-import { ref } from 'vue';
 import ThundermailLogo from '@kc/svg/thundermail-logo-blue.svg';
+import MessageBar from '@kc/vue/components/MessageBar.vue';
 
 const clientUrl = window._page.currentView?.clientUrl;
-
 </script>
 
 <template>
@@ -18,6 +17,11 @@ const clientUrl = window._page.currentView?.clientUrl;
           <a :href="clientUrl" class="logo-link">
             <img :src="ThundermailLogo" alt="Thundermail" class="base-template__logo" />
           </a>
+
+          <div id="message-bar" class="message-bar">
+            <message-bar />
+          </div>
+
           <div class="panel-contents">
             <slot />
           </div>
@@ -66,7 +70,6 @@ section {
     width: 100%;
     height: auto;
     min-height: 100vh;
-    max-height: var(--max-card-height);
     display: flex;
     align-items: stretch;
     justify-content: center;
@@ -90,11 +93,25 @@ section {
       height: 100%;
       display: flex;
       flex-direction: column;
-      gap: 3rem;
+      gap: 2rem;
       padding: 6rem 2rem;
 
-      &:has(.notice-bar) {
-        padding: 6rem 2rem;
+      .message-bar {
+        &:not(:has(:deep(.notice-bar))) {
+          display: none;
+        }
+      }
+
+      .logo-link {
+        display: block;
+        text-decoration: none;
+        width: min-content;
+
+        .logo {
+          height: 36px;
+          width: auto;
+          transition: opacity 0.2s ease;
+        }
       }
     }
   }
@@ -146,13 +163,7 @@ section {
         min-height: auto;
 
         .panel {
-
-
-          padding: 6rem 10rem 5.625rem 6rem;
-
-          &:has(.notice-bar) {
-            padding: 6rem 10rem 5.625rem 6rem;
-          }
+          padding: 6rem 6rem 5.625rem 6rem;
         }
 
         .panel-contents {
--- a/keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
+++ b/keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
@@ -93,10 +93,11 @@ section {
       height: 100%;
       display: flex;
       flex-direction: column;
-      gap: 2rem;
       padding: 6rem 2rem;
 
       .message-bar {
+        margin-block-end: 1.5rem;
+
         &:not(:has(:deep(.notice-bar))) {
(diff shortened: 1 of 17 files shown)


davinotdavid:
Failed tests are unrelated btw!

rwood-moz:
> Failed tests are unrelated btw!

FYI Earlier I filed #1058 for that, contact tests failing when run on PRs/local stack.

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
```diff
@@ -90,11 +93,27 @@ section {
...
+      .logo-link {
+        display: block;
+        text-decoration: none;
+        width: min-content;
+        margin-block-end: 2rem;
+
+        .logo {
+          height: 36px;
+          width: auto;
+          transition: opacity 0.2s ease;
```
MelissaAutumn: Transition is still present, but I don't think you moved over the hover opacity. If we're removing the hover opacity change, can you nuke the transition line?

keycloak/themes/tbpro/assets/vue/views/RegisterView/index.vue
```diff
@@ -73,12 +72,9 @@ export default {
...
 
-  <slot name="notice-bars">
-    <notice-bar data-testid="error-notice-bar" :type="NoticeBarTypes.Critical" v-if="usernameError || passwordError || passwordConfirmError || recoveryEmailError">
-      {{ $t('registerError') }}
-    </notice-bar>
-    <message-bar v-else/>
-  </slot>
+  <notice-bar data-testid="error-notice-bar" :type="NoticeBarTypes.Critical" v-if="usernameError || passwordError || passwordConfirmError || recoveryEmailError">
+    {{ $t('registerError') }}
+  </notice-bar>
```
MelissaAutumn, line 82: Only worried this might result in double errors. But I don't think we use this view anymore. 🤔

keycloak/themes/tbpro/assets/vue/views/LoginView/index.vue
```diff
@@ -46,9 +42,8 @@ export default {
 </script>
 
 <template>
-  <notice-bar :type="NoticeBarTypes.Critical" v-if="firstError">{{ firstError }}</notice-bar>
-  <message-bar v-else-if="message?.type"/>
-  <notice-bar :type="NoticeBarTypes.Info" v-else>
+  <notice-bar :type="NoticeBarTypes.Critical" v-if="firstError" class="notice-bar-login">{{ firstError }}</notice-bar>
+  <notice-bar :type="NoticeBarTypes.Info" v-else-if="!message?.type" class="notice-bar-login">
```
MelissaAutumn, line 63: I think you can drop the .Info as message-bar will cover that in the parent.

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
```diff
@@ -90,11 +93,27 @@ section {
...
       padding: 6rem 2rem;
 
-      &:has(.notice-bar) {
-        padding: 6rem 2rem;
+      .message-bar {
+        margin-block-end: 1.5rem;
+
+        &:not(:has(:deep(.notice-bar))) {
+          display: none;
+        }
```
MelissaAutumn: What is this for? 🤔

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
```diff
@@ -66,7 +70,6 @@ section {
     width: 100%;
     height: auto;
     min-height: 100vh;
-    max-height: var(--max-card-height);
```
MelissaAutumn, line 69: Yeaaah that's probably fine. We already have the min-height, and I'm hopeful that'll prevent the orb from resizing too much.

MelissaAutumn (changes requested):
I'm not at my work laptop atm, but here's an initial review.

keycloak/themes/tbpro/assets/vue/views/RegisterView/index.vue
davinotdavid: I don't think we use this view anymore either but added a guard there similar with the LoginView just in case. Good catch!

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
davinotdavid: Ah yes, removed, thanks!

keycloak/themes/tbpro/assets/vue/views/LoginView/index.vue
davinotdavid: Hm so we can't drop this one though as this is the notice bar that shows the `Access is currently for invitation only` as this message pertains only for the `LoginView`

keycloak/themes/tbpro/assets/vue/BaseTemplate.vue
davinotdavid: Left over attempt of solving this with CSS instead of v-if's but thought it was way more confusing and error prone to rely on folks adding a specific class 😅 . Removed, thanks!

keycloak/themes/tbpro/assets/vue/views/LoginView/index.vue
MelissaAutumn: Hmmmm

MelissaAutumn (approved):
Looks good, just for reference I've tested:
* Login
* Login with errors
* Password reset
* Email verification
* Invalid page
* Logout
* Sign up

PR #1061, opened 2026-07-04
Don't reference the reason attribute which may not exist. (Sentry THUNDERBIRD-ACCOUNTS-8D)

markstos:
Before, we could get "object has no attribute 'reason'"

Fixes: https://thunderbird.sentry.io/issues/7561475693/

## QA Log

- Reviewed f string syntax and stringification.

### DIFF
--- a/src/thunderbird_accounts/authentication/tasks.py
+++ b/src/thunderbird_accounts/authentication/tasks.py
@@ -93,7 +93,7 @@ def tag_abandoned_cart_in_mailchimp(self):
             continue
         except Exception as ex:
             errors += 1
-            logger.exception(f'tag_abandoned_cart_in_mailchimp: failed to tag {user.uuid}: {ex.reason}')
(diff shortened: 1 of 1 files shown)


Sancus:
Is the reason never useful? If it is sometimes useful it seems like getattr would be better?

markstos:
@Sancus Good call-out.  `ex.reason` is useful in the `TaskFailed` case, because `TaskFailed` always defines it. I've pushed an amended commit to restore using `ex.reason` there.

What remains the singular problem case: In the exception case, we have no reason to expect that `reason` will be available here. The bug appeared to be a "copy/paste-o" case of copying the other logging string from the TaskFailed case and assuming it works in the general case. 

The narrower fix is to stringify `ex` in the general case where we know nothing about custom properties which might be there.

MelissaAutumn:
Ideally we cover our basis with the real exceptions, but this seems fine. I believe logging.exception should catch the exception as well.

MelissaAutumn (approved):

PR #1062, opened 2026-07-06
Fix contact E2E test when running on dev stack

### LINKED ISSUE #1058: E2E contact.spec test fails when run on PRs in CI (local stack)
The E2E contact.spec test is failing when it runs against a local stack (in CI on PRs, or on your local machine). The same test runs fine on stage and prod.

Here's an [example Github workflow](https://github.com/thunderbird/thunderbird-accounts/actions/runs/28671928225/job/85037618273) showing the tests failing in CI when run on a PR/local stack and the errors.

rwood-moz:
## What changed?
Fix the contact E2E test so it works when running against a local dev stack (i.e. on PRs).

## Why?
The contact test was failing when running against a local dev stack (on PRs in CI and/or when running on local machine on local stack). The test is passing fine on stage/prod. The problem is the local user (admin@example.org) used by the E2E tests when running against local stack is not on the email allow list which is required in order to submit the contact form. The fix is, when running the test on a local dev stack mock the is-on-allow-list check to return true, so then the test can proceed and actually submit the contact form.

## Applicable Issues
Fixes #1058.

## QA Log

- Ran the updated contact test on my local machine against my local dev stack, test now passes
- Ran the updated contact test against stage, ensuring it still passes there as it always did
- You'll see when the E2E tests run in this PR's checks that the contact test now passes

### DIFF
--- a/test/e2e/tests/desktop/contact.spec.ts
+++ b/test/e2e/tests/desktop/contact.spec.ts
@@ -1,5 +1,5 @@
 import path from 'path';
-import { test, expect } from '@playwright/test';
+import { test, expect, type Page } from '@playwright/test';
 import { ContactPage } from '../../pages/contact-page';
 import { ensureWeAreSignedIn, overridePageData } from '../../utils/utils';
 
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
lgtm, thanks! also, it would be great to eventually update the realm JSON for local environment so that this is automagically setup but we can punt this for later

PR #1066, opened 2026-07-06
🐛 Reload Stalwart after publishing DKIM records (platform-infrastructure#591)

markstos:
With Stalwart 0.15, after publishing a DKIM record the persistent store is updatedbut the in-memory cache is not, and it is the in-memory cache that's used for signing emails.

So until Stalwart is reloaded, emails for the domain may not be DKIM signed.

Here we had the extra reload call in it's own `try` block with exception handling for clear identification of the error.

Also, exception handling of the DKIM post is improved by moving the DKIM POST call into the pre-existin try block.

## Applicable Issues

 * https://github.com/thunderbird/platform-infrastructure/issues/591

## QA Log

* Reviewed Stalwart 0.15 docs for DKIM POST and Reload:
 * https://stalw.art/docs/0.15/api/management/endpoints/#create-dkim-signature
 * https://stalw.art/docs/0.15/api/management/endpoints/#reload-settings
* Analyzed 0.15 code base
* Manual testing adding a DKIM record, testing sending, reloading Stalwart, re-testing sending.
  Local testing confirmed that DKIM signing only started after Stalwart was reloaded.

### DIFF
--- a/src/thunderbird_accounts/mail/clients.py
+++ b/src/thunderbird_accounts/mail/clients.py
@@ -11,6 +11,7 @@ from thunderbird_accounts.mail.exceptions import (
     AccountNotFoundError,
     DomainNotFoundError,
     FailedToCreateDKIM,
+    FailedToReloadStalwart,
     StalwartError,
 )
 
@@ -236,6 +237,17 @@ class MailClient:
         self._raise_for_error(response)
         return response
 
+    def _reload(self):
+        response = requests.get(
+            f'{self.api_url}/reload/',
+            headers=self.authorized_headers,
+            verify=settings.VERIFY_PRIVATE_LINK_SSL,
+        )
+        response.raise_for_status()
+        self._raise_for_error(response)
+
+        return response
+
     def get_domain(self, domain):
         response = self._get_principal(domain)
 
@@ -261,6 +273,7 @@ class MailClient:
         dkim_algorithms = settings.STALWART_DKIM_ALGOS if algorithms is None else algorithms
         for algorithm in dkim_algorithms:
             data = {
+                # Stalwart 0.15, creates defaults IDs when ID is None. Ex: `$algo-$domain`
                 'id': None,
                 'algorithm': algorithm,
                 'domain': domain,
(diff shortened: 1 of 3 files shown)


MelissaAutumn:
Are there any issues from reloading Stalwart? Like is there the possibility of downtime, or task failure (like an email failing to send?)

Seems like pretty risky work-around to me.

markstos:
> Are there any issues from reloading Stalwart?

I've done a deeper review here of the reload behavior. Unlike a _restart_  which would be a service stop/start with a micro-outage, the /reload/ behavior is designed to be zero-download online. Here's what Stalwart does:

 - Stalwart 0.15 prepares a new core and only swaps over it if there are no problems
 - Existing request/sessions continue with the existing core, so they are not interrupted. 
 - The queue manager updates it's queue settings based on the signal, but otherwise continues.

This pattern is similar to a range of Unix services like Apache, Nginx, rsyslog and systemd which support SIGHUP/reload to safely reload configuration online without a full restart.

But still, I don't think we want user actions to cause our system to spam stalwart with reload signals, and there's a way we could vary the design a bit to mitigate that. Rather than running the `reload` in real-time, we could run the reloads at most once-per-minute. Like this: we could set a boolean in Redis to request a Stalwart reload and a Celery worker could run once a minute to check for this boolean, clear the state and run the reload.

But important context here is what we'll be doing in 0.16, so we aren't overengineering a solution for 0.15 semantics, only to throw it away in less an a month. For that reason, I've also reviewed how 0.16 handles DKIM updates.

_Reviews 0.16 code_

The issue essentially goes away in 0.16-- no extra reload call will be required. Given that, and given that this not a public route that even attempting to spam Stalwart reloads is unlikely to cause significant problems, I recommend the change as submitted. The extra complication of adding Redis+Celery to this fix introduces additional testing challenges and complexity which bring their own risk and delay getting a fix out.

davinotdavid (approved):
IMHO I think this is fine for now / until we have 0.16 in place.

It indeed sounds excessive to run the reload on every DKIM creation but if it doesn't cause downtime and the volume of the record creations is minimal enough, I think it is ok.

We should closely monitor Stalwart though to make sure that this is not affecting performance / deliverability. If it does we can revert + add an hourly task in Celery as described.

baskret75:
This doesn't appear to have resolved the issue. After re-adding a domain + aliases DKIM signing is still not working.

markstos:
@baskret75 Thanks for the report. Please contact support at https://support.tb.pro with the details of what you are seeing and we'll get it sorted out for you.

PR #1069, opened 2026-07-06
Add @active_subscription_required decorator and start using it (#347)

markstos:
Updates one location to use it as example. This commit is not trying
to refactor every potential use of the decorator.

## Why?

This addresses several Sentry exceptions. It's good for DRY and consistency with @login_required.

## Limitations and Notes

- Initial scope does not refactor every potential case that could use it.

## Applicable Issues

 * https://github.com/thunderbird/thunderbird-accounts/issues/347

## QA Log

* Confirmed there was already relevant test coverage for the refactored route, thus the code is already tested by existing test coverage.

## Questions

Should the scope be expanded to refactor more locations in this PR or continue that work in a follow-up PR?

### DIFF
--- a/src/thunderbird_accounts/subscription/decorators.py
+++ b/src/thunderbird_accounts/subscription/decorators.py
@@ -1,7 +1,8 @@
 from functools import wraps
 
 from django.conf import settings
-from django.http import JsonResponse
+from django.http import HttpResponseRedirect, JsonResponse
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
 try:
@@ -36,7 +37,7 @@ def inject_paddle(func):
     return _inject_paddle
 
 
-def active_subscription_required(function=None, *, error_message=None, status=403):
+def active_subscription_required(function=None, *, error_message=None, response_data=None, status=403):
     """Require an authenticated user with an active subscription."""
 
     def decorator(view_func):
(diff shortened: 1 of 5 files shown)


src/thunderbird_accounts/subscription/decorators.py
```diff
@@ -30,3 +34,22 @@ def _inject_paddle(*args, **kwargs):
...
+    """Require an authenticated user with an active subscription."""
+
+    def decorator(view_func):
+        @wraps(view_func)
+        def _view_wrapper(request, *args, **kwargs):
+            if getattr(request.user, 'has_active_subscription', False):
+                return view_func(request, *args, **kwargs)
+
+            message = error_message or _('An active subscription is required.')
+            return JsonResponse({'success': False, 'error': str(message)}, status=status)
```
MelissaAutumn: Can we check the `accept` header of the request and if it's application/json return this jsonresponse, and if it's not then return a redirect response to `reverse('vue_app')`?

Not required for this ticket and we can punt it to another ticket if it adds too much complexity.

MelissaAutumn (commented):
Perf, thanks! I added a couple of sentry issues to this ticket so let's sprinkle it around the application in this PR. 

Off the top of my head almost everything in main.views with login_required needs it.

src/thunderbird_accounts/subscription/decorators.py
markstos: Agreed, it's nice for error handles to conditionally return HTML or JSON depending on the request. Added.

markstos:
I've pushed a commit with the expanded scope requests to address HTML vs H JSON responses and broadly implementing the new decorator. 

As mentioned in Sentry, the CalDav case turned out to use token auth not the usual auth, so the decorator did not apply there and it was removed from the scope.

MelissaAutumn (approved):
Working on my machine! Thank you!!

PR #1072, opened 2026-07-07
Expose waffle flags through an API route

### LINKED ISSUE #1064: Add an api route for exposing waffle flags
This should take an access token, we should grab the user associated with that access token and return their enabled waffle flags. 

We technically have https://accounts.tb.pro/waffle_status but I don't believe it works with access token, just session auth which I'm not sure appointment would be able to ping. 

It's a default waffle url, so maybe investigate exposing this url with a bearer keycloak access token.

davinotdavid:
## What changed?

Added a GET `api/v1/auth/waffle-flags/` route that gets the feature flags available for the logged in user.

## Why?

A centralized place to control feature flags across services would be great and since we already have django-waffle set up here we can simply make a request from other services instead of having service-specific env var-based flags.

## Limitations and Notes

Since this is machine-to-machine, we lose the ability to rely on Waffle cookies for percent-rollout sticky-ness across calls unless another rule also qualifies the user.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/1064

### DIFF
--- a/src/thunderbird_accounts/authentication/api.py
+++ b/src/thunderbird_accounts/authentication/api.py
@@ -7,10 +7,12 @@ from enum import StrEnum
 from urllib.parse import quote
 
 from django.conf import settings
+from mozilla_django_oidc.contrib.drf import OIDCAuthentication
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.throttling import UserRateThrottle
-from rest_framework.permissions import AllowAny
+from rest_framework.permissions import AllowAny, IsAuthenticated
 import sentry_sdk
+from waffle import get_waffle_flag_model
 
 from thunderbird_accounts.authentication.exceptions import (
     InvalidDomainError,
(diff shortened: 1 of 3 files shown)


src/thunderbird_accounts/authentication/api.py
```diff
@@ -66,6 +68,28 @@ def get_user_profile(request: Request):
...
+@api_view(['GET'])
+@authentication_classes([OIDCAuthentication])
+@permission_classes([IsAuthenticated])
+def get_waffle_flags(request: Request):
+    """Return the caller's active waffle flags.
+
+    Callers authenticate with `Authorization: Bearer <keycloak-access-token>`
+    and we resolve that to the matching local user via OIDCAuthentication.
+    """
+    flags = get_waffle_flag_model().get_all()
```
MelissaAutumn: I think it might be easier to pull in waffle.views.waffle_json, do the auth checks in this function and pass the request in to waffle_json and return that. We basically just want to wrap it with our token auth.

MelissaAutumn (changes requested):
We're just testing / allowing session auth here. I think we might need to create a django rest framework auth class (subclassing is fine) to allow ingesting access token via jwt bearer token (so headers['authorization'] = 'bearer <whatever>'.)

MelissaAutumn:
Oh I thought OIDCAuth was ours, but it's drj. Looks like it does what we need to do. Could you adjust the tests to explicitly use headers['authorization'] = 'bearer xyz'. You might need to mock `authentication.middleware.AccountsOIDCBackend.get_userinfo` (or just that entire class?) in order to skip the keycloak calls.

src/thunderbird_accounts/authentication/api.py
davinotdavid: Sure, changed to just forwarding what comes from waffle_json, thanks!

This will expose not only the flags but also the switches and samples but I think that's fine?

davinotdavid:
@MelissaAutumn awesome thanks for the review and suggestions! Updated the code in the latest commits.

src/thunderbird_accounts/authentication/tests/test_api.py
```diff
@@ -353,3 +354,76 @@ def test_missing_emails_just_error_out(self):
...
+        response = self.client.get(self.url, headers={'authorization': f'Bearer {self.user.oidc_id}'})
+        self.assertEqual(200, response.status_code, response.content)
+        self.assertTrue(response.json()['flags']['flag-on-for-specific-user']['is_active'])
+
+        # Authenticate as the other user and check that the flag is not active
+        response = self.client.get(self.url, headers={'authorization': f'Bearer {other_user.oidc_id}'})
+        self.assertEqual(200, response.status_code, response.content)
+        self.assertFalse(response.json()['flags']['flag-on-for-specific-user']['is_active'])
+
+    def test_requires_authentication(self):
```
MelissaAutumn, line 433: One last suggestion, a test for a bogus bearer token / bogus auth value.

MelissaAutumn (approved):
Thanks! One suggestion for an extra test. Good to go otherwise.

src/thunderbird_accounts/authentication/tests/test_api.py
davinotdavid: Added, thanks!

PR #1075, opened 2026-07-07
Downgrade some expected "already exists" errors to info-level logging (#958)

### LINKED ISSUE #958: Don't log "email already exist" errors for alias/sign-up form
We're getting noise in Sentry about Keycloak 409s. This happens during the sign-up form, but it could probably happen also during the alias creation on the dashboard.

https://thunderbird.sentry.io/issues/7227985128/?project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream
https://thunderbird.sentry.io/issues/6872183330/?project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

For the purpose of our sanity we should not log 409 or other "username/address already exists" errors to sentry. It might be a good metric collection point in the future though (count-wise, not address.)

markstos:
For web/db apps, a common source of exceptions stem from users double-clicking
some button or link to add a thing when it only needs a single click.

In these cases, the first request will trigger a successful addition and the second
one will trigger some kind of "already exists" condition, which is harmless to ignore.

For some known cases like that, we downgrade them from throwing exceptions to
info-level logging.

Other cases of adding things where we still would not expect exceptions to be
triggered reamin as exceptions.

## Applicable Issues

* Closes #958

## QA Log

 * Automated analysis of places that need to be updated, including those flagged by
   Sentry.
 * Added automated test coverage, focusing on high-value test cases.

### DIFF
--- a/src/thunderbird_accounts/authentication/api.py
+++ b/src/thunderbird_accounts/authentication/api.py
@@ -314,9 +314,7 @@ def sign_up(request: Request):
     except ImportUserError as ex:
         if ex.is_already_exists and not can_register_with_username(partial_username):
             logging.info(
-                'Signup duplicate suppressed because a local username/address now exists (status=%s, error=%s)',
-                ex.status_code or 'none',
-                ex.error_code or 'unknown',
+                f'Dupe signup suppressed (status={ex.status_code or "none"}, error={ex.error_code or "unknown"})',
             )
         else:
             sentry_sdk.capture_exception(ex)
--- a/src/thunderbird_accounts/authentication/api.py
+++ b/src/thunderbird_accounts/authentication/api.py
@@ -5,6 +5,7 @@ from thunderbird_accounts.mail.exceptions import EmailNotValidError
 from thunderbird_accounts.authentication.permissions import CanCreateTestAllowListEntries
 from enum import StrEnum
 from urllib.parse import quote
+import logging
 
 from django.conf import settings
 from mozilla_django_oidc.contrib.drf import OIDCAuthentication
@@ -311,7 +312,14 @@ def sign_up(request: Request):
             status=400,
         )
     except ImportUserError as ex:
-        sentry_sdk.capture_exception(ex)
+        if ex.is_already_exists and not can_register_with_username(partial_username):
+            logging.info(
+                'Signup duplicate suppressed because a local username/address now exists (status=%s, error=%s)',
+                ex.status_code or 'none',
+                ex.error_code or 'unknown',
+            )
+        else:
+            sentry_sdk.capture_exception(ex)
         return Response(
             {
                 'error': ex.error_desc if ex.error_desc else _('There was an unknown error, please try again later.'),
--- a/src/thunderbird_accounts/authentication/clients.py
+++ b/src/thunderbird_accounts/authentication/clients.py
@@ -44,6 +44,27 @@ class RequestMethods(enum.StrEnum):
     DELETE = 'delete'
 
 
+def _keycloak_error_details(exc: RequestException) -> tuple[Optional[int], str, dict]:
+    """Given a keycloak exception, extract the error details"""
+    response = exc.response
+    if response is None:
+        return None, '', {}
+
+    response_content = response.content.decode(errors='replace')
+    try:
+        error_data: dict = json.loads(response_content)
+    except (TypeError, JSONDecodeError):
+        error_data = {}
+
+    return response.status_code, response_content, error_data
+
+
+def _format_keycloak_error(exc: RequestException, status_code: Optional[int], response_content: str) -> str:
+    if response_content:
+        return f'Error<{status_code}>: {response_content}'
+    return f'Error<{exc}>: No response!'
+
+
 class TotpSetupResponse(TypedDict, total=False):
     secret: str
     encodedSecret: str
@@ -339,22 +360,22 @@ class KeycloakClient:
                 },
             )
         except RequestException as exc:
-            sentry_sdk.capture_exception(exc)
-
-            try:
-                error_data: dict = json.loads(exc.response.content.decode())
-            except (TypeError, JSONDecodeError):
-                # Could not determine explicit error information
-                error_data = {}
+            status_code, response_content, error_data = _keycloak_error_details(exc)
 
             # Note: status_code == 409 means the user already exists on keycloak which we should have caught before this
             # function call.
-            raise ImportUserError(
+            error = ImportUserError(
                 username=username,
-                error=f'Error<{exc.response.status_code}>: {exc.response.content.decode()}',
-                error_code=error_data.get('error', f'status-{exc.response.status_code}'),
+                error=_format_keycloak_error(exc, status_code, response_content),
+                error_code=error_data.get('error', f'status-{status_code}' if status_code is not None else None),
                 error_desc=error_data.get('error_description', error_data.get('errorMessage')),
+                status_code=status_code,
             )
+
+            if not error.is_already_exists:
+                sentry_sdk.capture_exception(exc)
+
+            raise error
 
         # Request returns an empty body on a 201 success, so retrieve pkid from location.
         # ex/ {'Location': 'http://keycloak:8999/admin/realms/tbpro/users/39a7b5e8-7a64-45e3-acf1-ca7d314bfcec', ... }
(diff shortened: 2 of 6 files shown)


src/thunderbird_accounts/authentication/api.py
```diff
@@ -296,7 +297,14 @@ def sign_up(request: Request):
             status=400,
         )
     except ImportUserError as ex:
-        sentry_sdk.capture_exception(ex)
+        if ex.is_already_exists and not can_register_with_username(partial_username):
+            logging.info(
+                'Signup duplicate suppressed because a local username/address now exists (status=%s, error=%s)',
```
MelissaAutumn: We should maybe avoid the old python formatting. fstrings are fine here.

src/thunderbird_accounts/authentication/clients.py
```diff
@@ -44,6 +44,27 @@ class RequestMethods(enum.StrEnum):
     DELETE = 'delete'
 
 
+def _keycloak_error_details(exc: RequestException) -> tuple[Optional[int], str, dict]:
```
MelissaAutumn: This should really live inside an exception that we throw from a requestexception instead of as a misc private function. 

Let's avoid creating these misc private functions if we can, it creates a lot of noise within a file. I think we can shove most of this code in the authentication.exceptions.KeycloakError class and just automagically handle the data as it comes in.

src/thunderbird_accounts/authentication/exceptions.py
```diff
@@ -1,6 +1,24 @@
 from typing import Optional
 
 
+POSTGRES_DUPLICATE_KEY_MARKER = 'duplicate key value violates unique constraint'
+
+
+def is_already_exists_error(
```
MelissaAutumn: This can live inside the KeycloakError class. No need for it to live in outside.

src/thunderbird_accounts/mail/views.py
```diff
@@ -445,7 +445,10 @@ def add_email_alias(request: HttpRequest):
     if (not is_catch_all and not email_alias) or not domain:
         return JsonResponse({'success': False, 'error': _('Email alias and domain are required.')}, status=400)
 
-    if (not is_catch_all and not is_custom_domain and is_reserved(email_alias)) or is_address_taken(full_email_alias):
+    if not is_catch_all and not is_custom_domain and is_reserved(email_alias):
+        return JsonResponse({'success': False, 'error': _('You cannot use this email address.')}, status=403)
+
+    if is_address_taken(full_email_alias):
```
MelissaAutumn: What's up with this change?

src/thunderbird_accounts/mail/views.py
```diff
@@ -487,10 +490,17 @@ def add_email_alias(request: HttpRequest):
         )
 
         if not created:
+            logging.info('Alias creation found an existing local email record for the requested address')
             return JsonResponse(
                 {'success': False, 'error': _('This email address is not available.')},
                 status=400,
             )
+    except IntegrityError:
+        logging.info('Alias creation hit a local duplicate-address race')
```
MelissaAutumn, line 498: I don't think we need this log item

src/thunderbird_accounts/mail/views.py
```diff
@@ -487,10 +490,17 @@ def add_email_alias(request: HttpRequest):
         )
 
         if not created:
+            logging.info('Alias creation found an existing local email record for the requested address')
```
MelissaAutumn, line 492: I don't think we need this log item

src/thunderbird_accounts/authentication/exceptions.py
```diff
@@ -1,6 +1,24 @@
...
+    error_code: Optional[str] = None,
+    error_desc: Optional[str] = None,
+    error: Optional[str] = None,
+) -> bool:
+    # HTTP 409 means "Conflict" and is returned by Keycloak when a resource already exists.
+    if status_code == 409:
+        return True
+
+    error_text = ' '.join(filter(None, [error_code, error_desc, error])).lower()
+    return POSTGRES_DUPLICATE_KEY_MARKER in error_text
```
MelissaAutumn: Is this something we hit? I don't remember duplicate key errors here, but also I wasn't really looking at the response other than status code.

MelissaAutumn (changes requested):
Good start, I'd like to see those private functions dropped and code moved inside our existing classes where it makes sense.

src/thunderbird_accounts/authentication/api.py
markstos: Good idea for consistency. I opened a PR to automatically enforce f-strings. https://github.com/thunderbird/thunderbird-accounts/pull/1088

src/thunderbird_accounts/mail/views.py
markstos: Removing in my next commit.

src/thunderbird_accounts/mail/views.py
markstos: Removing in next commit.

src/thunderbird_accounts/mail/views.py
markstos: It made sense in an earlier draft but doesn't add value now. Good catch. Reverting.

src/thunderbird_accounts/authentication/exceptions.py
markstos: That addition was experience-based rather than evidence-based. Having managed busy Postgres-backed websites in the past, at scale that became a common pattern: The user's double-click behavior would lead to a race condition to insert twice. One insert would win and the other would throw the "duplicate key" error.

We could remove it because we haven't hit it yet, but based on the scale we are trying to build we likely will. Also, it seems is the best solution: there's not a reasonable way to prevent the condition from happening: the best path seems to be accept that we'll trigger the duplicate key violation periodically, then explicitly accept this condition as OK in cases were we expect them to be safe, like a double-click for inserting record targeted here.

src/thunderbird_accounts/authentication/exceptions.py
markstos: The `is_already_exists_error` was put in the global scope with the understanding that the pattern of generating exceptions from double-click-to-insert affects  all things we insert-- the exception we noticed related to keycloak happens to be the first one. As we scale, the pattern will repeat for everything we insert, see note below.

But in the spirit of YAGNI, I've moved it into the narrowest scope where it's actually used now, the `ImportUserError` class. Later when we hit similar cases and refactor to handle those, we can refactor to move this back to a shared location.

src/thunderbird_accounts/authentication/clients.py
markstos: I'll refactor to move the code into the error class and in-line a couple of the small single-use functions.

When you suggest to "avoid creating these misc private functions", perhaps you are referring to small single-use functions? I can try lean towards inlining code more often in those cases.

markstos:
Feedback has been addressed in a fixup commit that can be reviewed and then squashed during merging.

src/thunderbird_accounts/authentication/api.py
MelissaAutumn: This still needs to be resolved though 😄

src/thunderbird_accounts/authentication/exceptions.py
```diff
@@ -48,6 +54,59 @@ def __init__(
...
+        # HTTP 409 means "Conflict" and is returned by Keycloak when a resource already exists.
+        if status_code == 409:
+            return True
+
+        error_text = ' '.join(filter(None, [error_code, error_desc, error])).lower()
+        return POSTGRES_DUPLICATE_KEY_MARKER in error_text
+
+    @property
+    def is_already_exists(self):
+        return self.is_already_exists_error(
```
MelissaAutumn: I feel like we should just inline `is_already_exists_error` here. I see how it may be slightly handy to test errors without wrapping them, but it kinda expands the scope of this class as a utility class rather than a exception.

MelissaAutumn (changes requested):
I believe we have a linting error now 😄 

And a suggestion to be slightly more narrow in the exception usage. Otherwise good stuff.

src/thunderbird_accounts/authentication/exceptions.py
markstos: I agree that we don't need to keep both `is_already_exists` and `is_already_exists_error`. I've collapsed those into a single function.

MelissaAutumn (approved):

MelissaAutumn:
Thanks!

Issue #1050, opened 2026-06-30
Catch errors on Mailchimp user creation during Sign Up


davinotdavid:
## Describe the bug

Currently, [we are tagging users in Mailchimp during sign up](https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/subscription/tasks.py#L531-L532) but not catching errors if we fail to do so.

In case we fail and these users never complete their payment, we attempt to tag them in Mailchimp with `abandoned_cart` tag. However, if there was a failure to create the user in Mailchimp during sign up, we will also fail to tag them. Repeatedly. Hourly.

## To Reproduce

There was a specific user account that Mailchimp refuses to create as being flagged as spammy but we can probably reproduce this issue by attempting to create a user with a fake email.

## Additional context

There's loads of events in Sentry but it is a single user failure, failing hourly.
https://thunderbird.sentry.io/issues/7577764651/events/e239b459059b445dbd52013a00612560/

markstos:
Resolved by linked commits

PR #1078, opened 2026-07-07
Stop attempting to post emails to Mailchimp that are being rejected (#1050)

markstos:
Attempting to post email to Mailchimp that Mailchimp rejects was causing an infinite loop
of exceptions being posted to Sentry as we retried every hour.

To address this, we start tracking which emails Mailchimp rejects and when.
For now, we use this to stop posting emails to Mailchimp that are being rejected.

Later, this may be a useful signal that the account has a history of being
associated with spam. In that case, we might want to use the ginal for
later spam preventation or fraud prevention efforts.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/subscription/tasks.py

## QA Log

 - Added related test coverage.


davinotdavid (approved):
LGTM, thanks!

### QA Log

- Reviewed code diff
- Checked Mailchimp docs to see if there's a more specific / deterministic error that we can capture in this case (there is none apparently bleh)

PR #1082, opened 2026-07-08
Handle pending mail account provisioning case (#1081)

### LINKED ISSUE #1081: Evaluate if /caldav/setup/  AccountNotFoundError should raise an exception or return 404 and log
If this condition is expected/possible in some some edge cases, then return 404 and info-level logging seems suitable. If it's really expected to be "impossible", raising an exception is appropriate. 

Sentry Issue: [THUNDERBIRD-ACCOUNTS-8Q](https://thunderbird.sentry.io/issues/7596663468/?referrer=github_integration)

```
AccountNotFoundError: AccountNotFoundError: None
  File "thunderbird_accounts/mail/views.py", line 635, in appointment_caldav_setup
    email_user = stalwart_client.get_account(user.stalwart_primary_email)
  File "thunderbird_accounts/mail/clients.py", line 431, in get_account
    raise AccountNotFoundError(principal_id)
```

markstos:
Appointment can call the CalDAV setup endpoint after subscription has marked a user active, but before the async Stalwart account provisioning task has completed. In that window the user has no local primary Stalwart email yet, so the old flow could throw AccountNotFoundError.

Return a machine-readable 409 `mail_account_not_ready` response for that transient state where user.stalwart_primary_email is not set.

We will continue to throw on other Stalwart lookup failures, as those remain unexpected.

## Applicable Issues

 * Fixes #1081 
 * Issue for Appointment's side:  https://github.com/thunderbird/thunderbird-accounts/pull/1082
 
## QA Log

* Confirmed we have test coverage for this route.
  * ( src/thunderbird_accounts/mail/tests/test_views.py:999, AppointmentCalDAVSetupTestCase.)

### DIFF
--- a/src/thunderbird_accounts/mail/tests/test_views.py
+++ b/src/thunderbird_accounts/mail/tests/test_views.py
@@ -1128,6 +1128,25 @@ class AppointmentCalDAVSetupTestCase(TestCase):
         self.assertEqual(payload['error'], _('An error has occurred while setting up the Appointment CalDAV.'))
 
     @override_settings(APPOINTMENT_CALDAV_SECRET='test-secret-123')
+    @patch('thunderbird_accounts.mail.views.AccountsOIDCBackend')
+    @patch('thunderbird_accounts.mail.views.MailClient')
+    def test_mail_account_not_ready(self, mock_mail_client_cls, mock_backend_cls):
+        Email.objects.filter(type=Email.EmailType.PRIMARY, account=self.account).delete()
+        mock_backend = Mock()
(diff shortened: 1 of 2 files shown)


davinotdavid (changes requested):
Nice, thank you! Code diff lgtm but I think we should add a test for this as there's no coverage for this exact case with the status 409 and code `mail_account_not_ready`.

Even though it is a small enough change, since Appointment will rely on this in its counterpart implementation, we should harden this in tests.

davinotdavid (approved):
lgtm, thanks!

PR #974, opened 2026-06-16
ci(keycloak): build the keycloak image multi-arch (amd64 + arm64)

aatchison:
## What

Builds the Keycloak image (`<repo>:keycloak-<sha>`) as a **multi-arch manifest list** (`linux/amd64` + `linux/arm64`) via `docker buildx`, instead of amd64-only.

```diff
- docker build -f Dockerfile.keycloak -t $ECR_TAG --platform="linux/amd64" .
- docker push $ECR_TAG
+ docker buildx build -f Dockerfile.keycloak -t $ECR_TAG \
+   --platform linux/amd64,linux/arm64 --push .
```
Plus `docker/setup-qemu-action` + `docker/setup-buildx-action` (guarded by the existing `BUILD_KEYCLOAK` condition).

## Why

The Thunderbird Pro **EKS** clusters (Graviton / arm64) are the migration target for Customer Auth Keycloak ([platform-infrastructure#132](https://github.com/thunderbird/platform-infrastructure/issues/132), [#531](https://github.com/thunderbird/platform-infrastructure/issues/531)). The current amd64-only image can't be scheduled on arm64 nodes. A manifest list keeps **ECS Fargate (amd64) working unchanged** while letting the **same `keycloak-<sha>` tag** run on EKS arm64.

## Notes
- Theme stage (`FROM node:*`) is arch-independent; base `quay.io/keycloak/keycloak:26.5` is already multi-arch, so arm64 builds cleanly (QEMU on the amd64 runner; modest extra build time).
- No change to the accounts app image (stays amd64 for ECS) or to the deployment artifact contract (`build-keycloak.outputs.keycloak-image` unchanged).

PR #1087, opened 2026-07-08
CI: decouple multi-arch image publishing into a standalone GHCR pipeline

aatchison:
## What

Splits container-image **publishing** out of the **deploy** pipeline into a standalone GHCR pipeline, and moves the multi-arch (amd64 + arm64) build onto **native runners** (no QEMU) using the split-worker strategy from [sredevopsorg/multi-arch-docker-github-workflow](https://github.com/sredevopsorg/multi-arch-docker-github-workflow).

## Two independent pipelines

| Pipeline | Workflow | Registry | Arch | Deploys? |
|---|---|---|---|---|
| **Deploy-stage** | `merge.yml` | ECR | amd64 | ✅ Pulumi → stage |
| **Publish-images** (new) | `publish-images.yml` + `build-multiarch-image.yml` | GHCR | amd64 + arm64 | ❌ publish only |

See `docs/ci/image-publish-pipeline.md` for a mermaid diagram of both.

## Changes

- **`merge.yml`** — reverts the deploy build to **single-arch amd64 → ECR** (undoing the QEMU/Buildx multi-arch added in #1027 / #974). Deploy, artifacts, and `detect-changes` (its own job) are otherwise unchanged.
- **`publish-images.yml`** (new) — on merge to `main` (or `workflow_dispatch`): a **separate `detect-changes` job** decides which image changed and reads the version, then builds `accounts` and `keycloak` to GHCR. No cloud credentials, no Pulumi.
- **`build-multiarch-image.yml`** — reusable native split-arch builder targeting **GHCR via `GITHUB_TOKEN`**: amd64 on `ubuntu-latest`, arm64 on `ubuntu-24.04-arm`, pushed by digest, merged into one manifest list tagged **`:<sha>`, `:v<version>`, `:latest`**. Separate GHCR packages per image (`.../thunderbird-accounts`, `.../thunderbird-accounts-keycloak`).
- **`docs/ci/image-publish-pipeline.md`** — architecture diagram + rationale.

## Verified

An earlier revision of this branch ran the native builds green (all four amd64/arm64 build jobs succeeded; arm64 on native Graviton runners ~165s for accounts). `actionlint` passes on all three workflows.

## ⚠️ Reviewer call-out: arm64 in production

Reverting `merge.yml` to amd64-only means **ECR no longer gets an arm64 image**. The Graviton/arm64 EKS need that #1027/#974 addressed is now served **only from GHCR**. This is intentional per the decoupling, **but** anything currently pulling the arm64 variant from ECR must be repointed to GHCR (or ECR multi-arch restored) before those consumers rely on it. Flagging so it's a conscious decision.

## Action versions (reviewer note)

All third-party actions in the **two new** workflows are pinned to a commit SHA (with a `# vX.Y.Z` comment); no existing workflow was touched. Getting to "latest" meant some **notable major bumps** that differ from the `v3`/`v4` tags the rest of the repo pins:

| Action | Repo elsewhere | New workflows |
|---|---|---|
| actions/checkout | `v4` | **v7.0.0** |
| docker/login-action | `v3` | **v4.4.0** |
| docker/setup-buildx-action | `v3` | **v4.2.0** |
| actions/upload-artifact | `v4` | **v7.0.1** |
| actions/download-artifact | `v4` | **v8.0.1** |
| dorny/paths-filter | `v3` | **v4.0.2** |
| SebRollen/toml-action | `v1.2.0` | v1.2.0 (already latest) |

Low-risk for our simple usage, and `upload-artifact@v7` ↔ `download-artifact@v8` are the intended compatible pair (v8 supports v7's direct-upload feature). Flagging in case reviewers prefer version **consistency** with the rest of the repo — any single one is easy to dial back.

## Follow-up

Deploy-side companion (pull the accounts image from GHCR on the Graviton EKS clusters): **thunderbird/accounts-deploy#25** (draft) — blocked on this PR merging + the GHCR images publishing.

## Notes

- `release.yml` is left as-is: its `imagetools` retag (from #1027) works fine on a single-arch amd64 image, so no revert needed there.

### DIFF
--- a/.github/workflows/build-multiarch-image.yml
+++ b/.github/workflows/build-multiarch-image.yml
@@ -1,34 +1,32 @@
 ---
-# Reusable workflow that builds one container image for linux/amd64 and
-# linux/arm64 on NATIVE runners (no QEMU), pushes each arch to ECR by digest,
-# then stitches the digests into a single multi-arch manifest list under the
-# real tag. Callers get back the fully-qualified image reference via the
-# `image` output. See merge.yml for how accounts + keycloak invoke this.
+# Reusable workflow: build ONE container image for linux/amd64 and linux/arm64
+# on NATIVE runners (no QEMU), push each arch to GHCR by digest, then stitch the
+# digests into a single multi-arch manifest list tagged :<sha>, :v<version> and
+# :latest. Callers get the fully-qualified :<sha> reference back via the `image`
+# output. Invoked by publish-images.yml for the accounts and keycloak images.
 #
 # Why native runners instead of `buildx --platform amd64,arm64` with QEMU:
 # emulated arm64 builds are minutes-slower and occasionally miscompile. Each
-# arch now builds on its own hardware and the two run in parallel.
+# arch builds on its own hardware and the two run in parallel.
 #
-# ============================ TEMPORARY: DRY-RUN =============================
-# This workflow is currently in a build-test mode: it builds each arch on its
-# native runner but does NOT push to ECR and does NOT produce a manifest list.
-# The AWS/ECR steps and the digest/merge machinery are commented out below so
-# the build can be exercised on any branch with no cloud credentials and no
-# side effects. Restore the commented blocks (and `secrets: inherit` in
-# merge.yml) once the native builds are confirmed green.
-# ============================================================================
+# Auth is GHCR + the automatic GITHUB_TOKEN (packages: write) -- no cloud
+# credentials, which keeps this pipeline decoupled from the ECR/Pulumi deploy.
 name: build-multiarch-image
 
 on:
   workflow_call:
     inputs:
+      image:
+        description: Fully-qualified GHCR image name WITHOUT a tag (e.g. ghcr.io/owner/name).
+        required: true
+        type: string
       dockerfile:
         description: Path to the Dockerfile to build.
         required: false
         type: string
         default: Dockerfile
-      tag:
-        description: Tag to publish the manifest list under (appended to the ECR repo).
+      version:
+        description: Semantic version (no leading v) published as the :v<version> tag.
         required: true
         type: string
       artifact-prefix:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-multiarch-image.yml
```diff
@@ -0,0 +1,165 @@
...
+  # One job per architecture, each on its matching native runner. Each pushes an
+  # anonymous (tag-less) image to GHCR and records its digest as an artifact.
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            arch: amd64
+            runner: ubuntu-latest
```
MelissaAutumn: only real nit, why ubuntu latest here but 24.04 for arm?

.github/workflows/merge.yml
```diff
@@ -45,7 +46,15 @@ jobs:
...
+              - 'vite.config.mts'
+              - 'tsconfig.json'
+              - 'env.d.ts'
+              # The accounts bundle imports the shared keycloak theme via the
+              # `@kc` vite alias, so `npm run build` compiles it into the accounts
+              # image -- theme edits must rebuild (and redeploy) accounts too.
+              - 'keycloak/themes/**'
             keycloak-theme-changed:
               - 'keycloak/themes/**'
               - 'Dockerfile.keycloak'
```
MelissaAutumn, line 60: In the future I'd like to be able to force keycloak builds. We have a weird edge-case where a keycloak changes aware CI job can be cancelled by a non keycloak change aware CI job. (If folks merge to main before the previous job is done.)

MelissaAutumn (approved):
Looks okay.

markstos:
@aatchison This can be merged after the review comments are noted.

.github/workflows/build-multiarch-image.yml
aatchison: Good catch — there's no `ubuntu-latest-arm` alias, so I've pinned the amd64 runner to `ubuntu-24.04` to match `ubuntu-24.04-arm`. The build matrix is now symmetric and reproducible (619db21).

.github/workflows/merge.yml
aatchison: Noted — tracked as #1107. Worth mentioning: the new GHCR pipeline (`publish-images.yml`) already avoids both halves of this — it uses `cancel-in-progress: false` and its `workflow_dispatch` force-builds both accounts and keycloak. #1107 covers bringing the same (force-build input + concurrency reconsideration) to the deploy-stage pipeline.

PR #1088, opened 2026-07-08
Enforce f-string interpolation with Ruff

markstos:
## What changed?

Adds rule to express that we prefer f-strings, enforcing it with pre-commit hook and CI.

## Why?

Saves time in the peer-review process from accidentally not using them and then
having them flagged by a human and fixed.

## Limitations and Notes

Devs need to each install the pre-commit hook:

```
uv tool install pre-commit
pre-commit install
```
## Applicable Issues

Inspired by PR #1075, where I used non-f-strings in a few places, a mistake
which which likely to be completed by myself or other new contributors without
automation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ These can be installed be appending `--extra <optional dependency>` like so:
 uv sync --extra cli --extra docs --extra subscription
 ```
 
+## Python linting
+
+Ruff is the source of truth for Python linting. Read `pyproject.toml` and `pulumi/ruff.toml` for the active rules.
+
 ## Re-bootstrapping the project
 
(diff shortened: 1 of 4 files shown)


MelissaAutumn:
I think the linting rule is fine, but I'm not a fan of pre-commit hooks. That maybe should be a team discussion on whether we adopt them. 

Can you drop the pre-commit hook, the lint rule is good as-is. Thanks!

Sancus:
pre-commit hooks just kinda get in the way when doing local dev imo. Correct place for linting enforcement is in ci/cd on PR. It has to be there anyway since pre-commit hooks are unenforceable.

markstos:
I've amended to remove the pre-commit hooks.

MelissaAutumn (approved):
tyvm!!

Issue #1039, opened 2026-06-29
Send icon in account dashboard redirects back to dashboard instead of loading Send


aaspinwall:
## Describe the bug
The Send icon in the Thundermail account dashboard redirects back to the Thundermail account dashboard instead of loading the Send dashboard.

When a user opens the account dashboard and clicks the Send icon, the Send URL appears in the address bar and the Send dashboard briefly displays, but it fails to load after approximately one minute and then redirects back to `accounts.tb.pro/dashboard`.

Upstream report: https://bugzilla.mozilla.org/show_bug.cgi?id=2051092

## To Reproduce

1. Sign into a Thundermail account
2. Navigate via Hamburger menu > Thundermail account name > Account Dashboard
3. Click on the Send icon
4. See the Send dashboard briefly load, then redirect back to `accounts.tb.pro/dashboard`

## Expected behavior
The Send dashboard should load at `send.tb.pro`.

## Actual behavior
The Send URL appears in the address bar and briefly displays the Send dashboard, but fails to load after ~1 minute, then redirects back to `accounts.tb.pro/dashboard`.

## System

- OS: Linux x86_64 (X11)
- Browser Version: Thunderbird 152.0 and 154

## Additional context
The Send link from "Your Services" needs to point at the Send dashboard route with the `showDashboard` flag so it loads the dashboard directly instead of bouncing back to the accounts dashboard. Originally reported by Ryan Camaratta (:ryancnap).

PR #1089, opened 2026-07-09
Revert tbProSendUrl to its root url

davinotdavid:
## What changed?

Essentially reverting https://github.com/thunderbird/thunderbird-accounts/pull/1040/changes

## Why?

`https://send.tb.pro/profile?showDashboard=true` gives us a 404.

## Limitations and Notes

Unsure why this change was done but it seemed deliberate? Apparently send has a redirect from `/profile` to `/send/profile` [in this router code](https://github.com/thunderbird/tbpro-add-on/blob/main/packages/send/frontend/src/apps/send/router.ts#L68-L69) so unsure if the fix here is:

a) Revert the change so that the link at least is not broken
b) Keep the change but add `/send/profile` to the URL
c) Investigate + fix it in `tbpro-add-on` code base instead

If this is urgent and needs to be fixed asap, I'd do a).

## Applicable Issues

Relates to https://github.com/thunderbird/thunderbird-accounts/pull/1040/changes
Also relates to https://github.com/thunderbird/thunderbird-accounts/issues/1039

### DIFF
--- a/assets/app/vue/components/YourServices.vue
+++ b/assets/app/vue/components/YourServices.vue
@@ -1,16 +1,10 @@
 <script setup lang="ts">
-import { computed } from 'vue';
 import { useI18n } from 'vue-i18n';
 
 const { t } = useI18n();
 
 const tbProAppointmentUrl = window._page?.tbProAppointmentUrl;
-const tbProSendUrl = computed(() => {
-  if (!window._page?.tbProSendUrl) return undefined;
-  const url = new URL('profile', window._page.tbProSendUrl);
-  url.searchParams.set('showDashboard', 'true');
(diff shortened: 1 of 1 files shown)


ryanjjung:
Tiny bit of context. It displays the 404 error text provided by [this Vue template](https://github.com/thunderbird/tbpro-add-on/blob/bf4dc49c09c76dacc7d12de41791cbcbb262ca73/packages/send/frontend/src/apps/common/NotFoundPage.vue), but the actual HTTP responses are 2xx class statuses throughout. We deliver a 200 response but a 404 message.

MelissaAutumn (approved):

PR #1096, opened 2026-07-09
Add temporary admin routes to backfill DNS records (#1079)

### LINKED ISSUE #1079: Backfill missing DKIM records in Cloudflare
In https://github.com/thunderbird/thunderbird-accounts/issues/1060 we found a number of DKIM records were not published to Cloudflare.

This issue is about backfilling those and also considering possibly automating self-healing of similar issues in the future, such as a task that periodically looks for such missing records and backills them.

## Process

First, we'll need to pull a list of our DKIM records out of Stalwart and then iterate over that list and check to see if we records in Cloudflare. We could do that check via DNS without affecting Cloudflare API quota we may have. 

In a dry-run mode, we can produce a report of the number and domain names of affected domains.

In live mode, we'll run the task and report on what was updated.

This could be implemented as a one-time migration. 

From there, we can evaluate if it's worth it to automate self-healing. Because over time we plan to support thousands of such records and domains, iterating them over them all for a single check will stop being resource efficient. Maybe a better approach instead is increase the number of spacing of our automated retries for the *initial*  add so that we weather longer outages of our DNS partner.

What happened with Cloudflare is something we hope would be an especially rare case: we permanently exceeded our quota and didn't notice for days, making the system behave as if Cloudflare was down for days-- not a normal situation for modern DNS providers.

## Related issues

A follow up to #1060 

Note that this work distinct from and higher priority than the other DKIM backfilling issue: https://github.com/thunderbird/thunderbird-accounts/pull/960
That issue adding the "tm1/2/3" DKIM selector records for customers that don't have them. Those customers can use their existing selectors until we upgrade to Stalwart 0.16 and resume that project.

This task is for backfilling customers who have their "tm" DKIM selector records in Stalwart but we failed to publish them to Cloudflare.

markstos:
Once deployed, these new routes work for admins:

  - Dry run: /mail/admin/hosted-dkim-backfill/
  - Single domain: /mail/admin/hosted-dkim-backfill/?domain=example.com
  - Limited dry run: /mail/admin/hosted-dkim-backfill/?limit=25
  - Live, throttled: /mail/admin/hosted-dkim-backfill/?live=1&limit=25&sleep_seconds=10
  - Live, no limit, one second throttling: /mail/admin/hosted-dkim-backfill/?live=1

After a single-use, this can be torn down. It's not expected to be needed
in the future thanks to self-healing added in #1095

## Limitations and Notes

- It's implemented to "stream" results to the browser as plain text. This provides real-time feedback and helps prevent time-outs from proxies for longer runs.

- If it's interrupted or if we do some smaller warm-up runs, it will pick up where it left off
  by re-querying to find the cases with missing records.

- The code is intentionally not "DRY", but rather was created to intentionally not modify
  any existing code so it can be cleanly removed later. 

- For audit purposes, it logs both the browser and to server logs to track which domains were updated.

## QA Log

 - Because this depends on Cloudflare to really test, the best way to QA it will be to use
   "dry run" mode on Staging, where it may find nothing to do, and then use dry-run on production,
    followed perhaps by a single-domain update, a "limited" and more throttle run and finally the
    fully, run with minimal throttling (still enough to stay without Cloudflare limits)

Overall, should be considered low risk as it only adds missing Cloudflare records.
 
## Applicable Issues

 * Fixes #1079

### DIFF
--- a/src/thunderbird_accounts/mail/hosted_dkim_backfill.py
+++ b/src/thunderbird_accounts/mail/hosted_dkim_backfill.py
@@ -1,0 +1,347 @@
+import logging
+import re
+import time
+from dataclasses import dataclass
+from typing import Optional
+
+import dns.resolver as dns_resolver
+from django.conf import settings
+from django.contrib.admin.views.decorators import staff_member_required
+from django.core.exceptions import ImproperlyConfigured
+from django.http import Http404, HttpRequest, StreamingHttpResponse
+from django.views.decorators.cache import never_cache
+
+from thunderbird_accounts.mail.clients import DNSRecordStatus, MailClient
+from thunderbird_accounts.mail.dkim import CloudflareDNSClient
+from thunderbird_accounts.mail.models import Domain
+
+
+TXT_TAG_SPEC_RE = re.compile(
+    r'[ \t]*(?P<tag>[A-Za-z][A-Za-z0-9_]*)[ \t]*=[ \t]*'
+    r'(?P<value>(?:[!-:<-~]+(?:[ \t]+[!-:<-~]+)*)?)[ \t]*'
+)
+
+
+@dataclass
+class BackfillCounters:
+    domains_scanned: int = 0
+    domains_with_stalwart_dkim: int = 0
+    domains_without_stalwart_dkim: int = 0
+    records_checked: int = 0
+    match: int = 0
+    missing: int = 0
+    conflict: int = 0
+    unknown: int = 0
+    applied: int = 0
+    failures: int = 0
+
+
+@never_cache
+@staff_member_required
+def admin_hosted_dkim_backfill(request: HttpRequest):
+    if not request.user or not request.user.is_authenticated or not request.user.is_superuser:
+        raise Http404()
+
+    response = StreamingHttpResponse(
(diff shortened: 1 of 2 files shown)


MelissaAutumn:
Just curious is there a reason why this can't be an admin action? I'd like to try and contain any data migrationty stuff. 
https://docs.djangoproject.com/en/6.0/ref/contrib/admin/actions/

markstos:
> Just curious is there a reason why this can't be an admin action? 

It seems the expected usage model for DJango actions is "select a bunch of items, then apply an action". Here, I don't think there's a clean way to select the list of objects to act on, because we have iterate over the Stalwart DKIM and Cloudflare records to figure out what's needs to sync'ed. Is there a way to use Django actions for a case like ours?

> I'd like to try and contain any data migrationty stuff.

By contain, do you mean formally represent in Django as one of their actions or migrations? A Django migration doesn't seem to fit well here either, as we are primarily updating Cloudflare, not a Django model.

Sancus:
There are definitely potential nits here but given we're running this on a small number of domains and it's temporary I don't see any showstoppers and neither does Fable.

PR #1103, opened 2026-07-11
Fix blank SPF host in local development

### LINKED ISSUE #887: Fix blank spf host for local development
Currently the SPF host name is defined as a sibling to the `dns_hostname`. When the DNS hostname is `localhost` for local development, this ends up blank:

```
src/thunderbird_accounts/mail/clients.py
688:        dns_top_host = '.'.join(dns_hostname.split('.')[1:])
725:                'content': f'v=spf1 include:spf.{dns_top_host} -all',
801:            dns_top_host = '.'.join(expected_host.split('.')[1:])
802:            expected_spf_include = f'include:spf.{dns_top_host}'
```

One possible fix is to explicitly define the `spf` full DNS name in configuration instead of assuming it's `spf.$dns_tophost`

## Workaround

 * In /etc/hosts, add a new host entry for localhost that includes a dot.
 * In `.env`, search for IMAP_HOST, SMTP_HOST and JMAP_HOST and replace this this value. 
 * `docker compose restart accounts`

Kkkakania:
## What changed?

SPF include records now use an explicit `SPF_HOST` setting instead of deriving the host from `SMTP_HOST`. Local development uses `spf.localhost`, while stage and production retain their existing effective values (`spf.stage-thundermail.com` and `spf.thundermail.com`). The variable is included in both Pulumi task definitions and their rendered environment lists.

Regression coverage verifies that the SPF host can differ from the SMTP domain and that `SMTP_HOST=localhost` produces `include:spf.localhost` rather than `include:spf.`.

AI disclosure: this contribution and PR text were produced by OpenAI Codex with Kkkakania's authorization. Codex inspected the current implementation, wrote the focused tests and change, and ran the validation listed below.

Validation:

- `uv run ruff check .`
- `uv run python manage.py test thunderbird_accounts.mail.tests --verbosity 0` (170 tests passed)
- `uv run python manage.py test thunderbird_accounts --verbosity 0` after the CI-equivalent Vite build (501 tests passed)
- both Pulumi config files parsed with PyYAML

## Why?

When `SMTP_HOST=localhost`, removing the first hostname label leaves an empty string and generates the invalid SPF include `spf.`. A dedicated setting represents the actual deployment contract and removes the assumption that the SPF host is always a sibling of the SMTP host.

## Limitations and Notes

No fallback derivation is retained: every checked-in environment now declares `SPF_HOST` explicitly.

## Applicable Issues

Closes #887

## Screenshots

Not applicable; this is a configuration and DNS-record generation change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,6 +50,7 @@ IMAP_TLS=False
 SMTP_HOST=localhost
 SMTP_PORT=25
 SMTP_TLS=False
+SPF_HOST=spf.localhost
 
 JMAP_HOST=localhost
 JMAP_PORT=443
--- a/.env.test
+++ b/.env.test
@@ -46,6 +46,7 @@ IMAP_TLS=True
 SMTP_HOST=localhost
 SMTP_PORT=465
 SMTP_TLS=True
+SPF_HOST=spf.localhost
 
 JMAP_HOST=localhost
 JMAP_PORT=443
--- a/pulumi/config.prod.yaml
+++ b/pulumi/config.prod.yaml
@@ -56,6 +56,7 @@
 .smtp_host: &VAR_SMTP_HOST {name: "SMTP_HOST", value: "mail.thundermail.com"}
 .smtp_port: &VAR_SMTP_PORT {name: "SMTP_PORT", value: "465"}
 .smtp_tls: &VAR_SMTP_TLS {name: "SMTP_TLS", value: "True"}
+.spf_host: &VAR_SPF_HOST {name: "SPF_HOST", value: "spf.thundermail.com"}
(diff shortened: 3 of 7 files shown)


markstos (approved):
Agreed, this is a logical way to resolve the issue.

@ryanjjung Anything else we need do on the Pulumi / operational side here pre-merge?

## QA Log

  - Reviewed issue, diff, validation
  - Confirmed that `spf.thundermail.com` and `spf.stage-thundermail.com` are the correct production values.

ryanjjung:
I'll defer to your wisdom on the fix itself. As far as getting an environment variable into our running containers, this looks right. The various `.env` files for local dev have been updated. The task definitions for stage and prod have as well. After merge, please check the deployment and test this change when it's live in stage. Looks fine, though.

markstos:
This has been deployed to staging and confirmed to work there.

Issue #884, opened 2026-05-29
Stage deploy: merge accounts-deploy and keycloak-deploy into a single Pulumi update to fix [409] stack-lock conflicts


aatchison:
## Problem

When a change touches IaC (`pulumi/**` or `.github/**`), both the `accounts-deploy` and `keycloak-deploy` jobs in `.github/workflows/merge.yml` run in parallel (both only `needs: detect-changes`). Each job runs `pulumi up` against the **same** stack (`thunderbird/stage`).

Pulumi takes a **stack-wide lock** for the duration of an update — it does not lock per-resource. So even though the two jobs `--target` different resource URNs, whichever job reaches `pulumi up` second fails with:

```
error: [409] Conflict: Another update is currently in progress.
```

This is **not** a conflicting-resource error; it's a concurrent-update lock collision.

### Evidence from run history

Reviewing the last ~40 `deploy-stage` runs, **every** run where both jobs ran concurrently had one fail on the lock. `accounts-deploy` is almost always the loser (it builds/pushes more before reaching `pulumi up`, so it grabs the lock second):

| Run | keycloak-deploy | accounts-deploy |
|-----|-----------------|-----------------|
| 26648551142 | success | **failure** |
| 26413289325 | failure | **failure** |
| 26288805751 | success | **failure** |
| 25581480211 | success | **failure** |
| 25520352029 | success | **failure** |

No run in that window deployed both successfully in a single run. The failure reproduces 100% of the time the two jobs compete.

## Proposed fix

Merge `accounts-deploy` and `keycloak-deploy` into a single `deploy` job that performs **one** `pulumi up` against `thunderbird/stage` with all relevant `--target` URNs. One update ⇒ one lock ⇒ no conflict.

- Build the accounts image only when `src-changed || iac-changed`.
- Build the keycloak image only when `keycloak-theme-changed || iac-changed`.
- Update the corresponding `config.stage.yaml` image keys and include the corresponding `--target` URNs only for whichever image(s) were rebuilt.
- Preserve the downstream `create-release` (accounts-only) and `e2e` gating.

## Acceptance criteria

- A single `pulumi up` runs per stage deploy (no parallel stack updates).
- IaC-only changes deploy both accounts + keycloak taskdefs without a 409.
- `create-release` still only fires when the accounts image was built/deployed.
- Prod promotion (`release.yml`) is unaffected.

PR #1109, opened 2026-07-13
ci(deploy-stage): force keycloak/accounts builds + stop concurrency dropping deploys

### LINKED ISSUE #1107: deploy-stage: allow forcing keycloak builds + avoid cancel-in-progress dropping keycloak deploys
Follow-up from review of #1087 (@MelissaAutumn).

`merge.yml` (deploy-stage) has two related gaps:

1. **No way to force a keycloak build/deploy.** Builds are gated purely by `detect-changes`; there's no manual override.
2. **cancel-in-progress edge case.** `concurrency: { group: deploy-stage, cancel-in-progress: true }` means a keycloak-aware run can be cancelled by a later non-keycloak run if someone merges to `main` before the first run finishes — so a keycloak change can be silently not built/deployed.

### Proposed
- Add a `workflow_dispatch` input (e.g. `force_keycloak` / `force_accounts`) to `merge.yml` to force builds regardless of change detection.
- Reconsider `cancel-in-progress` for deploy-stage, or ensure a cancelled run's pending image builds are not lost.

### Note
The new GHCR publish pipeline added in #1087 (`publish-images.yml`) already avoids both issues: it uses `cancel-in-progress: false` and its `workflow_dispatch` force-builds both accounts and keycloak. This issue tracks bringing the equivalent behavior to the deploy pipeline.

aatchison:
## What

Addresses both gaps in `merge.yml` (deploy-stage) called out in the #1087 review.

### 1. Force builds via `workflow_dispatch`

Adds two boolean inputs, mirroring the force behavior already in `publish-images.yml`:

- `force_accounts` — force building & deploying the accounts image
- `force_keycloak` — force building & deploying the keycloak image

They are wired into:

- the `deploy` job `if:` gate (so the job runs even when change-detection finds nothing), and
- the `BUILD_ACCOUNTS` / `BUILD_KEYCLOAK` env expressions (so the corresponding image build step, config merge, pulumi `--target`s, and artifacts all fire).

Each input independently forces only its own image, so a manual run can rebuild keycloak alone without touching accounts. Comparisons use `github.event.inputs.<name> == 'true'`, which is empty (falsey) on `push`, so normal merge-to-main behavior is unchanged.

### 2. Concurrency: stop silently dropping deploys

Changed the `deploy-stage` group from `cancel-in-progress: true` to `cancel-in-progress: false`.

**Decision + tradeoff:** the group is kept as a single global `deploy-stage` group (not narrowed per-ref). The deploy job runs a single `pulumi up` and Pulumi holds a stack-wide lock (see the deploy-job comment / #884), so narrowing the scope to allow parallel deploys would just make concurrent runs collide on that lock (`[409] Conflict`). Keeping one global group with `cancel-in-progress: false` therefore serializes deploys: every merged commit's run is queued and allowed to finish rather than being cancelled. This fixes the observed #1087 edge case where a Dependabot merge cancelled an in-flight keycloak-aware run, and the replacement run then change-detection-skipped the deploy — so that commit never reached stage. Tradeoff: deploys now queue (slightly slower to converge under rapid merges) instead of cancelling, in exchange for never dropping a deploy. This matches the `cancel-in-progress: false` already used by `publish-images.yml`.

## Validation

- `python -c "import yaml; yaml.safe_load(...)"` — valid
- `actionlint .github/workflows/merge.yml` — clean

Closes #1107

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -3,13 +3,31 @@ name: deploy-stage
 
 concurrency:
   group: deploy-stage
-  cancel-in-progress: true
+  # Serialize (don't cancel) deploys. A single global group keeps only one
+  # `pulumi up` in flight at a time (Pulumi holds a stack-wide lock -- see the
+  # deploy job comment / issue #884), and cancel-in-progress: false ensures a
+  # later run can't abort an in-flight deploy. Previously cancel-in-progress:
+  # true let a Dependabot merge cancel an in-flight keycloak-aware run, whose
+  # replacement then change-detection-skipped -- so that commit never deployed
+  # to stage (observed on #1087). Narrowing the group per-ref instead would let
+  # deploys run in parallel and collide on the Pulumi stack lock, so we keep the
+  # single group and queue runs.
+  cancel-in-progress: false
 
 on:
   push:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      force_accounts:
+        description: "Force building & deploying the accounts image, regardless of change detection"
+        type: boolean
+        default: false
+      force_keycloak:
+        description: "Force building & deploying the keycloak image, regardless of change detection"
+        type: boolean
+        default: false
 
 permissions:
   id-token: write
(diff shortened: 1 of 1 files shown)


markstos (commented):
Seems good to try to me based on the description and diff. 

Leaving for Mel to give a pass well.

aatchison:
Any love here?

MelissaAutumn (approved):
Sorry, I'm slow 🐌 . Looks good, thank you!

PR #1110, opened 2026-07-14
Replace Thunderbird Pro with Thundermail logo + UserMenu dropdown for both pages

### LINKED ISSUE #1094: Replace Thunderbird Pro in NavBar for Accounts
We should replace the Thunderbird Pro logo with the Mail logo for now.

Ideally, as part of this issue, we should also come up with a solution for the UserAvatar navigation since it might be extra confusing to move from Mail <-> Accounts and vice-versa.

This ticket is tracking a temporary solution for now as the Design team will come up with a more permanent solution (cc @laurelterlesky @thunderbecca)

davinotdavid:
## What changed?

- Replaced the Thunderbird Pro logo with the Thundermail logo for the dashboard / Accounts in both header and footer.
- The logo links to `/mail` in all cases now.
- UserAvatar component is always a dropdown now with a permanent option of `My Account` to navigate back to dashboard / Accounts.
- Added `WEBMAIL_URL` env var + notice-bar in Mail dashboard

## Why?

- We are moving away from the Thunderbird Pro term.
- Current navigation has been flagged in recent conversations as being confusing. This is an attempt of making the move from Accounts <-> Mail a bit more predictable / stable.

## Limitations and Notes

- This is a bandage / temporary solution for the case. Design has yet to chime in on how to fix this and I assume that there will be changes in the coming weeks as we are increasingly have more important pages being added (MFA, Security Settings, etc).

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/1094

## Screenshots

Mail Dashboard
image

Accounts Dashboard
image


davinotdavid:
Verbally approved by @Alecaddd @inattee and Brian!

PR #1112, opened 2026-07-14
Update docs and add an admin faq. Additionally add a checkbox to send password resets.

MelissaAutumn:
My brain is mush today, could I get a vibe check on the copy changes?

I also noticed the docs were a little outdated, so I added a checkbox to restore some functionality that was removed a while ago. Previously at some point in time and space we automatically sent a "reset password" email to any user created via admin panel. That was changed to not do that. For folks who are blocked by the reserver word list we do want that functionality back. So added it back. 

...one day we'll change that readthedocs url too, but that's not today.


docs/admin/faq.rst
```diff
@@ -0,0 +1,8 @@
+==============================================
+FAQ
+==============================================
+
+How can I create a thundermail address that's blocked by the reserved word list?
+--------------------------------------------------------------------------------
+
+Any user created on the admin panel automatically bypasses the reserved word list. Simply add the user as documented in :doc:`./users` and make sure the ``Send "Rest Password" email`` field is checked.
```
MelissaAutumn, line 8: RST is kinda annoying to write in, but sphinx has some handy features that I'm not sure if markdown supports?

markstos: Now Sphinx also supports the MyST parser, which allows Markdown + Sphinx extensions -- a number of the features that RST supports but Markdown doesn't.

markstos (approved):
Looks good, though I'd prefer to try MyST (Markdown + Sphinx extensions ) rather than RST, as I use Markdown broadly and  `.rst` nowhere else currently.

Issue #605, opened 2026-02-17
Validate domain names


MelissaAutumn:
This is entirely user error, and currently nothing bad will happen with an invalid custom domain in the unverified state, but we should definitely try to avoid friction where possible. 

A domain name should probably have at least one period in it. There's plenty of validators available in the wild like one built into django: https://docs.djangoproject.com/en/6.0/ref/validators/#domainnamevalidator

markstos:
I started working on this after I noticed that something bad *did* happen due to our failure to validate domains: a DNS backfill automation (#1079) broke due to a GIGO bug (Garbage-In-Garbage-Out), because the automation assumed that domain names are... valid domain names, after hitting one that wasn't the automation stopped halfway through. 

In terms of /which/ validator to use, I'll check-out the one in Django as well what Stalwart is going. If our validator allows domains that Stalwart then rejects, we'll continue to have problems. If Django's validator and Stalwart's don't differ much, let's go with Django's. If they differ much, let's go with Stalwart's approach.

PR #1114, opened 2026-07-14
Add domain validation (#605)

### LINKED ISSUE #105: E2E test failure when run on stage

markstos:
## What changed?

Custom Domain name validation was added.

## Why?

This was spurred by #1079 where an invalid domain got into the system and caused problems.

## Limitations and Notes

To choose exactly how to validate the domain, but Stalwart's domain validation code was analyzed along with Django. They turned out to be highly compatible, so Django's built-in domain validation was used for simplicity.

We can expect Stalwart will also find valid domains validated this way.

## Applicable Issues

* Fixes #105, References #1079

## Screenshots

image

## QA log

 - Compared Django algo to Stalwart
 - Automated test coverage
 - Manually tested positive and negative cases. Seem screenshot.

### DIFF
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
@@ -75,49 +75,6 @@ const spfIncludeValue = (record: DNSRecord): string =>
 const normalizeDomainName = (domainName?: string | null): string =>
   (domainName ?? '').trim().replace(/\.$/, '').toLowerCase();
 
-const normalizeCustomDomainInput = (domainName?: string | null): string =>
-  (domainName ?? '').replace(/\s+/g, '').toLowerCase();
-
-const domainLabelPattern = /^[a-z0-9\u00a1-\uffff](?:[a-z0-9\u00a1-\uffff-]{0,61}[a-z0-9\u00a1-\uffff])?$/i;
-const domainTldPattern = /^(?:[a-z\u00a1-\uffff-]{2,63}|xn--[a-z0-9]{1,59})$/i;
-
-const hasValidDomainTld = (label: string): boolean =>
-  label.length > 0
-  && label[0] !== '-'
-  && label[label.length - 1] !== '-'
-  && domainTldPattern.test(label);
-
-const isValidCustomDomainName = (domainName: string): boolean => {
-  if (!domainName.includes('.') || domainName.endsWith('.') || domainName.length > 255) {
-    return false;
-  }
-
-  const labels = domainName.split('.');
-  const tld = labels.pop();
-
-  return Boolean(tld && hasValidDomainTld(tld) && labels.every((label) => domainLabelPattern.test(label)));
-};
-
-const normalizedCustomDomain = computed(() => normalizeCustomDomainInput(customDomain.value));
-
-const isCustomDomainValid = computed(() => isValidCustomDomainName(normalizedCustomDomain.value));
-
-const customDomainInputError = computed(() => {
-  if (customDomainError.value) {
-    return customDomainError.value;
-  }
-
-  if (customDomain.value && !isCustomDomainValid.value) {
-    return t('views.mail.sections.customDomains.invalidDomain');
-  }
-
-  return null;
-});
-
-const canSubmitCustomDomain = computed(() =>
-  Boolean(customDomain.value) && isCustomDomainValid.value && !isAddingCustomDomain.value
-);
-
 const dkimRecordDomain = (record: DNSRecord): string | null => {
   if (record.type !== 'CNAME') {
     return null;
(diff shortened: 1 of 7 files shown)


davinotdavid (commented):
Hmm seems to work well but why are we doing the validation on the frontend as well? Can't we rely solely on the backend errors and get the return from the `addCustomDomain` call?

markstos:
> Hmm seems to work well but why are we doing the validation on the frontend as well? Can't we rely solely on the backend errors and get the return from the `addCustomDomain` call?

Having the validation on the frontend makes the validation immediately responsive at the expense of adding some more FE JavaScript to maintain. You are right that the FE could delegate the the validation backend and not display validation until the form is submitted. Which pattern do you recommend here?

davinotdavid:
@markstos I'd recommend keeping this validation only in the backend.

We already have other validations in the backend so I think the time savings of the frontend feedback just for this one validation is negligible. I'd lean towards less maintenance / error prone on maintaining both places.

markstos:
I'll refactor along those lines @davinotdavid and keep that principle in mind for future work.

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
```diff
@@ -417,38 +417,38 @@ const handleDNSRecords = async (domainName: string) => {
...
 
   try {
-    const data = await addCustomDomain(customDomain.value);
+    const data = await addCustomDomain(submittedDomain);
 
     if (data.success) {
-      emit('custom-domain-added', customDomain.value);
-      await handleDNSRecords(customDomain.value);
+      const domainName = data.domain_name ?? submittedDomain;
+      customDomain.value = domainName;
```
davinotdavid, line 437: Assuming that this is being done so that the `watch` can clear up the fields, could we instead remove these two lines and add the clear up of errors (`customDomainError` and `domainAlreadyConfigured`) here instead?

markstos:
@davinotdavid I have a refactored to use backend-only validation and re-tested the change manually.

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
```diff
@@ -417,38 +417,38 @@ const handleDNSRecords = async (domainName: string) => {
...
-    customDomainError.value = error;
+    customDomainError.value = String(error);
   }
 }
 
 const onCreateCustomDomain = async () => {
-  if (props.customDomains.some((domain) => domain.name === customDomain.value)) {
-    customDomainError.value = t('views.mail.sections.customDomains.domainAlreadyExists');
-    return;
-  }
```
davinotdavid, line 432: Good catch on removing this!

davinotdavid (approved):
lgtm, thanks! we might want to update the PR title to remove the frontend aspect of it.

## QA Log

- Code diff
- Tested manually

assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
markstos: Now that we've removed FE validation, this seems like a nice simplification of error handling. Good catch. I've pushed a commit to address it.

PR #1116, opened 2026-07-15
Refresh OIDC access tokens before fixing archive folders (#1106)

### LINKED ISSUE #1106: Address frequent exception: "fix_archive_folder failed!"
Sentry Issue: [THUNDERBIRD-ACCOUNTS-5Y](https://thunderbird.sentry.io/issues/7088369914/?referrer=github_integration)

```
fix_archive_folder failed!
```

markstos:
## What changed and why? 

The MFA management flow already checked the stored Keycloak access token's
expiration and refreshed it.

Here we move that freshness check into a shared helper and use it from
fix_archives_folder() before making JMAP calls.

This avoids forwarding stale access tokens to Stalwart, which was producing
frequent 401s in Sentry from the archive-folder repair path.

## Notes

We also tighten the behavior: If refresh fails, "fail fast" instead
of passing a known-stail token downstream.

If there was a specific reason the MFA code passed along expired tokens rather than
considering them failed, we can restore that behavior. My read that Keycloak was
going to consider them failed on the next hop, so it was clearer to treat them
as failed earlier.

## Applicable Issues

* Fixes #1106

## QA Log

 * Automated testing
 * Manual re-testing locally of MFA onboarding flow for regressions

### DIFF
--- a/src/thunderbird_accounts/authentication/mfa_management.py
+++ b/src/thunderbird_accounts/authentication/mfa_management.py
@@ -3,7 +3,6 @@ import time
 from dataclasses import dataclass
 from urllib.parse import urlencode, urlparse
 
-import jwt
 import sentry_sdk
 from django.conf import settings
 from django.core.cache import cache
@@ -26,7 +25,7 @@ from thunderbird_accounts.authentication.mfa import (
     has_recent_mfa_management_auth,
     make_pending_totp_cache_key,
 )
-from thunderbird_accounts.authentication.middleware import OIDC_ACCESS_TOKEN_KEY, refresh_user_access_token
+from thunderbird_accounts.authentication.tokens import get_user_access_token
 
 
 class MfaManagementError(Exception):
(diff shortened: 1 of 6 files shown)


Sancus:
This introduced a lot of test failures that need to be fixed from what I can see.

MelissaAutumn:
@markstos I think part of this will be resolved with https://github.com/thunderbird/thunderbird-accounts/pull/1054 but I haven't tested as it's possible tokens aren't being refreshed on POST routes (which is the majority of the server requests.) 

If that other ticket doesn't resolve the issue, once tests are fixed here we can review this. 

As a quicker more-temp fix, do you have some time to wrap the `account_id = client.get_account_id()` in fix_archives_folder in a try/except and log a warning and continue along without erroring out? It's a large chomp out of sentry quota :scream_cat:.

markstos:
I have fixed the tests here.

Sancus:
Unless I'm missing something, this PR doesn't really make any sense after #1054 -- the OIDC middleware runs before archive-folder so the token will be refreshed automatically.

markstos:
Agreed, this is largely superseded by #1054. I've left for Mel on Matrix about enabling the waffle flags to fully activate that fix.

PR #1118, opened 2026-07-15
Check for required fields before calling Zendesk (#1117)

### LINKED ISSUE #1117: Failed to create Zendesk ticket: <Response [422]>
Sentry Issue: [THUNDERBIRD-ACCOUNTS-62](https://thunderbird.sentry.io/issues/7094026264/?referrer=github_integration)

```
Failed to create Zendesk ticket: <Response [422]>
```

markstos:
This keeps us from calling Zendesk with data that would be rejected.

## Why?

This addresses errors we were getting with Zendesk rejecting empty request
with HTTP 422 status.

## Notes

Had we used Django's standard Forms API for form validation, it would
have caught that issue.

A larger refactor is recommended to switch to that design.

What we continue to do here is to post the key data as a JSON structure instead of as form fields, bypassing the ability to use Django's standard form processing methods.

## Applicable Issues

 * Fixes #1117 

## QA Log

 * It appears the frontend was already doing some validation, but apparently bots or other user agents were posting directly to the backend, which triggered the exceptions.
 * Added automated tests.

### DIFF
--- a/src/thunderbird_accounts/support/tests/test_views.py
+++ b/src/thunderbird_accounts/support/tests/test_views.py
@@ -205,6 +205,37 @@ class ZendeskContactSubmitTestCase(TestCase):
         mock_client_cls.assert_not_called()
 
     @patch('thunderbird_accounts.support.views.ZendeskClient')
+    def test_contact_submit_rejects_missing_required_zendesk_fields_before_upload(self, mock_client_cls):
+        url = reverse('contact_submit')
+        payload = {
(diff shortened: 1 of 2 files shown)


davinotdavid (approved):
lgtm, tested locally using cURL and manually getting myself a CSRF token (interesting that the bots are doing that).

also, if we move to Django Forms, I wonder if we would have the same control over the UI / UX as iirc this was one of the motivation to do it on the frontend + API call.

PR #1120, opened 2026-07-16
Update dashboard tests to verify app content loads after clicking service icons

### LINKED ISSUE #1092: Update E2E tests to ensure Send/Appt apps content load after clicking nav buttons
The existing TB Accounts E2E tests navigate to the dashboard (https://accounts.tb.pro/dashboard) and click on each of the `Thundermail`, `Send`, and `Appointment` buttons, however they only verify a new tab is opened with the correct URL to the associated app, and they don't wait for the content to actually load.

There was a bug where after clicking the `Send` icon from the main Accounts dashboard page a 404 message was displayed in the Browser instead of the Send app (#1089). The existing E2E tests didn't catch that because they were clicking the Send button/link and verifying the URL but not waiting for and verifying Send app content was actually loaded.

rwood-moz:
## What changed?
Updating the dashboard E2E tests to verify that app/page content actually loads after clicking on the Thundermail/Appointment/Send service icons on the dashboard page.

## Why?
There was an issue in production where clicking on the Send icon on the dashboard page resulted in the Send URL opening but a 404 error was displayed. The test previously just ensured the correct URL was loaded, but now let's actually verify the app page content loads.

## Limitations and Notes
Only checks one element on each of the app pages after clicking the icons. This is sufficient for this accounts dashboard test as the other app pages are testing in detail in their corresponding E2E tests in their own repos.

Worked with Codex AI to write this patch.

## Applicable Issues
Fixes #1092.

## QA Log
Ran the new updated dashboard test:

- On my local Firefox desktop
- On my local playwright google pixel 7 viewport simulator
- On Firefox desktop on OSX [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Nightly+Tests+Firefox+Desktop/168?tab=sessions&testListView=spec)
- On Chromium desktop on Win11 [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Nightly+Tests+Chromium+Desktop/161?tab=sessions)
- On Safari desktop on OSX [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Nightly+Tests+Safari+Desktop/81?tab=sessions&testListView=spec)
- On Android Chrome on a real Google Pixel 10 device [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Accounts/builds/TB+Accounts+Nightly+Tests+Android+Chrome/100?testListView=spec)

### DIFF
--- a/test/e2e/pages/dashboard-page.ts
+++ b/test/e2e/pages/dashboard-page.ts
@@ -12,6 +12,7 @@ import {
   TIMEOUT_3_SECONDS,
   TIMEOUT_5_SECONDS,
   TIMEOUT_30_SECONDS,
+  TIMEOUT_60_SECONDS,
 } from '../const/constants';
 import { waitForVueApp } from '../utils/utils';
 
@@ -25,6 +26,10 @@ type PopupPageAssertion = {
   expectedUrl: string;
   expectedElementName: string;
   expectedElement: (page: Page) => Locator;
+  beforeExpectedElements?: Array<{
+    expectedElementName: string;
+    expectedElement: (page: Page) => Locator;
+  }>;
 };
 
 const escapeRegExp = (value: string) => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
@@ -134,12 +139,20 @@ export class DashboardPage {
     await this.verifyPopupServiceAppLoads({
       link: this.appointmentLink,
       expectedUrl: serviceUrls.appointment,
+      beforeExpectedElements: [{
+        expectedElementName: 'Need help? Visit support',
+        expectedElement: page => page.locator('footer').getByRole('link', { name: /visit support/i }),
+      }],
       expectedElementName: 'Copy booking link',
       expectedElement: page => page.getByRole('button', { name: /copy booking link/i }),
     });
     await this.verifyPopupServiceAppLoads({
       link: this.sendLink,
       expectedUrl: serviceUrls.send,
+      beforeExpectedElements: [{
+        expectedElementName: 'Need help? Visit support',
+        expectedElement: page => page.locator('footer').getByRole('link', { name: /visit support/i }),
+      }],
       expectedElementName: 'Recover access',
       expectedElement: page => page.getByTestId('recover-access-button'),
     });
@@ -189,6 +202,7 @@ export class DashboardPage {
   private async verifyPopupServiceAppLoads({
     link,
     expectedUrl,
+    beforeExpectedElements = [],
     expectedElementName,
     expectedElement,
   }: PopupPageAssertion) {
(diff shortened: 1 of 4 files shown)


test/e2e/pages/dashboard-page.ts
```diff
@@ -109,15 +117,32 @@ export class DashboardPage {
...
+      link: this.appointmentLink,
+      expectedUrl: serviceUrls.appointment,
+      expectedElementName: 'Copy booking link',
+      expectedElement: page => page.getByRole('button', { name: /copy booking link/i }),
+    });
+    await this.verifyPopupServiceAppLoads({
+      link: this.sendLink,
+      expectedUrl: serviceUrls.send,
+      expectedElementName: 'Recover access',
+      expectedElement: page => page.getByTestId('recover-access-button'),
```
davinotdavid, line 157: Unless we can make sure that the account that is running the test is in the recover state, I believe this might be a bit brittle as the recover access button may not be there for all accounts.

For instance, when I login with my prod account in Accounts and click the Send button I get this:

Image

Also, [the intended change from the PR that originated this issue](https://github.com/thunderbird/thunderbird-accounts/pull/1040/changes) was attempting to load a different page (dashboard instead of encrypted files) which would make a change in Send fail E2E tests in Accounts.

Perhaps it would be best to check if the send logo is there + it is _not_ a "Not found" page?

test/e2e/pages/dashboard-page.ts
```diff
@@ -109,15 +117,32 @@ export class DashboardPage {
...
     const serviceUrls = await this.getConfiguredServiceUrls();
-    await this.verifyLinkOpensPopup(this.appointmentLink, serviceUrls.appointment);
-    await this.verifyLinkOpensPopup(this.sendLink, serviceUrls.send);
+    // Thundermail is an internal Accounts route and reuses the current page, while
+    // Appointment and Send are configured external services that open in popups.
+    await this.verifyPopupServiceAppLoads({
+      link: this.appointmentLink,
+      expectedUrl: serviceUrls.appointment,
+      expectedElementName: 'Copy booking link',
+      expectedElement: page => page.getByRole('button', { name: /copy booking link/i }),
```
davinotdavid, line 147: This one is a lot more stable as we don't have any foreseeable changes coming in, however I'd also say that we can instead do the same approach as the one in the comment about the Send assertions?

It might require tweaking in Send / Appointment as they currently have different 404 elements to assert from though!

davinotdavid (commented):
Nice, thanks for the work! Commenting instead of requesting changes because I want to double check first if the idea makes sense.

test/e2e/pages/dashboard-page.ts
rwood-moz: Hey @davinotdavid great observations, thanks for the review. So in BrowserStack every test is ran in a brand new Browser and browser context, which means the Send keys must be restored every time; so it is guaranteed when the TB Accounts dashboard test runs, that the Send keys have not been restored yet. That also means after clicking the Send icon in the TB Accounts dashboard, the TB Send dashboard (with the `Recover Access` button) is displayed because the keys must be restored (you can't even click on the Encrypted Files page).

test/e2e/pages/dashboard-page.ts
rwood-moz: I'd prefer to leave this as/is because this works successfully on both desktop and mobile (android chrome), I'd rather not attempt to check for 404 elements (and also what if there's a different error besides 404/Not found).

test/e2e/pages/dashboard-page.ts
MelissaAutumn: If we're just checking send is alive and showing a page, then maybe checking for one of the footer links might be the way to go. (Or the site logo)

rwood-moz: I thought it would be important to ensure the `Recover Access` button was available since that is a required action, but I'll also add a second element check also for one of the footer links (that appears on both desktop and mobile).

MelissaAutumn: Sounds good, make that check first and then the recover access button second. That way we can see if it's only a partial failure. Thanks!

davinotdavid (approved):
lgtm, thanks!

PR #1122, opened 2026-07-20
Add purge-incomplete-signups waffle switch

davinotdavid:
## What changed?

- Adds a [django-waffle switch](https://waffle.readthedocs.io/en/stable/types/switch.html) to enable / disable the purge of users. 

## Why?

- In case something unexpected happens, it would be nice to be able to toggle the purge on / off through Django Admin without having to change code / make a PR.

## Limitations and Notes

After this lands, we will need to create a `purge-incomplete-signups` switch through click ops in Django Admin and turn it on if / when need be.

## Applicable Issues

Related with https://github.com/thunderbird/thunderbird-accounts/issues/964

### DIFF
--- a/src/thunderbird_accounts/authentication/tasks.py
+++ b/src/thunderbird_accounts/authentication/tasks.py
@@ -3,6 +3,7 @@ import logging
 from datetime import timedelta, datetime, UTC
 
 import sentry_sdk
+import waffle
 from celery import shared_task
 from django.conf import settings
 from django.db import transaction
@@ -10,7 +11,7 @@ from django.db.models import Exists, OuterRef, QuerySet
 from django.utils import timezone
 
 from thunderbird_accounts.authentication.models import User, AllowListEntry
-from thunderbird_accounts.authentication.utils import delete_user_data  #  noqa: F401 - needed for tests
+from thunderbird_accounts.authentication.utils import delete_user_data
 from thunderbird_accounts.celery.exceptions import TaskFailed
 from thunderbird_accounts.subscription.mailchimp import MailchimpClient
 from thunderbird_accounts.subscription.models import Subscription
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -194,11 +202,12 @@ def purge_incomplete_signups(self):
                     )
                     continue
 
-                # Add the user to the group
-                user.groups.add(group)
-                purge_errors = []
-                # TODO: Turn back on once we're confident
-                # purge_errors = delete_user_data(user)
+                if waffle.switch_is_active(PURGE_INCOMPLETE_SIGNUPS_SWITCH):
```
MelissaAutumn: Can we have a log entry to announce what the switch status is? 

Like 

```python
switch_status = waffle.switch_is_active(PURGE_INCOMPLETE_SIGNUPS_SWITCH)
logger.info(f'Running with {PURGE_INCOMPLETE_SIGNUPS_SWITCH} = {switch_status}')
```

MelissaAutumn (approved):
Looks good, thanks! One suggestion but can be merged after.

src/thunderbird_accounts/authentication/tasks.py
```diff
@@ -194,11 +202,15 @@ def purge_incomplete_signups(self):
                     )
                     continue
 
-                # Add the user to the group
-                user.groups.add(group)
-                purge_errors = []
-                # TODO: Turn back on once we're confident
-                # purge_errors = delete_user_data(user)
+                switch_status = waffle.switch_is_active(PURGE_INCOMPLETE_SIGNUPS_SWITCH)
+                logger.info(f'Running with {PURGE_INCOMPLETE_SIGNUPS_SWITCH} = {switch_status}')
```
MelissaAutumn: Erm, sorry I meant outside of the loop probably at the top of the function. 😅

davinotdavid: homer fading into bush

PR #1134, opened 2026-07-21
Fix MFA login timeout loop (#1133)

### LINKED ISSUE #1133: MFA login timeout causes subsequent valid login attempts to fail
## Describe the bug

For accounts with MFA, when you initially have a login attempt that has timed out, all subsequent login attempts will be denied and a timeout error message will be displayed.

## To Reproduce

1. Navigate to https://www.tb.pro/en-US/
2. Click the **SIGN IN** button
3. Enter credentials
4. On the next **Sign In** webpage you will presented with the textbox: **One-time code***
5. Wait for a few minutes, i.e. create a delay to cause a timeout
6. Attempt to log in again, i.e. repeat steps 1-4 again
7. This time, enter your new MFA code
8. You will NOT be logged in and will be presented with the following error even though your credentials and MFA code were correct: **Your login attempt timed out. Login will start from the beginning.**
9. Subsequent successful log in attempts after the initial timeout will all be met with the timeout error

## Expected behavior

After a timeout login attempt has occurred, subsequent successful login attempts should be logged in.

## Actual behavior

After a timeout login attempt has occurred, all subsequent successful login attempts are denied (i.e. the user is unable to log in) and the timeout error message is shown on each attempt.

## System

- OS: 
Windows 11
Chrome Version 150.0.7871.129 (Official Build) (64-bit)

## Additional context

Work-arounds
1. Clear the browser cache, history...
2. Use another browser

markstos:
Since we implmented MFA, the Keycloak login forms started bypassesing the
native submit event that Keycloak's authChecker uses to stop session polling.

After a timed-out MFA challenge, that poller could keep running during the next
successful submit and redirect the browser back into the expired login flow,
making it impossible to login without refreshing the browser state.

Here we restore the logic to stop polling Keycloak upon form submission
that was accidentally lost during MFA impementation.

## Applicable Issues

 * Fixes #1133

## QA Log

 * Used AI to pinpoint the regression.
 * Manually tested MFA setup and login after this flow.
 * I did not wait the 30 minutes during MFA login to repro the issue exactly.

### DIFF
--- a/keycloak/themes/tbpro/assets/vue/components/CancelForm.vue
+++ b/keycloak/themes/tbpro/assets/vue/components/CancelForm.vue
@@ -1,5 +1,6 @@
 <script setup>
 import {ref, useTemplateRef, watch} from "vue";
+import { submitKeycloakForm } from '@kc/vue/keycloakForm';
 
 defineProps({
   action: 'string',
@@ -20,7 +21,7 @@ const cancel = () => {
   watch(
     cancelForm,
     () => {
-      cancelForm?.value?.submit();
+      submitKeycloakForm(cancelForm?.value);
     }, {once: true});
 
   // This will cause the cancelForm not be null
--- a/keycloak/themes/tbpro/assets/vue/keycloakForm.ts
+++ b/keycloak/themes/tbpro/assets/vue/keycloakForm.ts
@@ -1,0 +1,8 @@
+type KeycloakWindow = Window & {
+  stopKeycloakSessionPolling?: () => void;
+};
+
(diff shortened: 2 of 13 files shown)


Sancus:
I am not entirely sure this is a real bug or if that's the bug. Yes, AI suggests this, but I can't reproduce the bug, not even leaving the MFA window open overnight reproduced it.

AI was also not able to reproduce this bug or test that it was fixed via actual in-browser testing.

markstos:
If we can't reproduce it and it's there's only known case it, perhaps less close this PR and the original issue for now until there's a better reproduction.

PR #1136, opened 2026-07-21
Store paddle transaction_id in sessionStorage for reuse

### LINKED ISSUE #805: When doing paddle checkout, reload a draft transaction when possible
## Describe the bug

We had a case where a Subscriber was charged twice (already refunded the second charge) during the checkout process in Paddle. Unclear still if this is a problem within Paddle checkout or within our codebase as a race condition might happen and can be the cause of this.

## To Reproduce

Not a clear path of execution, unfortunately but we assume that:

1. Subscriber went through Paddle checkout
2. Before Paddle sent us any event or webhook the Subscriber refreshed the page
3. A new transaction_id was created and an empty checkout form was presented
4. Subscriber went through the Paddle checkout again

## Expected behavior

We should store the initial transaction_id somewhere (e.g. local storage?) and open the paddle checkout with that transaction_id so that in case there is a race condition, we don't show a brand new checkout form upon page reload.

davinotdavid:
## What changed?

- Store paddle `transaction_id` in sessionStorage when loading the checkout page and removing it once checkout has been completed or if there's an error state.

## Why?

- Not re-using the `transaction_id` generates all sorts of problems such as: double payments creating double subscriptions, reloading the checkout page without paying creating duplicated draft transactions, etc.

## Limitations and Notes

- It is still possible to reproduce seeing the checkout page if you reload the page exactly after clicking the payment button. However, filling the checkout form again doesn't cause double payments!

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/805

## Screenshots

https://github.com/user-attachments/assets/6c0d70f5-637f-4398-8508-f0230ac946d7

### DIFF
--- a/assets/app/vue/defines.ts
+++ b/assets/app/vue/defines.ts
@@ -15,3 +15,4 @@ export const COMMUNITY_SUPPORT_URL = 'https://support.mozilla.org/questions/thun
 export const NOT_INTERESTED_SURVEY_LINK = 'https://www.surveymonkey.com/r/HYYMDB6';
 export const TBPRO_WAIT_LIST = 'https://tb.pro/waitlist/';
 export const CAPTURE_TELEMETRY = true;
+export const PADDLE_TRANSACTION_STORAGE_KEY = 'accounts/subscribe/paddleTransactionId';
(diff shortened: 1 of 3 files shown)


MelissaAutumn (changes requested):
For some reason this doesn't work on my end. But if I flip it to localStorage it seems to work fine. I think that's probably fine with an additional check on the dashboard to remove any value from `accounts/subscribe/paddleTransactionId` on load.

davinotdavid:
@MelissaAutumn Hmm this is weird 🤔 I am not sure why `sessionStorage` is failing there but it definitely worked here. I've changed to `localStorage` though (and moved the key into a const), thanks!

MelissaAutumn (approved):

PR #1146, opened 2026-07-22
Add basic accounts sign-in test for new k8s env

### LINKED ISSUE #1137: Add basic accounts sign-in test for K8s env
As part of the new K8s/argo/kargo infrastructure work we need a very basic E2E test that will sign into TB Accounts:

- For now this test needs to be able to run against a TB Accounts URL / environment regardless on if the other services (thundermail/appt/send) are integrated into the same test environment, just to prove Accounts sign-in works and the dashboard appears
- Add a new playwright test tag (`deployment-analysis`) and an associated npm command so that we can choose to run only this test (and it won't be included in our other existing environment CI runs/GHA jobs)
- No need to run this in BrowserStack as it will just run on desktop Firefox for now, and it may possibly run on a local URL specific to the test environment (and not accessible by BrowserStack)
- The TB Accounts hub URL and sign-in credentials will be specified via env vars in the local `test/e2e/.env` file and secrets (like the other tests currently do)
- This new test will be triggered by kargo and probably run inside a playwright container (not in this PR)

cc @aatchison

rwood-moz:
## What changed?
Adding a basic sign-in test that will be used to verify TB Accounts sign-in on the new K8s environment.

## Why?
Add a basic Accounts sign-in E2E test that will be triggered by kargo when verifying the health of kargo freight (i.e. TB Accounts deployment artifacts). The test is added with a new playwright tag `deployment-analysis`. In the future we can expand the suite by tagging more E2E tests; but this will get the process working.

## Limitations and Notes
This test is built to run on desktop browsers only; there is no BrowserStack support as that is not necessary (see #1137) for more info.

Prerequisite: The test requires and existing TB Accounts / Thundermail test user on the target environment (credentials read from a local test/e2e/.env).

To run the test:

```
cd test/e2e
cp .env.stage.example .env
```

Then edit your tests/e2e/.env and set the Accounts URL and existing test user credentials. Then:

``` 
npm ci
npx playwright install
npm run deployment-analysis-e2e
```
Or to watch the test run:

```
npm run deployment-analysis-e2e-headed
```

## Applicable Issues
Fixes #1137.

## QA Log
Ran the new test on my local machine against Firefox desktop, in headless and non-headless mode and the test passes fine.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -148,6 +148,18 @@ To run the E2E tests in debug mode (still in `test/e2e`):
 npm run e2e-test-debug
 ```
 
+To run the deployment-analysis E2E tests, follow the above setup instructions and then run:
+
+```bash
+npm run deployment-analysis-e2e
+```
+
+Or to watch the tests run:
+
+```bash
+npm run deployment-analysis-e2e-headed
+```
+
 ## Running on BrowserStack
 
 You can run the E2E tests from your local machine pointed at the stage environment, but against browsers provided in the BrowserStack Automate cloud.
--- a/test/e2e/const/constants.ts
+++ b/test/e2e/const/constants.ts
@@ -1,5 +1,5 @@
 // environment where the tests will run
-export const ACCTS_TARGET_ENV = String(process.env.ACCTS_TARGET_ENV);
+export const ACCTS_TARGET_ENV = String(process.env.ACCTS_TARGET_ENV ?? 'dev');
 
 // tb accounts urls
 export const ACCTS_HUB_URL = String(process.env.ACCTS_HUB_URL);
--- a/test/e2e/const/constants.ts
+++ b/test/e2e/const/constants.ts
@@ -33,6 +33,7 @@ export const PLAYWRIGHT_TAG_E2E_SUITE = '@e2e-suite';
 export const PLAYWRIGHT_TAG_E2E_PROD_DESKTOP_NIGHTLY = '@e2e-prod-desktop-nightly';
 export const PLAYWRIGHT_TAG_E2E_SUITE_MOBILE = '@e2e-mobile-suite';
 export const PLAYWRIGHT_TAG_E2E_PROD_MOBILE_NIGHTLY = '@e2e-prod-mobile-nightly';
+export const PLAYWRIGHT_TAG_DEPLOYMENT_ANALYSIS = '@deployment-analysis';
 
 // timeouts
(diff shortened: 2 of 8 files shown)


aatchison:
Nice, focused addition @rwood-moz — the `@deployment-analysis` tag + smoketest split reads well. Reviewed it against the harness it depends on; a couple of things worth tightening before it's wired into freight verification. Context: we're writing the Kargo `AnalysisTemplate` in `accounts-deploy` that will drive this, so we'll own setting the env var in item 1.

**Higher-value:**

1. **The skip guard hinges on an env value nothing here sets, and a mismatch fails hard instead of skipping.** `test.skip(ACCTS_TARGET_ENV == 'dev', …)` is the only thing keeping this off a subscription-less user, and `ACCTS_TARGET_ENV = String(process.env.ACCTS_TARGET_ENV)` is the literal `"undefined"` when unset. So unless the runner exports exactly `ACCTS_TARGET_ENV=dev`, the test runs, `navigateToDashboard()` polls for `/dashboard`, the no-subscription user is redirected to `/subscription`, and it fails rather than skips. Our AnalysisTemplate will set this explicitly, but it'd be more robust to fail-safe (skip unless on a known-subscribed env) so a missing var can't turn into a red freight check.

2. **Docstring vs. actual sign-in path.** The spec says sign-in "is performed automatically (via `auth.desktop.setup.ts`) … as a desktop-setup dependency", but `beforeEach` also calls `ensureWeAreSignedIn()`, which re-signs-in and re-saves `storageState`. Under `--grep deployment-analysis` the untagged setup test (`'desktop browser authenticate'`) is filtered out, so `beforeEach` is actually what authenticates — meaning either the docstring is stale, or (if setup does run) you sign in twice. Suggest picking one path and updating the comment to match.

**Polish:**
- `verifyDashboardDisplayed(smoketest = false)` — a positional boolean flag; a named method (`verifyDashboardBasics()`) or options object would read cleaner at call sites. Note the smoketest path still requires a *subscribed* user (`navigateToDashboard` hard-asserts `/dashboard`); it only relaxes the env-specific price/storage assertions — maybe a one-line comment so that's clear.
- Typos surfaced in the report: test title `succssful` → `successful`; docstring `perfomed` → `performed`.
- Trailing whitespace on the blank docstring line; `==` vs `===` on the env check.

Approach looks right — no concerns with the E2E strategy itself, and desktop-firefox-only/no-BrowserStack is a sensible scope. 🙌

aatchison (approved):

PR #1148, opened 2026-07-23
Improve dev ergonomics for running tests and using sudo

markstos:
Before, `./run-tests.sh` ran zero-tests by default, expected /bin/bash
to exist and unnecessarily used `sudo` to call Docker. This changes
are backwards compatible for passing args: that still runs tests as before.

`sudo` was removed from other calls to docker as well. Under normally configured
Mac and Linux configurations, the `docker` command is a client command that permission
to run communicate with a docker daemon that's running as root. It's not necessary
to call the client with `sudo`.

This change simplifies the syntax for devs and removes a password prompt.
It's not expected to change security much, as the docker daemon normally
runs as root with or without sudo.

## QA Log

 - Before/after manual testing of ./run-tests.sh
 - Reserching `sudo docker` vs `docker` impact
 - Testing `sudo docker` vs `docker` on Linux

 - Manually testing on Mac appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ Make sure that the containers are already running.
 To run all tests:
 
 ```shell
-./run-tests.sh thunderbird_accounts
+./run-tests.sh
 ```
 
 this will execute the following command for you:
--- a/boot-docker-build.sh
+++ b/boot-docker-build.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-sudo docker compose up --build accounts vite-dev celery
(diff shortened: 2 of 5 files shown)


MelissaAutumn (approved):
I mostly use run-tests with a subset of tests `./run-tests.sh thunderbird_accounts.subscription.tests.test_commands` so allowing it to default to just thunderbird_accounts didn't occur to me. Sounds good to me there.

Docker changes are okay by me. 

Thanks!

PR #1150, opened 2026-07-23
Correctly serve template variables

### LINKED ISSUE #1135: Fix favicon.ico
We should probably create a django route to serve favicon.ico so the admin panel has an icon too.

### LINKED ISSUE #1149: Accounts incorrectly encodes/decodes injected variables
Big whoops on my end. We're not shoving our variables onto the main template correctly. Fix will be up shortly.

There's no security concerns here, just a trick to get your page to reload infinitely.

MelissaAutumn:
Fixes #1149 
Fixes #1135 

We now let Django handle all of our escaping needs, and then just grab the blob and decode it. From my testing it works fine, I had to adjust the tos tests to read in the blob. 

This fixes some encoding bugs that can lead to a infinite redirect. And also I fixed favicon for admin panel.

### DIFF
--- a/src/thunderbird_accounts/core/views.py
+++ b/src/thunderbird_accounts/core/views.py
@@ -13,6 +13,7 @@ from django.contrib import messages
 from django.http import HttpRequest, HttpResponseRedirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
+from django.utils.html import json_script
 from django.utils.translation import gettext_lazy as _
 from django.contrib.messages import get_messages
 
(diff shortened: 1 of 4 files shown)


src/thunderbird_accounts/core/views.py
```diff
@@ -141,32 +142,48 @@ def home(request: HttpRequest):
         # Clear form_data for any additional reloads
         request.session['form_data'] = {}
 
+    # Json script will create a json blob that's correctly escaped
+    script_tag = json_script(
+        {
+            'connectionInfo': settings.CONNECTION_INFO,
+            'appPasswords': app_passwords,
+            'userDisplayName': user_display_name,
```
davinotdavid: Nit but could we either order this alphabetically or move `userDisplayName` near the bottom around the `username`, `userEmail` etc?

davinotdavid (approved):
Nice! LGTM while testing locally and doing a code diff.

We can also remove the `import json` from `src/thunderbird_accounts/core/views.py` now that its unused!

src/thunderbird_accounts/core/views.py
MelissaAutumn: Oh thanks, will do!

PR #1153, opened 2026-07-27
Add dinero.js to detect and correctly display currency. (Fixes #1152)

### LINKED ISSUE #1152: Yen is displayed incorrectly
We're always assuming 100 cents = 1 dollar.

MelissaAutumn:
Fixes #1152 

[Dinero.js](https://www.dinerojs.com/) seems to be a choice for this. I'm sort of shocked there's no native js (or python) library function for this but oh well. Paddle doesn't provide the monthly price because we're billing annually. It'd be better if we just gave up on displaying pricing in monthly, it's kinda weird tbh. 

Technically support's "total spend" value will still be incorrect for non 100 cents = 1 dollar currencies. We should just pull in a Paddle price preview but we can cut another ticket for that later.

### DIFF
--- a/assets/app/vue/views/DashboardView/utils.ts
+++ b/assets/app/vue/views/DashboardView/utils.ts
@@ -1,7 +1,8 @@
 import type { ComposerNumberFormatting, ComposerTranslation, ComposerDateTimeFormatting } from 'vue-i18n';
 import type { SubscriptionData } from './types';
 import { i18n } from '@/composables/i18n';
-
+import { dinero, add, toDecimal, toSnapshot, allocate } from 'dinero.js';
+import * as currencies from 'dinero.js/currencies';
 /**
  * Converts bytes to the most appropriate unit (KB, MB, or GB) with the unit label
  * @param bytes - The number of bytes to convert
@@ -34,10 +35,40 @@ export const formatBytes = (bytes: string | null): string | null => {
 }
 
 /**
+ * Takes in the amount in cents (raw data from Paddle), currency code and billingInterval and outputs the correctly formatted price in monthly interval.
+ * Ex/
+ *  * (12_000, CAD, year) = CA$10
+ *  * (12_000, USD, year) = $10
+ *  * (11_700, JPY, year) = ¥975
+ */
+export const formatLocalizedMonthlyPrice = (
+  amountCents: number,
+  currency: string,
+  billingInterval: string | undefined
+): string => {
+  // Use currency if supported, otherwise fallback to USD.
+  const currencyType = currency in currencies ? currencies[currency as keyof typeof currencies] : currencies['USD'];
+  const annualBilling = billingInterval?.toLowerCase() === 'year';
+
+  const priceObj = dinero({ amount: amountCents, currency: currencyType });
+
+  // allocate expects a list of equal amounts, so in order to split this by 12 correctly we must fill a list 12 times...bleh  
+  const allocation = Array(12).fill(0, 0, 12).map(() => (1.0 / 12.0) * 100.0);
+  const price = annualBilling ? toDecimal(allocate(priceObj, allocation)[0]) : toDecimal(priceObj);
+
+  return new Intl.NumberFormat(i18n.locale.value, {
+    style: 'currency',
+    currency,
+    // @ts-ignore: lsp is complaining this property doesn't exist, so ignore that "error".
+    trailingZeroDisplay: 'stripIfInteger' 
+  }).format(parseFloat(price));
+};
+
+/**
  * Formats subscription data from the backend into a display-friendly format
  * @param subscriptionData - The subscription data from the backend
- * @param n - The number formatting function from vue-i18n
- * @param t - The translation function from vue-i18n
+ * @param n - The number formatting function from vue-i18n (not used)
+ * @param t - The translation function from vue-i18n (not used)
  * @param d - The date formatting function from vue-i18n
  * - Converts currency code to symbol
  * - Converts price from cents to dollars
@@ -47,30 +78,11 @@ export const formatBytes = (bytes: string | null): string | null => {
  */
 export const formatSubscriptionData = (
   subscriptionData: SubscriptionData,
-  n: ComposerNumberFormatting,
-  t: ComposerTranslation,
+  _n: ComposerNumberFormatting,
+  _t: ComposerTranslation,
   d: ComposerDateTimeFormatting
 ): SubscriptionData => {
-  // Convert price from cents to dollars
-  let priceInDollars = parseFloat(subscriptionData.price) / 100;
-
-  // Convert yearly pricing to monthly equivalent
-  const period = subscriptionData.period.toLowerCase();
-  if (period === 'year') {
-    priceInDollars = priceInDollars / 12;
-  }
-
-  const fractionDigits = Number.isInteger(priceInDollars) ? 0 : 2;
-  
-  // n is not doing a good job in letting us specify the actual currency...so use javascript
-  const intlN = new Intl.NumberFormat(i18n.locale.value, {
-    style: "currency",
-    currency: subscriptionData.currency,
-    maximumFractionDigits: fractionDigits,
-    minimumFractionDigits: fractionDigits
-  });
-
-  const formattedPrice = intlN.format(priceInDollars);
+  const formattedPrice = formatLocalizedMonthlyPrice(parseInt(subscriptionData.price, 10), subscriptionData.currency, subscriptionData.period);
 
   // Format autoRenewal date if it exists
   const formattedAutoRenewal = subscriptionData.autoRenewal 
--- a/assets/app/vue/views/DashboardView/utils.ts
+++ b/assets/app/vue/views/DashboardView/utils.ts
@@ -1,7 +1,7 @@
 import type { ComposerNumberFormatting, ComposerTranslation, ComposerDateTimeFormatting } from 'vue-i18n';
 import type { SubscriptionData } from './types';
 import { i18n } from '@/composables/i18n';
-import { dinero, add, toDecimal, toSnapshot, allocate } from 'dinero.js';
+import { dinero, toDecimal, allocate } from 'dinero.js';
 import * as currencies from 'dinero.js/currencies';
 /**
  * Converts bytes to the most appropriate unit (KB, MB, or GB) with the unit label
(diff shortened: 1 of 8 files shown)


src/thunderbird_accounts/subscription/admin.py
```diff
@@ -165,7 +165,7 @@ def price_name(self, obj):
             return obj.price.name
         return None
 
-    @admin.display(description=_('Price Amount'), ordering='price__amount')
+    @admin.display(description=_('Price Amount (in USD)'), ordering='price__amount')
```
markstos, line 168: Could you clarify why we are hardcoding USD here while we support other currencies elsewhere? Because this is our canonical pricing an currency?

assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue
```diff
@@ -70,12 +49,13 @@ const loadPlanPrice = async () => {
         quantity: 1,
         priceId,
       })),
+      currencyCode: 'JPY',
```
markstos: Accidental hard-coding?

assets/app/vue/views/DashboardView/utils.ts
```diff
@@ -33,11 +34,41 @@ export const formatBytes = (bytes: string | null): string | null => {
...
+ *  * (12_000, USD, year) = $10
+ *  * (11_700, JPY, year) = ¥975
+ */
+export const formatLocalizedMonthlyPrice = (
+  amountCents: number,
+  currency: string,
+  billingInterval: string | undefined
+): string => {
+  // Use currency if supported, otherwise fallback to USD.
+  const currencyType = currency in currencies ? currencies[currency as keyof typeof currencies] : currencies['USD'];
```
markstos, line 50: Is the fallback to USD safe? If the currency is unsupported, formatting with the original currency code but calculating with USD metadata can silently produce wrong math.

If a currency is unsupported, how we are handling it at all?

assets/app/vue/views/DashboardView/utils.ts
```diff
@@ -33,11 +34,41 @@ export const formatBytes = (bytes: string | null): string | null => {
...
 }
 
+/**
+ * Takes in the amount in cents (raw data from Paddle), currency code and billingInterval and outputs the correctly formatted price in monthly interval.
+ * Ex/
+ *  * (12_000, CAD, year) = CA$10
+ *  * (12_000, USD, year) = $10
+ *  * (11_700, JPY, year) = ¥975
+ */
+export const formatLocalizedMonthlyPrice = (
```
markstos, line 44: `utils` as a namespace provides little clue as to what's inside, as almost anything can end up shoved in a `utils` namespace. It looks like the everything that's in the file now follows the same theme: formatting subscriptions, formatting prices, formatting bytes. How about renaming `utils.ts` to `formatters.ts` as part of adding these additional formatters to it?

markstos (changes requested):
Thanks for this work.

I added some feedback and questions. Also, add some "QA Log" notes to share your manual testing of the changes and address linter issues with unused imports.

Thanks!

```
❯ corepack npm exec -- eslint assets/app/vue/views/DashboardView/utils.ts assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue\n\n/home/mark/Documents/Mozilla/git/thunderbird-accounts-1152-yen-display-bad/assets/app/vue/views/DashboardView/utils.ts\n  4:18  warning  'add' is defined but never used. Allowed unused vars must match /^_/u         @typescript-eslint/no-unused-vars\n  4:34  warning  'toSnapshot' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars\n\n/home/mark/Documents/Mozilla/git/thunderbird-accounts-1152-yen-display-bad/assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue\n  11:10  warning  'dinero' is defined but never used. Allowed unused vars must match /^_/u      @typescript-eslint/no-unused-vars\n  11:18  warning  'add' is defined but never used. Allowed unused vars must match /^_/u         @typescript-eslint/no-unused-vars\n  11:23  warning  'toDecimal' is defined but never used. Allowed unused vars must match /^_/u   @typescript-eslint/no-unused-vars\n  11:34  warning  'toSnapshot' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars\n\n✖ 6 problems (0 errors, 6 warnings)

```

src/thunderbird_accounts/subscription/admin.py
MelissaAutumn: Yes!

assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue
MelissaAutumn: I blame migraine, thanks for the spot!

assets/app/vue/views/DashboardView/utils.ts
MelissaAutumn: USD is canonical pricing, so I think it's okay. The worst thing will happen is that they'll see USD on confirm plan, then their probably obscure currency on checkout (as we don't use dinero on checkout.)

assets/app/vue/views/DashboardView/utils.ts
MelissaAutumn: Sounds like a good follow-up issue!

assets/app/vue/views/DashboardView/utils.ts
```diff
@@ -33,11 +34,41 @@ export const formatBytes = (bytes: string | null): string | null => {
...
+  }).format(parseFloat(price));
+};
+
 /**
  * Formats subscription data from the backend into a display-friendly format
  * @param subscriptionData - The subscription data from the backend
- * @param n - The number formatting function from vue-i18n
- * @param t - The translation function from vue-i18n
+ * @param n - The number formatting function from vue-i18n (not used)
+ * @param t - The translation function from vue-i18n (not used)
```
davinotdavid, line 71: Nit as it is not part of the functional aspect of this PR but if this is not used, we could remove it.

`formatSubscriptionData` is apparently being called in one place only https://github.com/thunderbird/thunderbird-accounts/blob/bugs/1152-yen-display-bad/assets/app/vue/views/DashboardView/components/YourCurrentSubscription.vue#L32 so it is a clean cleanup

package.json
```diff
@@ -30,6 +30,7 @@
     "@phosphor-icons/vue": "^2.2.1",
     "@thunderbirdops/services-ui": "^1.6.7",
     "@vueuse/core": "^14.2.1",
+    "dinero.js": "^2.0.2",
```
davinotdavid, line 33: Reminder to self that we should probably pin versions of packages here

src/thunderbird_accounts/support/customer.py
```diff
@@ -294,6 +294,7 @@ def _get_total_spend(user: User) -> list[dict]:
     for transaction in transactions:
         totals[transaction.currency] += int(transaction.total)
 
+    # FIXME: Support is getting the incorrect value, we should just pull directly from Paddle instead.
```
davinotdavid, line 297: I couldn't find a bug ticket for this so opened one here for tracking:

https://github.com/thunderbird/thunderbird-accounts/issues/1170

assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue
```diff
@@ -109,30 +87,20 @@ export default {
...
     <template v-slot:form-elements>
       <div class="plan-card-container">
-        <plan-card
-          :name="planName"
-          :description="planDescription"
-          :price="priceDisplay"
-          :price-loading="priceLoading"
-        />
+        <plan-card :name="planName" :description="planDescription" :price="priceDisplay"
+          :price-loading="priceLoading" />
```
davinotdavid, line 103: We should all use the same formatter so that we don't have these kinds of diffs again IMO.

The previous code was being formatted with Prettier (which I think reads better and we do have a [prettierrc](https://github.com/thunderbird/thunderbird-accounts/blob/main/.prettierrc) but happy to change) and the new one seems to be using the Vue (Official)'s own formatter 🤔

davinotdavid (approved):
Tested locally with various currencies and LGTM. Thanks!

Some comments but nothing blocking nor required to be done in this PR

src/thunderbird_accounts/support/customer.py
MelissaAutumn: Thanks, slipped my mind.

assets/app/vue/views/SignUpView/views/StepConfirmPlan/index.vue
MelissaAutumn: Ah I missed the prettier addition...which was probably a while ago. Will add that to my sublime config, thanks for the pointer!

package.json
MelissaAutumn: Pin versions, is that not package-lock.json? Or did I miss another convo?

markstos (approved):
My feedback has been addressed.

package.json
davinotdavid: yes! it doesn't affect our workflows at the moment but it would just be nice to have the versions here be in sync with package-lock.json by removing the caret

MelissaAutumn: When did we have this discussion btw? Just wanted to re-read and follow along.

davinotdavid: I honestly don't remember 😅 perhaps we can revisit this in the future when needed then

markstos: To Mel's point, if we had a standard on this, it would be good to reflect in a policy/procedure document and have automated enforcement if we could.

As to the potential benefit: It's true the lock file applies either way. 

Exact pinning is safer for those edge cases where we do need to recreate the lock file or it gets modified more than we intend. Or to frame it another way:

- If we were publishing a module that other modules would pull in, then using caret versioning provides downstream consumers flexibility about what version to pull in, so it's best for that case.

- But as a /top-level application/ -- as we are-- that's not expected to be pulled in a dependency, pinning exact versions is a more precise declaration of our intent and what we have tested with.

My last co adopted exact pinning and it went smoothly. `npm`  has tooling to help with it with `npm install --save-exact`

So if did have a conversation about this, I'd be in favor of moving exact pinning for this project.

MelissaAutumn: Someone wanna cut a ticket for this discussion? I'm worried that we'll miss patch releases that are automatic and free when we just do npm i && uv lock.

markstos: On it.

markstos: https://github.com/thunderbird/thunderbird-accounts/issues/1182

PR #1156, opened 2026-07-28
Prod scaling and task management enablement

ryanjjung:
Today I am responding to an alert about Accounts Celery in prod using too much memory. This appears to be a handful of runaway tasks, but I am unable to terminate those tasks since we do not have `FLOWER_UNAUTHENTICATED_API` enabled in prod. This is safe to enable because there is no public access to the Flower server; this requires being able to create a bastion host in our env and log into it.

I have also adjusted our scaling parameters to allow this to scale up farther if this sort of thing happens again. I've also updated the images for the prod environment.

### DIFF
--- a/pulumi/config.prod.yaml
+++ b/pulumi/config.prod.yaml
@@ -291,7 +291,7 @@ resources:
               protocol: tcp
               from_port: 0
               to_port: 65535
-      
+
   tb:secrets:PulumiSecretsManager:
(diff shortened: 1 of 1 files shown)


markstos:
Thanks for this. These runaway tasks you noticed-- did you noticing anything suggesting we should make some code-side updates to keep them from running away in the future?

Issue #1147, opened 2026-07-23
Add Infra to support making MaxMind GeoIP database files to `thunderbird-accounts` container


markstos:
To support showing the geo-location of login sessions, we need to add a bit of infra that maps IP addresses back to physical locations. 

A solution for this that's being used elsewhere in Mozilla is the MaxMind GeoIP database. I've inquired internally to make sure the licensing details are worked out.

There's (private) documentation about how to access Mozilla-licensed files internally:
https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/708739197/Shared+MaxMind+GeoIP+database+files

## Infra needs

We need whatever GeoIP file available in the accounts container. So there's a need for the original file to be bootstrapped and then the file should be refreshed at least once a month to get the latest values.

The simplest solution seems to be to download a fresh copy of the file at image-build time. Ideally, we could architect this in such a way that other users of this code could substitute their own file/URL or opt-out of this feature. 

I *believe* the file we'll need is the GeoIP-City file, but we can order this work after some version of is tested locally to be sure. 

## About Open Source

IP-to-location isn't a feature that there's a good fully open source alternative too. For those who wish to use this software, there's a free/open-source "GeoLite2" data source that can be used instead or users can purchase their own license. We'll also look at having the code fallback to "Unknown Location" rather than having a hard failure if this dependency is missing.

markstos:
For licensing and cost reasons, we are pivoting to looking at free DB IP Lite for this vendor for now, though the automation pattern should be approximately the same for either this vendor or MaxMind.

https://db-ip.com/db/lite.php 

The database is about 10 million rows, available in MMDB or CSV format. 

One architecture idea to manage this is a monthly Celery task to dump the latest CSV into a PostgreSQL table, with a staging table that gets swapped to become live.

Another option is to download the MMDB format, possibly caching recent versions in our own object storage. K8s pods would copy this into an `emptyDir` volume. Django would mount and query the MMDB file from there. This requires a new Python dep to query the MMDB format.

Using MMDB will mean faster queries and less storage, but perhaps not in ways that matter. PostgreSQL allows us to re-use standard patterns.

markstos:
After speaking with @mzeier we are going to try the PostgreSQL option first for architectural simplicity. Later if the performance difference of the MMDB format matters, we can pivot to that. 

I'll work on a PR for the automation for downloading and updating the file.

MelissaAutumn:
[...]

Issue #210, opened 2025-07-21
Create a "Security settings" page


MelissaAutumn:
## Details

Create a security settings page that lists account activity (via keycloak), and allows the user to manage allow list (via stalwart)

## Zeplin

https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/685f1e1e0c7f6f9199fe2802

## Screenshots

<img width="1280" height="1439" alt="Image" src="https://github.com/user-attachments/assets/4977e90f-aeb5-4ed7-857d-f603336b72db" />
<img width="693" height="481" alt="Image" src="https://github.com/user-attachments/assets/50286809-35e4-4f26-9cae-920821141b9b" />

malini:
A more up-to-date design is available [here](https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/6905157d74a53119cf97bb33)

malini:
removing Milestone tag to stop fighting thundernotion. We'll track this as a normal task, feel free to add sub issues as needed though

MelissaAutumn:
We should really disable that functionality.

kewisch:
#321 is itself a milestone, which is meant for small tasks that aren't big enough for a milestone. Tasks have Milestone parents, Milestones have Epic parents, Epics have Program parents.

If there is an expectation this issue will have sub-issues, then it must remain a Milestone. There is a notion page that is already linked to this issue, so that will be the authoritative source, and is why thundernotion is fighting you.

If this is a Milestone and should lead up to Tailwind, then it needs to lead up to the Tailwind Epic. We didn't have that on GitHub yet because the Epics-on-GitHub was just recently introduced.

> We should really disable that functionality.

I added these checks because before we had a wild net of different sub-issue types, which just don't map well with Notion and make project management more complicated (and was breaking the sync). If there is any more information I can add to the comments to make more clear what needs to be done, please let me know.

markstos:
I've updated the issue with the most current link to Zeplin. The screenshots in the issue are outdated. See Zeplin for the most current design plans.

markstos:
Work on this by @davinotdavid  has started at: https://github.com/thunderbird/thunderbird-accounts/pull/378 Here's the latest status after sync'ing with him:

- I'm going to look into how to store and retrieve the session device information. 
- I'll loop Davi back into help integrate the latest design from Zeplin.

PR #1162, opened 2026-07-31
Add "Current Sign-ins" feature

markstos:
*The  🎨 emoji notes some optional design feedback.*

## What changed?

This adds a new feature for users to list active keycloak sessions and sign out of individually
or all at once.

## Why?

 * Requested by issue #210 to show show active login sessions

## Context

Both this Django "accounts" app as well as Stalwart use Keycloak to manage authentication and user sessions. So, IMAP and JMAP sessions go through Keycloak, including our webmail.

Each session has two sides: The server Keycloak side, and Keycloak clients.
Because [we haven't implemented backchannel logouts)](https://github.com/thunderbird/thunderbird-accounts/issue/1158), this work does /not/ instantly cause clients to get logged out. They will get logged out when their *local* session expires or when they check in with Keycloak again.
Analysis found that's 6 minutes or less. For other "thunderbird-accounts" sessions, it 
could be up to 15 minutes, but that should close back to about 5 after [#1054 Update oidc middleware with two new feature flags that introspect and allow posting](https://github.com/thunderbird/thunderbird-accounts/pull/1054) lands. be prepared to terminate the local session immediately.

🎨 Perhaps after users click "sign out", the confirmation should mention mention this potential lag.

## Scope Notes

There several different kinds of authentication details that users would like to see and manage.

- Out of scope is a  "Login Log", a report of authentication events. (These are not sessions that can be logged out of)

- 🎨  We should perhaps link to the "change app password" feature or put on the same page.
- 

## Design Notes

[There are related designs in Zeplin](https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/6905157d74a53119cf97bb33). It appears the app evolved some since the design was made, as there is
now a "security settings" page, so the design was implemented there. Some other deviations from the spec'ed design:

- The spec included locations as "$City, $Country", but only globally recognized cities were used in the examples, like "Toyko". That doesn't work for cities like Richmond-- there more than 50 places named Richmond in the US. So the state was added as well.

- 🎨 The spec did not address the case of the current session. What was actually implemented displays "This is you" for the current session and does NOT provide a sign-out link for this table entry--  we already have a standard "Log out" menu item for that.

- 🎨 We are required to provide attribution for the data source we used. A link was added for that.

- 🎨 The spec did not address exactly what happens when Sign-out is clicked or what happens if  sign out fails.

## Handling geo-ip to location mapping

In [#1147](https://github.com/thunderbird/thunderbird-accounts/issues/1147) we explored a number of ways to addresss the need to map IPs to locatio names. The final solution implemented here after consider some others is to pull in a fresh copy of an CC-licensed MMDB-format database into our accounts container each time we build one. Because such DBs are freely available, this is a
hard depenency. While I believe we already already test by running containers, I'll note that the path the file is loaded from is relative to the container so this module *must* be used within a
container.

## Applicable Issues

 * #210 
 * #1147

## Screenshots

Happy Path view: _(This includes work in the stacked Connected Apps PR):_

image

_This screenshot is outdated because it includes the IP addresss column, but shows how locations look when there is one detected._

image

Confirmation (🎨 Should probably re-worked into some other modal design):  *

image
image

Example error display (Should be rare).  (These designs are out of date in that they still show an "IP Address" column, but the key parts: the modal and the error display are what's key here)*

image


kewisch:
> We have no "Connected Apps" management. Unlike live sessions, these apps have log-lived OAuth tokens that can't be managed the same way. Thunderbird Desktop would be an example.

Is there followup work planned on this? I expect users will want to also disconnect their Thunderbirds etc.

markstos:
> Is there followup work planned on this? 

Competitors offer the ability to view and disconnect "Connected Apps" with long-lived access tokens and I recommend that we have it on our roadmap as well. CC: @radishmouse

kewisch:
#210 is not scoped out with our current template, but to me this is in scope for the milestone. The ultimate goal is that users are in control of their sessions and can end them if they want to clean up or suspect a device has been stolen.

markstos:
@laurelterlesky I'd like design input on the issues highlighted with 🚧 here.

laurelterlesky:
Yes, let's alert users of the 'lag' time / expected time after sign out. This will help set expectations.

Would it be useful for the design team to update the [active sessions design](https://zpl.io/KE4yZv7) to address the second item you flagged @markstos ?

markstos:
@laurelterlesky Yes, some updated designs to address the flagged item, thanks!

markstos:
I met with @solangevalverde about related design refinements yesterday. 

@solangevalverde Could you provide a timeline update when we may get design refinements back?

kellymcs01:
The process flow looks good, the screen flow makes sense, and the page for managing the devices is intuitive. Overall LGTM.

markstos:
I've merged the "Connected Apps" PR which expanded the scope into this PR and they are together ready for official review.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,126 @@
...
+const loading = ref(true);
+const errorMessage = ref<string | null>(null);
+
+const removeAccess = async (id: string) => {
+  const app = connectedApps.value.find((connectedApp) => connectedApp.id === id);
+  if (!app) {
+    return;
+  }
+
+  if (!window.confirm(t('views.mail.views.securitySettings.removeAccessConfirmation', { app: app.label }))) {
```
MelissaAutumn: We have a modal component that we should probably use. I wish it had a simple api like window.confirm though 😄 

https://thunderbird.github.io/services-ui/?path=/docs/components-modaldialog--docs

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,126 @@
...
+    errorMessage.value = t('views.mail.views.securitySettings.errorRemovingAccess');
+  }
+};
+
+onMounted(async () => {
+  try {
+    const data = await getConnectedApps();
+    const sortedData = [...data].sort((a, b) => (b.last_access || 0) - (a.last_access || 0));
+
+    connectedApps.value = sortedData.map((app: ConnectedApp, index) => ({
```
MelissaAutumn: Should this be a computed instead?

src/thunderbird_accounts/authentication/api.py
```diff
@@ -62,13 +67,78 @@ class RecoveryCodesRegenerateThrottle(UserRateThrottle):
...
+        return None
+
+    try:
+        return jwt.decode(access_token, options={'verify_signature': False}).get('sid')
+    except (jwt.PyJWTError, TypeError):
+        return None
+
+
+def mark_current_session(request: Request, sessions: list[dict]) -> list[dict]:
+    oidc_id_token = request.session.get('oidc_id_token')
```
MelissaAutumn: I want to say the id token actually doesn't get refreshed....that's a djgnao mozilla oidc quirk. sid should be on the access token though. Let's just skip the id token check.

src/thunderbird_accounts/authentication/api.py
```diff
@@ -327,3 +396,55 @@ def sign_up(request: Request):
...
+    # Sign out from the Keycloak session
+    keycloak_client = KeycloakAccountClient()
+    keycloak_client.sign_out_session(user_access_token, session_id)
+
+    oidc_id_token = request.session.get('oidc_id_token')
+    if oidc_id_token:
+        # Verify if the request's keycloak session_id matches the one in the ID token.
+        auth_backend = AccountsOIDCBackend()
+        payload = auth_backend.verify_token(oidc_id_token)
+        keycloak_session_id = payload.get('sid')
```
MelissaAutumn: Can we switch this to the current access token?

MelissaAutumn (changes requested):
Initial end of day review, will follow up tmw or fri. Posting this now so I don't accidentally leave pending comments for 15 days like the last time this happened 😅

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: In practical terms I don't think it makes it a difference, as locale was perhaps the only input that might change, but I've proceeded with the refactor on the principal of separating the API data and presentation data. The refactor also will cover the edge case of the user changing their locale mid-session, if that's even a thing.

solangevalverde:
UX flow was already communicated during past sync with Mark, here is the[ zeplin link](https://zpl.io/NEBElPk) with the hi-fi frame. lmk if there are any Qs :)

src/thunderbird_accounts/authentication/api.py
markstos: Good catch. I reviewed this area and confirmed we can simplify there.

src/thunderbird_accounts/authentication/api.py
markstos: Yes, I fixed this as part of fixing the above issue as well.

markstos:
@solangevalverde I've given the hi-frame another pass and brought the work into alignment with it. I left a few comments in Zeplin and details that appears that seemed worth adjusting to development.

assets/app/vue/views/MailView/views/SecuritySettingsView/index.vue
```diff
@@ -19,11 +21,14 @@ export default {
...
   <div class="security-settings-view">
+    <welcome-header />
+
     <card-container
       class="security-settings-card"
       :title="t('views.mail.views.securitySettings.securitySettings')"
     >
       <account-activity />
+      <connected-apps />
       <allow-list />
```
MelissaAutumn, line 32: Please disable this component, or gate it with a waffle feature flag. It's non-functional.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,166 @@
...
+      ),
+      accessGiven:
+        typeof app.access_given === 'number' && Number.isFinite(app.access_given)
+          ? formatDate(new Date(app.access_given), locale.value, t)
+          : t('views.mail.views.securitySettings.unknownAccessGiven'),
+      lastAccess:
+        typeof app.last_access === 'number' && Number.isFinite(app.last_access)
+          ? formatDate(new Date(app.last_access), locale.value, t)
+          : t('views.mail.views.securitySettings.unknownLastAccess'),
+    }))
```
MelissaAutumn, line 37: Why don't we do this formatting on the backend? It's the only consumer of this api endpoint afaik.

src/thunderbird_accounts/core/geoip.py
```diff
@@ -0,0 +1,50 @@
...
+        'state': (subdivision.get('names') or {}).get('en') or subdivision.get('name'),
+        'country_code': country.get('iso_code'),
+        'continent': continent.get('code'),
+    }
+
+
+def enrich_sessions_with_geoip(sessions: list[dict]) -> list[dict]:
+    locations_by_ip = {}
+    for session in sessions:
+        ip_address = session.get('ip_address')
```
MelissaAutumn, line 50: We have access to pydantic now in case you want to typehint better, but either way could you add a doc comment for this function?

src/thunderbird_accounts/core/geoip.py
```diff
@@ -0,0 +1,50 @@
...
+import maxminddb
+
+logger = logging.getLogger(__name__)
+
+# Expected to be provided by the container built by Dockerfile
+GEOIP_CITY_MMDB_PATH = Path('/app/data/dbip-city-lite.mmdb')
+
+
+def lookup_ip_location(ip_address: str) -> dict | None:
+    if not GEOIP_CITY_MMDB_PATH.exists():
```
MelissaAutumn, line 20: Doc comment pls

assets/app/vue/views/MailView/views/SecuritySettingsView/formatters.ts
```diff
@@ -0,0 +1,67 @@
...
+  yesterday.setDate(yesterday.getDate() - 1);
+
+  const dateToCheck = new Date(date.getFullYear(), date.getMonth(), date.getDate());
+
+  // Format time based on current locale
+  const timeString = date.toLocaleString(locale, {
+    hour: 'numeric',
+    minute: '2-digit',
+    hour12: true
+  });
```
MelissaAutumn, line 25: We're overriding locale with the hour12 😄. I don't remember if we hard-code 12 hour time elsewhere in the app, but if it's not could you remove it?

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,166 @@
...
+        :records="connectedApps"
+        :column-labels="{
+          primary: t('views.mail.views.securitySettings.recordsTableHeaderApp'),
+          location: t('views.mail.views.securitySettings.recordsTableHeaderLocation'),
+          accessGiven: t('views.mail.views.securitySettings.recordsTableHeaderAccessGiven'),
+          lastAccess: t('views.mail.views.securitySettings.recordsTableHeaderLastAccess'),
+          actions: t('views.mail.views.securitySettings.recordsTableHeaderActions'),
+        }"
+      >
+        <template #action="{ record }">
```
MelissaAutumn, line 115: I believe we're using v-slot:<name> everywhere else. We might want to discuss using alternative syntax before we start using it. Otherwise it might be confusing.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
```diff
@@ -1,80 +1,179 @@
...
+  id: string;
+  label: string;
+  ipAddress: string;
+  isCurrent: boolean;
+  location: string;
+  accessGiven: string;
+  lastAccess: string;
+};
+
+const isRenderableSession = (session: ActiveSession) =>
```
MelissaAutumn: If this is a point of concern we should filter this on the server-side.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
```diff
@@ -1,80 +1,179 @@
...
+    return;
+  }
+
+  signOutModal.value?.hide();
+
+  try {
+    await signOutSession(session.id);
+    activeSessions.value = activeSessions.value.filter((activeSession) => activeSession.id !== session.id);
+  } catch (error) {
+    console.log(error);
```
MelissaAutumn: Can we give some context to this stray log? Like console.log('sign out failed:', error); ?

assets/app/vue/views/DashboardView/components/PrivacyAndDataCard.vue
```diff
@@ -1,38 +1,61 @@
...
+import { PrimaryButton, VisualDivider } from '@thunderbirdops/services-ui';
 import CardContainer from '@/components/CardContainer.vue';
+import { signOutAllSessions } from '@/views/MailView/views/SecuritySettingsView/api';
 
 const { t } = useI18n();
+const router = useRouter();
+
+const goToActiveSessions = () => router.push('/mail/security-settings#current-sign-ins');
+const goToConnectedApps = () => router.push('/mail/security-settings#connected-apps');
+const goToContact = () => router.push('/contact');
```
MelissaAutumn: Instead of js functions can we use the router elements so that middle click and other accessibility functionality will be preserved? 

ex/

```html
        <router-link to="/mail">
          <img :src="logoSrc" alt="Thundermail" />
        </router-link>
```

I think it has a v-slot to customize the wrapper.

MelissaAutumn (changes requested):
Some more changes 😄.

Also sessions don't work with app password usage, that should probably be mentioned somewhere. Apologies for not raising it, we might want to look into fixing that soon.

Alternatively we could have a standing "App password" session, where log out will bring a warning that this will remove your app password and sign you out of all apps using it.

markstos:
> Also sessions don't work with app password usage, that should probably be mentioned somewhere. 

@MelissaAutumn My recommendation would to move App Password Management to the same page, which would clarify that it is related but separate, but so far updating App Password management has been out of scope here.

assets/app/vue/views/MailView/views/SecuritySettingsView/index.vue
markstos: Can you clarify non-functional? It's visible and tested to work.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: It's a browser-local formatting that factor's in the browser's time zone. So, we can't do that on the backend.

src/thunderbird_accounts/core/geoip.py
markstos: I've improved the types here and added the docstring.

assets/app/vue/views/MailView/views/SecuritySettingsView/formatters.ts
markstos: Good catch. Fixed.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: I get 72 matches for '<template #' and 12 for '<template v-slot'.  It appears to be that the shorthand notation is more popular. This would be helpful to clarify our joint standard here in a style guide.

MelissaAutumn:
> > Also sessions don't work with app password usage, that should probably be mentioned somewhere.
> 
> @MelissaAutumn My recommendation would to move App Password Management to the same page, which would clarify that it is related but separate, but so far updating App Password management has been out of scope here.

I'm not sure that's the best idea but a link to app passwords would help. There's some app password management work for I believe it's own page. I don't know if that's current work though. But agreed, out of scope.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
MelissaAutumn: Hashtags it is then!

assets/app/vue/views/MailView/views/SecuritySettingsView/index.vue
MelissaAutumn: Sorry I thought the line was enough, the `allow-list` component is not finished. Please hide it with a feature flag or a v-if="false".

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
markstos: Upon further review of Keycloak's data model, the checks here are not necessary. Addressed in: b46181f783107aa021f55dbe8041c3983c814a6e

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
markstos: Good suggestion. Implemented in cf55f0e06c17e61a11a3f628cc990d60a54ff82f

assets/app/vue/views/DashboardView/components/PrivacyAndDataCard.vue
markstos: Good call here to improve a11y. I figured out a way to combine the router-link syntax with the primary-button component in:  8315bf054309a83f5920ef976989bae5c2624626

markstos:
@laurelterlesky Something I mentioned in my PR write-up and @MelissaAutumn echo'ed in her review is that perhaps we should link back to "app password management" from the screen that now has Current Sign-ins and Connected Apps. Later, perhaps they should be on the same page, but there's a goal to control the scope here so we can get something out the door.

laurelterlesky:
OK. Sounds good. I'll be able to jump back into this work later today.

assets/app/vue/views/MailView/views/SecuritySettingsView/index.vue
MelissaAutumn: @markstos this is still pending I think.

src/thunderbird_accounts/authentication/api.py
```diff
@@ -62,13 +66,69 @@ class RecoveryCodesRegenerateThrottle(UserRateThrottle):
...
+    if not access_token:
+        return None
+
+    try:
+        return jwt.decode(access_token, options={'verify_signature': False}).get('sid')
+    except (jwt.PyJWTError, TypeError):
+        return None
+
+
+def mark_current_session(request: Request, sessions: list[dict]) -> list[dict]:
```
MelissaAutumn, line 80: no public non-api/view functions pls

src/thunderbird_accounts/authentication/api.py
```diff
@@ -62,13 +66,69 @@ class RecoveryCodesRegenerateThrottle(UserRateThrottle):
...
+    keycloak_client = KeycloakAccountClient()
+    sessions = keycloak_client.get_active_sessions(user_access_token)
+    sessions = mark_current_session(request, sessions)
+    return Response(enrich_sessions_with_geoip(sessions))
+
+
+@api_view(['GET'])
+@authentication_classes([SessionAuthentication])
+def get_connected_apps(request: Request):
+    if not request.user.is_authenticated:
```
MelissaAutumn, line 120: I think we can swpa these instances for the `@permission_classes([IsAuthenticated'])` like how waffle flags has it.

MelissaAutumn: https://www.django-rest-framework.org/api-guide/permissions/#isauthenticated

src/thunderbird_accounts/authentication/api.py
```diff
@@ -62,13 +66,69 @@ class RecoveryCodesRegenerateThrottle(UserRateThrottle):
...
+        raise NotAuthenticated()
+
+    user_access_token = get_user_access_token(request)
+    if not user_access_token:
+        raise NotAuthenticated('OIDC session has expired')
+
+    keycloak_client = KeycloakAccountClient()
+    sessions = keycloak_client.get_active_sessions(user_access_token)
+    sessions = mark_current_session(request, sessions)
+    return Response(enrich_sessions_with_geoip(sessions))
```
MelissaAutumn, line 114: If we're not displaying ip address should we should hide it from the api response imo. Less chance for ips to leak.

assets/app/vue/views/DashboardView/components/PrivacyAndDataCard.vue
```diff
@@ -1,38 +1,62 @@
...
-          <link-button>{{ t('views.dashboard.privacyAndDataCard.logoutFromAllDevicesButtonLabel') }}</link-button>
+        <div class="active-sessions-actions">
+          <router-link v-slot="{ href, navigate }" custom to="/mail/security-settings#current-sign-ins">
+            <primary-button :href="href" variant="outline" size="small" class="manage-active-sessions-button" @click="navigate">
+              {{ t('views.dashboard.privacyAndDataCard.manageActiveSessionsButtonLabel') }}
+            </primary-button>
+          </router-link>
+
+          <div class="link-text">
+            <a href="/logout/" @click.prevent="logoutFromAllSessions">
```
MelissaAutumn, line 29: I don't understand how this is supposed to work. You click it nothing happens, you open a new tab and you're logged out? 

This should probably be guarded with a modal imo but that can be a future ticket.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/SecurityAccessTable.vue
```diff
@@ -0,0 +1,136 @@
...
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="record in records" :key="record.id">
+          <td>{{ record.label }}</td>
+          <td>
+            <span :title="record.ipAddress">{{ record.location }}</span>
+          </td>
+          <td>{{ record.accessGiven }}</td>
+          <td>{{ record.lastAccess }}</td>
```
MelissaAutumn, line 53: nit: It'd be nice if the date time classes were using semantic element like https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time

assets/app/vue/views/MailView/views/SecuritySettingsView/components/SecurityAccessTable.vue
```diff
@@ -0,0 +1,136 @@
...
+          <th scope="col">{{ columnLabels.accessGiven }}</th>
+          <th scope="col">{{ columnLabels.lastAccess }}</th>
+          <th scope="col">{{ columnLabels.actions }}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="record in records" :key="record.id">
+          <td>{{ record.label }}</td>
+          <td>
+            <span :title="record.ipAddress">{{ record.location }}</span>
```
MelissaAutumn, line 50: Ah that's where the ip address is. Seems a little odd?

assets/app/vue/locales/en.json
```diff
@@ -304,14 +307,42 @@
...
-          "accountActivity": "Account Activity",
-          "accountActivityDescription": "Recent devices signed in to your account.",
+          "accountActivity": "Current sign-ins",
+          "accountActivityDescription": "Recent devices signed in to your account",
+          "loadingActiveSessions": "Loading active sessions...",
+          "errorLoadingActiveSessions": "We couldn’t load your active sessions. Please try again.",
+          "unknownDevice": "Unknown Device",
+          "unknownLocation": "Unknown location",
+          "signOutConfirmation": "Sign out of {device}?",
+          "signOutConfirmationDescription": "This will end the session on this device.",
```
MelissaAutumn, line 317: I think this is worded a little weird, personally it should be more like "This will end the session and sign you out on this device.".

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,166 @@
...
+
+<template>
+  <details-summary
+    id="connected-apps"
+    :title="t('views.mail.views.securitySettings.connectedApps')"
+    :expandable="false"
+    default-open
+  >
+    <template #icon>
+      <ph-plugs-connected size="24" />
```
MelissaAutumn, line 91: Can we add a `aria-label=""` so screen readers know to ignore this icon? (This might be better living in the slot tbh.)

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,166 @@
...
+    console.log(error);
+    errorMessage.value = t('views.mail.views.securitySettings.errorRemovingAccess');
+  }
+};
+
+onMounted(async () => {
+  try {
+    connectedAppsData.value = await getConnectedApps();
+  } catch (error) {
+    console.log(error);
```
MelissaAutumn, line 75: Can we give context to this error log?

src/thunderbird_accounts/settings.py
```diff
@@ -389,6 +390,8 @@ def before_send(event: Event, hint: Hint) -> Event | None:
     OIDC_OP_JWKS_ENDPOINT = os.getenv('OIDC_URL_JWKS')
     OIDC_STORE_ACCESS_TOKEN = True  # Needed to talk to jmap
     OIDC_STORE_REFRESH_TOKEN = True  # Needed to refresh existing sessions. Custom for #498
+    OIDC_STORE_ID_TOKEN = False  # The access token is the canonical source for the Keycloak session id.
+    OIDC_USE_NONCE = True  # Require standard OIDC nonce generation and validation.
```
MelissaAutumn, line 394: This defaults as true, why explicitly mark it in this PR?

src/thunderbird_accounts/settings.py
```diff
@@ -389,6 +390,8 @@ def before_send(event: Event, hint: Hint) -> Event | None:
     OIDC_OP_JWKS_ENDPOINT = os.getenv('OIDC_URL_JWKS')
     OIDC_STORE_ACCESS_TOKEN = True  # Needed to talk to jmap
     OIDC_STORE_REFRESH_TOKEN = True  # Needed to refresh existing sessions. Custom for #498
+    OIDC_STORE_ID_TOKEN = False  # The access token is the canonical source for the Keycloak session id.
```
MelissaAutumn, line 393: Ditto. But we also add session id to all tokens fyi,.

Dockerfile
```diff
@@ -3,6 +3,12 @@ FROM python:3.13-trixie
 # Dockerfile for Thunderbird Accounts
 #
 
+# This a Stable URL for this content that's regularly refreshed with new data.
+# If ever it fails, compatible data sources are available in a number of places.
+# Check: https://github.com/wp-statistics/DbIP-City-lite/
+# Upstream:  https://db-ip.com/db/lite.php
+ARG GEOIP_CITY_MMDB_GZ_URL=https://cdn.jsdelivr.net/npm/dbip-city-lite/dbip-city-lite.mmdb.gz
```
MelissaAutumn, line 10: Aw ye, another supply chain attack opportunity. 😄 

(no action on this, I always find these additional cdn amusing.)

MelissaAutumn (changes requested):
Gotta keep shovelling this feedback on you. Sorry!

src/thunderbird_accounts/authentication/api.py
markstos: I see now that file was working like the V in MVC.

I reluctantly moved the methods to related "utils" class as nothing else seemed to be a better fit and they didn't seem worth starting a new file for.

src/thunderbird_accounts/authentication/api.py
markstos: This is an intentional design detail to make the IPs available in tool tips for advanced users who care and becomes sometimes-- especially with our free GeoIP database--  the location translation is going to be bunk.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/SecurityAccessTable.vue
markstos: I read some about how this is used. Currently it doesn't seem like it doesn't a11y much if the inner text is clear, but that may change. There's also an open WHATWG proposal to let the browser render the date itself based on the <time> tag, so besides being semantic, it's also forward-looking to support the `<time>` tag. 

I've pushed a commit to address this.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/SecurityAccessTable.vue
markstos: Explained above-- it's also what some other providers do. It's a way to make the UI cleaner and simpler for non-tech users while allowing those who really need the IP a way to get it.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: Good idea, though the correct tag is `aria-hidden="true"`.  I've addressed the two cases here an found several more places in other code where we have decorative icons without this, so I'll open a follow-up issue to address those.

Ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden

Dockerfile
markstos: It's a fair concern. We can mitigate it by storing and checking a content checksum. Making a self-note here about that.

assets/app/vue/views/MailView/views/SecuritySettingsView/index.vue
markstos: I see, although that allow-list component had already been merged to main, it became  unintentionally visible during this work. I've pushed a commit with `v-if` disabling.

laurelterlesky:
I've finished up the iteration on Account access -> https://zpl.io/1MZ8PQZ

Hopefully I've captured all the pieces. Let me know if you see something that needs a tweak.

App passwords are OOS. Let's do a future iteration on that when it's time.

laurelterlesky:
xs and dark mode coming next week

Dockerfile
markstos: Ah, the original design of a stable URL that points to changing content is at odds with the idea of adding a stable checksum to it.

I reviewed the risk here and I think it's acceptable: we don't execute the file. There are some narrower risks that I think are acceptable here.

src/thunderbird_accounts/settings.py
markstos: For explicit clarity. 

It's helpful to see the complete set of env vars that that can influence an app and their values in one place.

src/thunderbird_accounts/settings.py
markstos: Same response as above.

For explicit clarity. 

It's helpful to see the complete set of env vars that that can influence an app and their values in one place.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: Good suggestion. Implemented.

assets/app/vue/views/DashboardView/components/PrivacyAndDataCard.vue
markstos: Logout from All sessions was removed the design spec, so this is dead code now. Removing.

assets/app/vue/locales/en.json
markstos: I've pushed an update with your recommendation.

src/thunderbird_accounts/authentication/api.py
markstos: You are right— that's cleaner, more idiomatic Django syntax. Fixed.

markstos:
I believe I've updated all the outstanding feedback from both @MelissaAutumn  and @laurelterlesky  with the following notes:

- Basic relative date display is implemented. To manage the scope here, the ideal solution of a re-usable, more fully-featured relative date component was pushed to a new issue: https://github.com/thunderbird/services-ui/issues/326

- Long text trunction is deferred to new issue to handle this a shared component, to manage scope creep: https://github.com/thunderbird/services-ui/issues/325

- Following the latest design spec, the related UI is split between a summary at `/settings` and details at `/settings/security`, with `/settings` being accessed by a new Settings icon. All of this is now gated by a new `settings-page` Waffle flag.

More information architecture work is underway to move more content to the new settings
page, led by @davinotdavid. Fleshing out the rest of the new settings page is out of scope here.

## Latest screenshots

These don't show locations that actually resolve, but the formatting of those hasn't change from above.

image

image

MelissaAutumn:
Only issue left for me is there's a contrast issue with the "Current session" text. 
https://webaim.org/resources/contrastchecker/?fcolor=737584&bcolor=F7F7F7 (Also available in Firefox's accessibility checker in dev tools.) 

I think if we bump it to `--colour-ti-secondary` and bold it, it should be more readable. 

Also lint errors (and maybe unresolved imports.)

markstos:
> Only issue left for me is there's a contrast issue with the "Current session" text.
> https://webaim.org/resources/contrastchecker/?fcolor=737584&bcolor=F7F7F7 (Also available in Firefox's accessibility checker in dev tools.)
> 
> I think if we bump it to --colour-ti-secondary and bold it, it should be more readable.

@laurelterlesky Could you advise on resolving this contrast issue?

markstos:
@MelissaAutumn I've pushed some more fixup commits to deal with linting issues and confirmed with @laurelterlesky that using a secondary, not muted, color was correct in the table. Good catch there!

MelissaAutumn:
Looks like we have some test failures :sweat_smile:

markstos:
@MelissaAutumn I've pushed a fixup commit which should resolve the test failures.

MelissaAutumn (approved):

PR #1165, opened 2026-08-03
fix: make custom domain deletion idempotent

### LINKED ISSUE #1164: double-clicking "delete domain" can result in unhandled Domain.DoesNotExist exception
If you quickly click "delete" on a custom domain, the first one will succeed and the second will throw an error `Domain.DoesNotExist`.

... at least this pattern is the common reason why exceptions like this get thrown.

The recommended remediation is specifically catch the `DoesNotExist` error, log at the info level to note this happened but otherwise consider the operation a "success" since we were trying to delete the domain and found that it was already deleted. Goal achieved!

----
Sentry Issue: [THUNDERBIRD-ACCOUNTS-99](https://thunderbird.sentry.io/issues/7626631532/?referrer=github_integration)

```
Domain.DoesNotExist: Domain matching query does not exist.
...
  File "thunderbird_accounts/subscription/decorators.py", line 47, in _view_wrapper
    return view_func(request, *args, **kwargs)
  File "thunderbird_accounts/mail/views.py", line 347, in remove_custom_domain
    domain = request.user.domains.get(name=domain_name)
```

mgalore:
## Summary
- treat a second custom-domain deletion request as a successful no-op when the domain was already removed
- add regression coverage for the rapid double-delete path

Fixes #1164.

## Testing
- `npm ci && npm run build && npm run build-theme`
- `coverage run --source='thunderbird_accounts' manage.py test thunderbird_accounts -v 0`
- `coverage report` (516 tests passed; 85% total coverage)
- `ruff check`
- `git diff --check`

### DIFF
--- a/src/thunderbird_accounts/mail/tests/test_views.py
+++ b/src/thunderbird_accounts/mail/tests/test_views.py
@@ -621,6 +621,25 @@ class RemoveCustomDomainTestCase(TestCase):
 
     @patch('thunderbird_accounts.mail.views.mail_tasks.delete_hosted_dkim_dns_records.delay')
     @patch('thunderbird_accounts.mail.views.MailClient')
+    def test_deleting_an_already_removed_domain_is_successful(
+        self,
+        mock_mail_client_cls,
+        mock_delete_hosted_dkim_dns_records,
+    ):
+        mock_instance = Mock()
+        mock_mail_client_cls.return_value = mock_instance
+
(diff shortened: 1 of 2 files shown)


markstos (approved):
Looks good, thanks.

Thanks also to @JiataiWang who submitted  a similar contribution.

## QA Log

- Reviewed diff, self-testing notes and test coverage

PR #1169, opened 2026-08-03
Refactor: enforce keeping Vue and Django routes in sync.

### LINKED ISSUE #408: Have a JSON file for routes shared with Vue and Django

markstos:
## What changed?

Forces public Vue routes to stay since with list in Django by using
a unit test that fails if they go out of sync.

## Why?

This fixes #408.

The solution is an alternative to #473, which solved the issue with a
shared JSON file.

Here, instead of adding another layer of abstraction, we use a "trust but verify"
design. Comments on both sides hint to the user to keep the other side in
sync, and then a unit test verifies this actually done.

This meets the quality goal to keep these in sync. with less complexity in the codebase.

## Notes

The refactor exposed that the current had been of sync, with Django
listing several routes as public Vue routes that are not.

The removed entries were /privacy, /terms, /logout, and /error.
They are no longer Vue routes.

## Applicable Issues

 * #408

## QA Log

- Automated tests
- Checked and double-checked that the removed routes are no longer Vue routes.

### DIFF
--- a/src/thunderbird_accounts/core/tests/test_views.py
+++ b/src/thunderbird_accounts/core/tests/test_views.py
@@ -1,14 +1,48 @@
 import json
+import os
+import re
 from django.utils.html import strip_tags
 from unittest.mock import patch, Mock
 
 from django.conf import settings
-from django.test import TestCase, Client as RequestClient
+from django.test import SimpleTestCase, TestCase, Client as RequestClient
 from django.urls import reverse
 
 from thunderbird_accounts.authentication.models import User
+from thunderbird_accounts.core.views import PUBLIC_VUE_ROUTES
 from thunderbird_accounts.legal.models import LegalDocument, LegalDocumentResponse
 from thunderbird_accounts.mail.models import Account
+
+
+class PublicVueRouteSyncTestCase(SimpleTestCase):
+    """Test that public routes stay in sync between Vue and Django"""
+
+    def test_django_public_vue_routes_match_vue_router_public_routes(self):
+        vue_public_routes = self._extract_vue_public_routes()
+
+        self.assertSetEqual(vue_public_routes, PUBLIC_VUE_ROUTES)
+
+    def _extract_vue_public_routes(self):
+        router_path = os.path.join(settings.BASE_DIR, 'assets', 'app', 'vue', 'router.ts')
+        router_source = router_path.read_text()
+        route_matches = list(re.finditer(r"^\s+path:\s*'([^']+)'", router_source, re.MULTILINE))
+        public_routes = set()
+
+        for index, route_match in enumerate(route_matches):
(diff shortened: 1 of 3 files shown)


src/thunderbird_accounts/core/tests/test_views.py
```diff
@@ -1,16 +1,50 @@
...
+class PublicVueRouteSyncTestCase(SimpleTestCase):
+    """Test that public routes stay in sync between Vue and Django"""
+
+    def test_django_public_vue_routes_match_vue_router_public_routes(self):
+        vue_public_routes = self._extract_vue_public_routes()
+
+        self.assertSetEqual(vue_public_routes, PUBLIC_VUE_ROUTES)
+
+    def _extract_vue_public_routes(self):
+        router_path = Path(settings.BASE_DIR) / 'assets' / 'app' / 'vue' / 'router.ts'
```
MelissaAutumn: Let's stick with os.path.join instead of these slashes.

MelissaAutumn (approved):
I like this, thanks!

src/thunderbird_accounts/core/tests/test_views.py
markstos: Sounds good, less overloading magic.

markstos: I circled back here to see if I could add a `ruff` rule to enforce that we prefer `os.path.join`, but in the process I found the community has moved in the opposite direction. `pathlib` and `Path()` is more modern and considered by many to be the readable approach... and Ruff does have a rule but only in the *opposite* direction, to enforce using `pathlib` instead of `os.path`. 

Ruff's path about the rule links to some references which helped me to understand the rationale:

https://docs.astral.sh/ruff/rules/os-path-join/

With a further search, I found that our code base already uses throughout-- in 14 places. This refinement introduced the only case where we using `os.path`.

markstos: Oy, turns out my "simple" refactor to use `os.path.join` broke the bold. Given what I learned since then about `pathlib` vs `os.path`, I reverted the refactor to back to use `pathlib`:

https://github.com/thunderbird/thunderbird-accounts/pull/1177/changes/0ec0b1a1c7177602f1979581adc31d8cf1ed1453

PR #1181, opened 2026-08-04
Add input validation for internal Paddle endpoint (#1180)

### LINKED ISSUE #1180: Missing input validation on /api/v1/subscription/paddle/tx/set/
Sentry Issue: [THUNDERBIRD-ACCOUNTS-9J](https://thunderbird.sentry.io/issues/7646754912/?referrer=github_integration)

## Example failure mode

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
...
  File "thunderbird_accounts/subscription/views.py", line 81, in set_paddle_transaction_id
    data = json.loads(request.body.decode())
```

markstos:
/api/v1/subscription/paddle/tx/set/ is only called by us, internally for a logged-in
user with CSRF protection. Sentry captured a zero-byte request here, perhaps
from a DevTools modification to the call.

The validation covers that edge case. A further investigation was done
to check there could be malicious benefit to custom-crafting the inputs
here and there's no real security risk for a user to do so.

## What changed?

Input validation was added.

## Why?

Sentry picked up an unhandled error due to bad input here.

## Applicable Issues

 * Fixes #1180

## QA Log

- Added automated test coverage
- Confirmed zero-byte call would be handled by this.
- Analyzed further to see if there's a deeper root cause or a security issue.
  Appears it was a likely triggered by DevTools modification and there's not
  deeper security issue found.

### DIFF
--- a/src/thunderbird_accounts/subscription/tests/test_views.py
+++ b/src/thunderbird_accounts/subscription/tests/test_views.py
@@ -26,6 +26,26 @@ class PaddleCheckoutIsDoneTestCase(TestCase):
         data = txid_response.json()
         self.assertTrue(data.get('success'))
 
+    def test_set_paddle_transaction_id_rejects_invalid_json(self):
+        response = self.client.put(
+            reverse('paddle_txid'),
+            data='invalid json',
+            content_type='application/json',
+        )
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json(), {'success': False, 'error': 'Invalid request data'})
+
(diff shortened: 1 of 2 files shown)


MelissaAutumn (approved):
Thanks, can we cut another ticket for the front-end side of this? At some point we're setting txid but the frontend isn't sending json as expected.

markstos:
@MelissaAutumn I already analyzed the angle of a problem further upstream and it does not appear to be the case. Rather, it appears someone was playing in DevTools.

PR #1183, opened 2026-08-05
fix(mail): return 503 instead of 500 when the mail backend is unavailable

### LINKED ISSUE #1123: Reconsider HTTP 500 for "Error Verifying domain".
Perhaps we should automate retries in the error case of "Error verifying domain"  or downgrade to a "503 Service Unavailable" to represent the temporary nature of the backend being down, or both. But yes, we should still probably notify Sentry if the backend is down after some retries.

Relative priority is low. This is rarely triggered.

Related code is in thunderbird_accoutns/mail/views.py, in `verify_custom_domain` around line 325. 

Sentry Issue: [THUNDERBIRD-ACCOUNTS-91](https://thunderbird.sentry.io/issues/7620714555/?referrer=github_integration)

```
Error verifying domain: 500 Server Error: Internal Server Error for url: https://mailstrom-prod-management-i.thundermail.com:8080/api/principal/deploy
```

shariqueahmad108-ship-it:
Picks up #1123.

`verify_custom_domain()` catches everything in one broad `except Exception` and always returns a 500. But the case the issue is about — the Stalwart backend being down — isn't really an internal error on our side; it's a transient upstream outage. The client gets a flat 500 with no way to tell "the backend is having a moment, retry" from "something is actually broken", and the domain ends up marked `FAILED` for what may just be a blip.

So I split the handling: a `requests.RequestException` (what `MailClient` raises via `raise_for_status` on a backend 5xx, plus connection/timeout errors) now returns **503 Service Unavailable** with a "try again in a few minutes" message, leaves the domain status untouched instead of flipping it to `FAILED`, and still reports to Sentry so a real outage is alerted. Anything genuinely unexpected keeps the original 500 + `FAILED` behaviour.

Tests added in `mail/tests/test_views.py`:
- backend raises `requests.RequestException` → 503, domain stays `PENDING`
- an unexpected error → 500, domain marked `FAILED`

Local run: `VerifyCustomDomainTestCase` is green (8 tests), `ruff check` clean.

One thing worth a maintainer's eye: I deliberately stopped marking the domain `FAILED` on the transient path, since telling the user to retry while also marking it `FAILED` felt contradictory. Happy to revert that part if you'd rather keep the status change.

Closes #1123.

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -262,6 +262,7 @@
           "domainAlreadyExists": "Domain already exists",
           "domainAlreadyConfigured": "This domain is already configured. Please {link} in case you believe this is a mistake.",
           "reachOutToSupport": "reach out to support",
+          "mailBackendUnavailable": "Mail server unavailable. Please try again in a few minutes.",
           "verificationValidationErrors": {
             "mxLookupError": "Unable to find MX record pointing to Thundermail",
             "ipLookupError": "Unable to resolve IP addresses",
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/components/ActionsMenu.vue
@@ -54,6 +54,14 @@ const handleRetry = async () => {
   try {
     const data = await verifyDomain(props.domain.name);
 
+    // Transient mail-backend outage: surface the error notice and don't touch the
+    // domain's verification state (don't apply results or mark it FAILED).
+    if (data.code === 'mail_backend_unavailable') {
+      emit('custom-domain-error', t('views.mail.sections.customDomains.mailBackendUnavailable'));
+      showMenu.value = false;
+      return;
+    }
+
     emitVerificationResult(data);
 
     if (data.success) {
--- a/assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
+++ b/assets/app/vue/views/MailView/sections/CustomDomainsSection/components/CustomDomainForm.vue
@@ -463,6 +463,13 @@ const onVerifyDomain = async () => {
   try {
     const data = await verifyDomain(customDomain.value);
 
+    // Transient mail-backend outage: surface the error notice and leave the
+    // domain's verification state untouched (don't apply results or mark FAILED).
+    if (data.code === 'mail_backend_unavailable') {
+      customDomainError.value = t('views.mail.sections.customDomains.mailBackendUnavailable');
+      return;
+    }
+
     applyVerificationResult(customDomain.value, validationResult(data), {
       showMissingIssues: true,
       cacheResult: true,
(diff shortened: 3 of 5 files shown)


markstos (changes requested):
Thanks for this contribution.

This is exactly the direction we want to head in with this.

Before merging, could you please update this PR to:

 - Add a stable response code such as `mail_backend_unavailable` in `src/thunderbird_accounts/mail/views.py` and a  localized message in `assets/app/vue/locales/en.json`: “Mail server unavailable. Please try again in a few minutes.” Handle that code in  `CustomDomainForm.vue` and `ActionsMenu.vue` before applying verification results or emitting FAILED; display it through the existing error notice instead.

  - Narrow the backend classification to check for transient error HTTP status codes in `src/thunderbird_accounts/mail/views.py`, since `requests.RequestException`   also includes non-transient 4xx responses.

  - Update `src/thunderbird_accounts/mail/tests/test_views.py` so the `/principal/deploy` failure is raised by `create_domain()`, after DNS verification succeeds, rather than by `check_domain_dns()`.

markstos:
Not to self: as a follow-up in another issue/PR let's look adding some intelligent retry logic right into MailClient to smooth over more Stalwart blips automatically.

shariqueahmad108-ship-it:
Thanks for the detailed review! Addressed all three in the latest commit:

- **Stable code:** returns `mail_backend_unavailable` now; added the localized `mailBackendUnavailable` string in `en.json`, and `CustomDomainForm.vue` / `ActionsMenu.vue` handle the code via the existing error notice instead of emitting FAILED.
- **Narrowed classification:** only transient failures (5xx, or connection/timeout with no response) return 503 — non-transient 4xx now falls through to the existing 500 + FAILED path.
- **Test:** the `/principal/deploy` failure is now raised by `create_domain()` after DNS verification succeeds, and I added a case asserting a 4xx is treated as a genuine failure, not a retry.

Ready for another look whenever you have a chance.

markstos:
@ryanjjung FYI:  I don't expect Infra is using Sentry exception rates as a strong signal that Stalwart is having problems, but here we stop sending Sentry exceptions for transient issues to reduce noise for issues that are not actionable.

markstos:
@shariqueahmad108-ship-it You'll see I pushed a refinement commit.

If you contribute more in the future, it's useful to manually test any visual changes and include screenshots of the results. Here, it appears the visual banner was missed-- I addressed that in my follow-up commit. Here's how you can visually test a feature like this:

1. Manually or with AI, make a temporary modification to the code that triggers the condition to display. (Or I suppose for this, you could turn off the backend Stalwart server temporarily.
2. From the project root: `docker compose up -V`. That will run all the necessary containers in the foreground until you cancel.
3. In a browser, visit http://localhost:8087/
4. OK... there are some steps to bootstrap which are documented in the README.md, but may not be perfect.  

There's a public Matrix #tb-services channel on the Mozilla.org Matrix server for questions or suggestions for contributor doc improvements. 

## Screenshots

image

image

markstos (approved):

PR #1184, opened 2026-08-05
Relax SRV priority check to be "lowest" instead of "1". (#1163)

### LINKED ISSUE #1163: Issue a  warning if our SRV entry doesn't have the highest priority, and maybe mention it in the tips
Some DNS providers like Metanet's PLESK panel, don't allow `1` just `0`

* METANET's Plesk panel (see https://github.com/thunderbird/dns-scripts/issues/13) renders the SRV weight ("Relative Gewichtung für Einträge mit gleicher Priorität") as a dropdown with fixed steps of 5 — the only selectable
values are: `sehr hoch (50), 45, 40, 35, 30, 25, 20, 15, 10, 5, niedrig (0)`
* `1` is not allowed, so a METANET customer cannot enter the record as documented.
* from @kewisch  [in Matrix](https://matrix.to/#/!YXxorOjEQwXtUAQdOA:mozilla.org/$HkCH90MPuWrdDAztMq-GOrlOY2ImIrFj93Wt7TO5WDA?via=mozilla.org&via=matrix.org):
`Saw this somewhere recently. I think all we should check is a warning if our SRV entry doesn't have the highest priority, and maybe mention it in the tips. If we don't have an issue please file.`

markstos:
This solves a problem with registrars who don't allow an exact "1" value.

We also change our default recommended value to zero, the lowest. This
addresses registrars like METANET and is also aligned with our intent
to have the highest priority (lowest numbered) records.

## What changed?

- Instead of checking for exactly the value "1", we check that we have the lowest value.
- Instead of recommending a priority of "1", we now recommended "0", an even lower default.

## Why?

As explained in #1163, this solves issues with some registrars.

## Notes

 * It's backwards compatible with all properly configured domains. It would start to flag an edge case as a problem if a user had our correct record at priority 1, but some broken record at
   priority 0. I've heard of no such cases and since they not configured ideally anyway, it could
   be considered a feature to surface them.

## Applicable Issues

 * Fixes #1163

## QA Log

 * Besides automated testing, manual testing for the happy and sad paths was done. See screenshots.

## Screenshots

image

image


src/thunderbird_accounts/mail/dns.py
```diff
@@ -92,21 +90,18 @@ def check_srv_record_status(domain_name: str, record: dict) -> tuple[DNSRecordSt
...
         live_value = f'{rdata.priority} {rdata.weight} {rdata.port} {target}'
         live_values.append(live_value)
-        if (
-            rdata.priority == expected_priority
-            and rdata.weight == expected_weight
-            and rdata.port == expected_port
-            and target.lower() == expected_target
-        ):
-            has_match = True
+        highest_priority = rdata.priority if highest_priority is None else min(highest_priority, rdata.priority)
```
MelissaAutumn: If you set highest_priority to 999 or like MAX_INT you can remove the if. Most likely performantly insignificant though lol!

MelissaAutumn: Also maybe good to have a comment saying "lower is better" for folks unfamiliar with dns.

MelissaAutumn (approved):
Code looks okay on my end.

MelissaAutumn:
Ah the test failure is because this needs a rebase. Cool, all good here.

src/thunderbird_accounts/mail/dns.py
markstos: Clever idea for avoiding unnecessary "if". I implemented by using the Maximum allowed SRV priority value according to the SRV. Nice touch.

I also refactored to avoid awkward "highest priority" wording because lowest-is-best and addded a code comment about the spec oddity as well.

Thanks for the feedback.

PR #1189, opened 2026-08-10
fix: handle expected Paddle webhook rejections

### LINKED ISSUE #1187: Handle "Too much time has elapsed between the request and this process" case
We are seeing some Sentry noise periodically from pen testers testing `POST /api/v1/subscription/webhook`.

We can improve our "sad path" handling there by distinguishing between the error states we expect from Sentry and those that we don't. For expected conditions, we handle those by only logging the condition. For truly unexpected cases, we can continue to notify Sentry.

Here are the two expected Paddle rejections we know about:

```
  EXPECTED_PADDLE_REJECTIONS = {
      "Unable to extract the 'Paddle-Signature' header from the
      request",
      "Too much time has elapsed between the request and this
      process",
  }
```

Those are worth logging at the info level. Other failures can continue to notify Sentry.

To be clear: There's not a known vulnerability here. The task relates to helping keep our Sentry alerts actionable.

Sentry Issue: [THUNDERBIRD-ACCOUNTS-78](https://thunderbird.sentry.io/issues/7427584030/?referrer=github_integration)

```
Too much time has elapsed between the request and this process
```

atirna:
## What changed?

- Treat the two known Paddle webhook verification rejections as expected authentication failures: log them at info level and reject the request.
- Preserve Sentry reporting for unexpected verification exceptions.
- Add focused regression coverage.

AI assistance: an agent drafted the implementation and tests; the contributor approved the final scope and publication.

Tests: `uv run python manage.py test thunderbird_accounts.authentication.tests.test_permissions`; `uv run ruff check src/thunderbird_accounts/authentication/permissions.py src/thunderbird_accounts/authentication/tests/test_permissions.py`

## Why?

Keeps known penetration-test Paddle requests out of Sentry while retaining visibility for unexpected failures.

## Limitations and Notes

No UI changes.

## Applicable Issues

Closes #1187

## Screenshots

Not applicable (no UI changes).


gokul-debugger:
I checked this against the locked `paddle-python-sdk` 1.15.0 behavior. `Verifier.verify()` does not raise for either of these conditions; it logs the message at critical level and returns `False`. As a result, the new `except` branch and tests using `side_effect = Exception(...)` do not reproduce the production path.

A narrowly scoped filter on the `paddle_billing` logger, or another approach that handles these log records before Sentry processes them, may be needed instead. The relevant SDK behavior is visible in [`Verifier.py`](https://github.com/PaddleHQ/paddle-python-sdk/blob/v1.15.0/paddle_billing/Notifications/Verifier.py).

markstos (approved):
Here's some PR feedback. In the future, it's useful to see any self-QA you've done besides adding automated tests, as sometimes tests can pass but something still isn't right.

After consideration, I removed the test coverage. I realize there's a general philosophy of "everything should have a test"-- and there's a good starting point but every test as a cost and risk associated and here I don't think the the test carried its weight.

In terms of cost, there was 36 lines of additional testing to maintain to check re-classifying two log levels. This makes the test suite run every so slightly slower to include them, and the test suite is an accumulation of little choices like this.

Then there's risk: the risk is very low here if the classification didn't actually work: we might end up seeing the logs again in Sentry.  

## QA Log

- Reviewed code
- Reviewed `logging` module docs
- Manually tested that filtering works as expected
- Analyzed Paddle SDK to confirm this `paddle_billing` logging namespace is used.
- Analyzed Sentry SDK to confirm that reclassifying to `info` is sufficient here.

markstos:
@gokul-debugger Thank you for peer-review on this PR. You had good insight and validated my decision later to remove the test coverage. The first draft of the PR including tests but the behavior wasn't actually right. Not all test coverage enhances the quality of the work. Some can signal quality or correctness that's not actually there.

In the future when I have competing contributor PRs to review, I'll be looking at the quality of the self-QA notes that go beyond what the AI does, because it's the quality of that human review that determines how much time I'll spend or save doing the final QA.

Thanks to both contributors here for a rapid collaboration to improve Thundermail!

atirna:
hey @markstos thank you so much taking time to write the feedback, duly noted i'll be sure to add a QnA log as you suggested next time and i'll be more deliberate with my use of tests as to not cause unnecessary bloat.

gokul-debugger:
Thank you for the thoughtful feedback and acknowledgment, Mark. The distinction between adding coverage and validating the real production path is a useful lesson. I’ll include concise self-QA notes in future contributions, especially when behavior depends on third-party libraries or logging integrations. I’m glad the review helped, and congratulations to @atirna on getting the fix merged.

PR #1203, opened 2026-08-18
Show tooltip upon clicking the Connect Now button in Mail Dashboard

### LINKED ISSUE #1209: Add tooltip for "Connect Now" button
When we click the "Connect Now" button, we don't know if the user has Thunderbird Desktop installed or not so we should proactively show a helpful tool-tip with instructions to download and install it if need be.

davinotdavid:
## What changed?

Added a tooltip on click of the `Connect Now` button for better feedback in case TB Desktop is not installed.

To test this, please add the `show-connect-now` to the django-waffle flags in Django Admin.

## Why?

In case the user does not have TB Desktop installed, the button feels broken as there's no feedback whatsoever of clicking it.

## Limitations and Notes

AFAIK, there's no way to reliably catch a failure on the custom mapping protocol so showing the tooltip simply upon click seems to be the best course of action here.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/1209

## Screenshots

image

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -114,6 +114,8 @@
               "connectTitle": "Connect Thunderbird Desktop",
               "connectDescription": "Add Thundermail to Thunderbird Desktop automatically.",
               "connectButton": "Connect now",
+              "connectButtonTooltip": "Thunderbird Desktop did not open? Please {downloadLink} and click 'Connect now' again.",
+              "connectButtonTooltipDownloadLink": "download and install the app",
               "downloadTitle": "Download Thunderbird Desktop",
               "downloadDescription": "We recommend using the desktop application for the best experience.",
               "downloadButton": "Download",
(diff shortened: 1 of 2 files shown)


markstos (changes requested):
The code looks good to me. For the tooltip, could we give it a 5 second delay to appear? What happens when I click "Connect now" is that two things happen at once. One, I get a browser modal asking me if I want to open the link in Thunderbird and then at the same time, in the background, I get this "did not open?" tooltip.

It would feel cleaner if we could give the "happy path" a chance to work before displaying this tooltip about the sad path. So if the happy path is working, the user is not distracted with this modal interrupting their happy path journey.

Image

## QA Log

 - Reviewed diff,
 - Manually tested.

davinotdavid:
@markstos Hmm that's fair. As a middle ground I've added a 3 seconds delay there as the custom mapping protocol should trigger instantly if TB Desktop is installed / configured and I think that 5 seconds is a bit too long for people to wait in case they don't have it installed / configured.

markstos (approved):

Issue #1190, opened 2026-08-10
In settings, allow to view / manage "Connected Apps" sessions


markstos:
Besides normal "Login Sessions" to webmail and the account settings, another kind of session that's managed separately is known as "Connected Apps" in UX, and technically known as [OAuth with the offline access scope](https://auth-wiki.logto.io/offline-access). This is used by Thunderbird and other email clients that allow you authenticate *once* at setup via OAuth and don't require you to enter the password again.

With normal login sessions, can "logout" and have the client notice immediately, if it has implemented single-log-out (SLO) so that we can contact it immediately, or will notice relatively soon when an access token expires and can't be refreshed.

"Connected Apps" are in their own category because they have long-lived access tokens.

More about [Keycloak offline tokens](https://skycloak.io/blog/keycloak-offline-tokens-explained/).

The presentation to view and manage Connected Apps is similar. A list of connected apps may include:

    * App name / type of app or acceses
    * App logo
    * Date access was granted
    * Date of last access
    * Last access IP / location
   
In reviewing the access details for a connected app, they may be presented with a list of what the app has access to, along with a "remove access" button.

Issue #1158, opened 2026-07-28
Support back channel logouts (related to SLO aks single-logout)


MelissaAutumn:
Keycloak supports back channel logouts. We should enable and catch these, then invalidate the user's django session. 

<img width="996" height="507" alt="Image" src="https://github.com/user-attachments/assets/fd63c30f-1c72-4a5e-ba0f-f1c4343199e9" />

As part of implementing this, we should consider if we still need the stop-gap from https://github.com/thunderbird/thunderbird-accounts/pull/1054

PR #1204, opened 2026-08-18
Add ability to view/remove Connected Apps (aka OAuth-authenticated apps with offline_access scope)

markstos:
## Refactor notes

The following related refactors are included:

- To support this work, a central Django error handler was added that
returns JSON errors. This DRY's a common case of adding try/catch/rethrow JSON in routes. Now we can leave error handling up to the central handler
in more cases.

- Keycloak tests are moved out of MFA tests for clarity

- On the frontend, now that we have two related tables of access
  details for "Current Sign-ins" and "Connected Apps", refactor
  so that they share more code.

## What changed?

Implements view / management of connected apps as described in #1190

## Why?

 * This was a natural small scope expansion to the Current Sign-In work.

## Limitations and Notes

 * This is a manually stacked PR! It merges into the main "Curreng Sign-in work" not the main branch.

## Applicable Issues

 * [In settings, allow to view / manage "Connected Apps" sessions · Issue #1190](https://github.com/thunderbird/thunderbird-accounts/issues/1190)

## QA Log

 * The IP to Location work is copied from the earlier case, so is not re-tested.
 * Manually tested the key cases:
   * No Connected Apps
   * Connected App appears after connecting with Thunderbird Desktop with offline_access scope
   * Revoking the access works (Entry disappears from table)
 * Reviewed auth to make sure route guards match other recent work.

For testing with Thunderbird, I used this extension to be able to connect Thunderbird desktop to local Stalwart/Keycloak:

https://github.com/raa-org/thunderbird-custom-idp
AI was used to assist with the various step steps.

## Screenshots

image

image


assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
```diff
@@ -89,33 +95,24 @@ onMounted(async () => {
...
+          primary: t('views.mail.views.securitySettings.recordsTableHeaderDevice'),
+          ipAddress: t('views.mail.views.securitySettings.recordsTableHeaderIpAddress'),
+          location: t('views.mail.views.securitySettings.recordsTableHeaderLocation'),
+          lastAccess: t('views.mail.views.securitySettings.recordsTableHeaderLastActive'),
+          actions: t('views.mail.views.securitySettings.recordsTableHeaderActions'),
+        }"
+        :attribution="t('views.mail.views.securitySettings.ipGeolocationAttribution')"
+      >
+        <template #action="{ record }">
+          <span v-if="record.isCurrent" class="current-session-label">This is you</span>
```
davinotdavid: This should be a i18n string.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,137 @@
...
+const loading = ref(true);
+const errorMessage = ref<string | null>(null);
+
+const removeAccess = async (id: string) => {
+  const app = connectedApps.value.find((connectedApp) => connectedApp.id === id);
+  if (!app) {
+    return;
+  }
+
+  if (!window.confirm(t('views.mail.views.securitySettings.removeAccessConfirmation', { app: app.label }))) {
```
davinotdavid, line 34: I have a feeling that this might be an actual dialog / modal confirmation but can be punted into a separate ticket for later since we don't have designs for it.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
```diff
@@ -89,33 +95,24 @@ onMounted(async () => {
...
-      <p class="location-attribution">
-        <a href="https://db-ip.com" rel="noopener noreferrer" target="_blank">
-          {{ t('views.mail.views.securitySettings.ipGeolocationAttribution') }}
-        </a>
-      </p>
+      <security-access-table
+        :records="activeSessions"
+        :column-labels="{
+          primary: t('views.mail.views.securitySettings.recordsTableHeaderDevice'),
+          ipAddress: t('views.mail.views.securitySettings.recordsTableHeaderIpAddress'),
```
davinotdavid: I am probably out of date in the design decisions but are we showing IP address on the table? According to [Zeplin mockups](https://zpl.io/4EPPEEz), we don't have this column.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,137 @@
...
+      <security-access-table
+        :records="connectedApps"
+        :column-labels="{
+          primary: t('views.mail.views.securitySettings.recordsTableHeaderApp'),
+          ipAddress: t('views.mail.views.securitySettings.recordsTableHeaderIpAddress'),
+          location: t('views.mail.views.securitySettings.recordsTableHeaderLocation'),
+          lastAccess: t('views.mail.views.securitySettings.recordsTableHeaderLastAccess'),
+          actions: t('views.mail.views.securitySettings.recordsTableHeaderActions'),
+        }"
+        :attribution="t('views.mail.views.securitySettings.ipGeolocationAttribution')"
```
davinotdavid: This `attribution` prop is the same for both usages of `security-access-table` so perhaps we could bake it into the component itself?

davinotdavid (commented):
I've tested connecting my local TB Desktop with `localhost` using the `admin@example.org / admin` user (which seems to connect just fine) but TB Desktop doesn't show up in the Connected Apps. Is the extension really mandatory for testing?

I am probably outdated here but I thought the idea was to combine both Current Signins and Connected Apps into a single `Account Activity` table. If that's still the case, then this PR would need to change a bit.

markstos:
>  Is the extension really mandatory for testing?

At least for  using Thunderbird desktop as a testing client, it's the only way I know. Even after setting up the extension, it's necessary that the `offline_access` scope is given, which it is not by default. This exact JSON file works:

```json
{
  "disableExchangeAutodiscovery": true,
  "providers": [
    {
      "name": "local-stalwart-via-keycloak",
      "hostname": "stalwart",
      "issuer": "keycloak",
      "clientId": "thunderbird-local",
      "usePkce": true,
      "authorizationEndpoint": "http://keycloak:8999/realms/tbpro/protocol/openid-connect/auth",
      "tokenEndpoint": "http://keycloak:8999/realms/tbpro/protocol/openid-connect/token",
      "redirectUri": "https://localhost",
      "scopes": {
        "imap": "openid email profile offline_access",
        "smtp": "openid email profile offline_access"
      }
    }
  ]
}
```

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
markstos: Pushed fix.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: Sounds good to defer to for now.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
markstos: Good catch. I've pushed a fix that removes the IP address column but preserves the IP address a a tooltip on the location, providing those want that technical detail a way to find it.

image

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: Good catch on an opportunity to the DRY the code a bit. Pushing  a fix for this.

markstos:
@solangevalverde Here's how Current Sign-ins and Connected Apps are looking post implementation:

image

@davinotdavid I've addressed all your feedback now.

assets/app/vue/views/DashboardView/components/PrivacyAndDataCard.vue
```diff
@@ -35,6 +36,17 @@ const logoutFromAllSessions = () => signOutAllSessions();
         </div>
       </div>
   
+      <div class="divider"></div>
```
davinotdavid, line 39: Unsure if this would help further reduce some code here but we have a `VisualDivider` component:

https://thunderbird.github.io/services-ui/?path=/docs/components-visualdivider--docs

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
```diff
@@ -0,0 +1,135 @@
...
+const { t, locale } = useI18n();
+
+type DisplayConnectedApp = {
+  id: string;
+  clientId: string;
+  label: string;
+  ipAddress: string;
+  location: string;
+  lastAccess: string;
+};
```
davinotdavid, line 22: This could go into `types.ts` instead of being here

assets/app/vue/views/MailView/views/SecuritySettingsView/components/SecurityAccessTable.vue
```diff
@@ -0,0 +1,130 @@
...
+  lastAccess: string;
+  isCurrent?: boolean;
+};
+
+type ColumnLabels = {
+  primary: string;
+  location: string;
+  lastAccess: string;
+  actions: string;
+};
```
davinotdavid, line 20: Similar for this, could go into `types.ts` instead

davinotdavid (approved):
I am still unable to test this locally as I am failing to setup the extension locally even with the provided JSON. I think we can give this a try on stage though, as code LGTM.

Related to the main PR though, it seems that the active sessions are not being revoked as I am still able to peruse Thundermail without being logged out after I click the `sign out` for a previously opened session in another browser.

Also, as a heads up (again not in this PR), the buttons in the Accounts dashboard seem to be re-done in CSS instead of using the components from the design system:

Image

A few nits that can be addressed in the main PR:

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
davinotdavid: @markstos FYI, I don't see the tooltip address in the latest revision but I see that the column has been removed.

markstos:
> it seems that the active sessions are not being revoked as I am still able to peruse Thundermail without being logged out after I click the sign out for a previously opened session in another browser.

I believe the issue here is that the the Keycloak session has been revoked on the server side but the client hasn't been notified. Making that happen immediately is a separate issue. There's the concept of [backchannel logout](https://github.com/thunderbird/thunderbird-accounts/issues/1158) where the identity server notifies other identity clients immediately as well as the related concept of single-logout (SLO) where the identity server forces you to logout of *all* logins for an identity at the same time.  SLO may use a mix of backchannel logout and front-channel-- where the browser is redirected through URLs to make sure they are all touched.

The Context session of the original PR, it describes how there can be a delay of 5 to 15 minutes for the change to take effect. https://github.com/thunderbird/thunderbird-accounts/pull/1162#

MelissaAutumn:
Access token life is set to 5 minutes on prod fyi, and while we can propagate a backchannel logout notification, I don't believe we can back channel logout desktop or external oauth clients. I don't think that's the end of the world, i'm also  unsure how other email companies handle this

markstos:
>  I don't believe we can back channel logout desktop or external oauth clients

Agreed. 

>  I don't think that's the end of the world

Also agreed. I think it may be best handled with some communication: "session revoked. It may take up to 15 minutes for the device to be signed out."  (

15 minutes adds some buffer beyond the expected 5 minutes, the same way we get notices about DNS taking up to 24 hours to propagate or an email taking 24 hours to arrive when they are usually fairly instant.

Some options would include having Stalwart and Stormbox to have super-short sessions or constantly poll Keycloak to see if they were just logged out, but it's not clear that's the best trade-off users.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/AccountActivity.vue
markstos: I double checked the tooltips are there now.

assets/app/vue/views/DashboardView/components/PrivacyAndDataCard.vue
markstos: Good catch. Addressing now for consistency.

assets/app/vue/views/MailView/views/SecuritySettingsView/components/ConnectedApps.vue
markstos: Good call. Moved.

markstos:
This has been manually merged into the "Active Logins" branch.

PR #1210, opened 2026-08-19
Add custom-domains-revamp feature flag + initial boilerplate

davinotdavid:
## What changed?

- Added `custom-domains-revamp` feature flag that:
  - Hides the existing Custom Domain section in `/mail`.
  - Adds `Dashboard` (pointing to /mail) and `Custom Domains` (pointing to /custom-domains) to the main navigation.
- Copied over the existing `Custom Domains` section code & components into its new folder.
- Added `vue/test-utils` and `happy-dom` dev dependencies to facilitate unit / component testing.

## Why?

- This is essentially boilerplate steps to facilitate the development & targetted code review of the actual UI / UX implementation changes for this milestone.
- The idea to copy over the code instead of moving is deliberate since the time between having the milestone fully done / delivered might lag behind potential tinier fixes in the main Custom Domains code. Also, a lot of the code will be reused / repurposed to it would make sense to kick-start it from what we currently have anyways.
- Doing this behind a feature flag increases confidence that the next PRs about this won't block / leak into production.

## Limitations and Notes

- This PR might seem large but vast majority of the code here is just copying over what is currently in the Custom Domains section for future updates! The review here is mostly checking the feature flag and navigation changes more than anything.

## Applicable Issues

Related with https://github.com/thunderbird/thunderbird-accounts/issues/944

## Screenshots

[`/mail` route with the feature flag on]

image

[`/custom-domains` route]

image


markstos (commented):
I'm aligned with the overall idea here of adding a feature flag for this. Some feedback.

- Say more about you expect the new `happy-dom` dep to be used. Right it's used in one HeaderBar test and it's not clear that it carry's its weight.
- Should the actual Django Waffle flag be implemented in this PR? Right now, `/custom-domains` gets unconditionally registered so it could always be navigated to. Or is the idea to use a FE-only feature flag intended for us, but that users might discovered as well?

davinotdavid:
@markstos 

Thanks for the feedback!

- About `happy-dom`, you're right that it is currently only used in one test file but that's also because we only have one component / unit test file at the moment 😅 I guess we could instead rely only on the E2E and remove the dependency and the header test file altogether but having an extra layer of tests that runs much faster seems to be a good idea in my opinion.

- Ah nice catch. Just to cover our bases, I've added a guard not to register the /custom-domain route if the feature flag is not on, similarly with what we are doing for MFA.

markstos (changes requested):
## happy-dom

Regarding happy-dom— sounds like happy-dom is a useful forward-looking dep.

## rebase

When you did your most recent rebase, it sounds like there was a bad merge with upstream:

 The new feature-flagged implementation uses raw fetch() in `assets/app/vue/views/
 CustomDomainsView/CustomDomainsSection/api.ts:8`. Upstream changed the original implementation to `useAuthFetch()` in `assets/app/vue/views/MailView/sections/CustomDomainsSection/api.ts:1`, which handles expired/authentication sessions and redirects for reauthentication.

Consequently, enabling the new feature switches users back to the pre-upstream behavior: an authentication failure can be consumed as ordinary JSON or fail while parsing the response instead of initiating reauthentication. The copied API should be resynchronized with the upstream source.

## Waffle flags

It looks like we are trying to document all of our Waffle flags in [./docs/feature-flags.rst]() but the resource has drifted out of date and isn't linked from the README. I'll open a small PR to address that.

davinotdavid:
@markstos oooh nice, great catch on the divergent `api.ts`. Also added the feature flag to the docs!

markstos (approved):
Feedback is addressed.

Sancus:
Please squash merge unless the commit messages have been amended to properly link to issues and/or the PR itself and are actually intended to land individually.

markstos:
@Sancus Good reminder. I also prefer a cleaner and simpler commit history.

PR #1211, opened 2026-08-20
Fix potential Stalwart v0.16 JMAP API regressions

Sancus:
## Summary

Note: These are all AI-discovered issues and I haven't exhaustively reviewed them all for validity, I am just putting this PR up for reference to avoid duplicated work.

#628 

### JMAP transport security and robustness

1. **TLS verification was disabled and JMAP HTTP calls could hang indefinitely.** TLS verification is now configurable and enabled by default, with a bounded timeout on session discovery and API requests.
   - Fix: [`JMAPClient` transport configuration](https://github.com/thunderbird/thunderbird-accounts/blob/0d3a748cc326398978cc0488a69dd4860747cc59/src/thunderbird_accounts/mail/clients/jmap_client.py#L22-L48)
   - Test: [`test_uses_configured_tls_verification_and_timeout`](https://github.com/thunderbird/thunderbird-accounts/blob/0d3a748cc326398978cc0488a69dd4860747cc59/src/thunderbird_accounts/mail/tests/test_clients/test_jmap.py#L42-L64)

2. **Authenticated JMAP requests followed redirects.** Redirects are now disabled for both discovery and API calls so credentials cannot be forwarded by a redirect.
   - Fix: [session and method request handling](https://github.com/thunderbird/thunderbird-accounts/blob/0d3a748cc326398978cc0488a69dd4860747cc59/src/thunderbird_accounts/mail/clients/jmap_client.py#L65-L74)
   - Test: [`test_authenticated_requests_do_not_follow_redirects`](https://github.com/thunderbird/thunderbird-accounts/blob/0d3a748cc326398978cc0488a69dd4860747cc59/src/thunderbird_accounts/mail/tests/test_clients/test_jmap.py#L68-L82)

3. **A session response could direct authenticated method calls to an unrelated `apiUrl` origin.** The discovered URL must now match the configured scheme, host, and effective port.
   - Fix: [origin validation](https://github.com/thunderbird/thunderbird-accounts/blob/0d3a748cc326398978cc0488a69dd4860747cc59/src/thunderbird_accounts/mail/clients/jmap_client.py#L53-L85)
   - Test: [`test_rejects_session_api_url_on_another_origin`](https://github.com/thunderbird/thunderbird-accounts/blob/0d3a748cc326398978cc0488a69dd4860747cc59/src/thunderbird_accounts/mail/tests/test_clients/test_jmap.py#L85-L91)

4. **Session discovery wrote a debug JSON artifact into the current directory.** The unconditional file write has been removed.
[...]

PR #1217, opened 2026-08-20
fix(jmap): keep the configured origin when the session advertises another host

### LINKED ISSUE #1216: JMAPClient ignores STALWART_JMAP_API_URL after the session handshake, following the server-advertised apiUrl instead
## Summary

`JMAPClient` discards the configured `STALWART_JMAP_API_URL` after the session handshake and
sends every subsequent request — with credentials — to whatever URL the **server** advertises
in its JMAP session resource. In any deployment where Stalwart's advertised public URL is not
the address the app should be using (split-horizon DNS, in-cluster service, private link), the
configured setting is silently ineffective.

## The code

`src/thunderbird_accounts/mail/clients/jmap_client.py` — `base_url` is used in exactly one
request, the session lookup:

```python
# line 45
r = requests.get(f'{self.base_url}/.well-known/jmap', ...)
...
# line 61
self.api_url = session.api_url          # server-supplied, overrides config from here on
```

and `request()` then uses only `self.api_url`:

```python
# line 101
def request(self, request_data: JMapRequest, ...) -> JMapResponse:
    if not self.api_url:
        raise RuntimeError('Session not available')
```

`base_url` never appears again — grep shows it only at lines 18/19/20/24 (constructor) and 45.
So `settings.STALWART_JMAP_API_URL`, passed at `mail_client_jmap.py:59`, controls *where the
client asks for the URL*, not *where it sends requests*.

## Why this is not theoretical

On our tb-dev cluster, Stalwart advertises a hostname belonging to a **different deployment**.
Measured from inside the accounts pod, with the admin credential the app already holds:

| Request | Result |
|---|---|
| `GET <in-cluster>/.well-known/jmap` | **200** — `apiUrl = https://mail.stage-thundermail.com/jmap/` |
| `POST https://mail.stage-thundermail.com/jmap/` | **401** |
| `POST <in-cluster>/jmap` | **200**, `Core/echo` round-trips |

So `MailClientAdminJMAP` fails on every real call:

```
preflight_check -> HTTPError: 401 Unauthorized for url: https://mail.stage-thundermail.com/jmap/
get_domain      -> HTTPError: 401 Unauthorized for url: https://mail.stage-thundermail.com/jmap/
```

The 401 is correct behaviour *from the server's point of view* — it is being handed another
environment's credential. That is the part worth emphasising: the client **transmitted our
Stalwart admin credential to a host we did not configure it to talk to**, over the public
internet. Our advertised-hostname misconfiguration is ours to fix, but a client that only ever
contacts its configured host would have failed closed instead of leaking the credential.

## Suggested fix

Keep the session handshake (it supplies `accountId`, `identityId`, capabilities), but do not let
it relocate the client. Preferred, since the JMAP path may legitimately differ per server:

```python
self.api_url = urljoin(self.base_url, urlparse(session.api_url).path)
```

i.e. take the **path** from the session and the **origin** from configuration, so the client can
never be redirected off its configured host. Same treatment applies to `downloadUrl` /
`uploadUrl` / `eventSourceUrl` if they get used later.

A simpler variant, which is what our deployment currently runs as an overlay patch: skip session
discovery for the admin client entirely and POST to `{base_url}/jmap`. That works today against
Stalwart v0.16.16 (the 200 in the table above) and is what keeps our environment functional.

## Two smaller things in the same file

1. **`verify_ssl` is hardcoded `False`** (`self.verify_ssl = False`, line ~35), so
   `settings.VERIFY_PRIVATE_LINK_SSL` has no effect. It also disabled verification for the
   *public* host above, which is not a private link at all. Suggest taking it from settings.
2. **`get_session()` writes `./d_get_session.json`** on every session fetch (lines ~56-57).
   That dumps session data including account ids into the process CWD, and raises outright under
   a read-only root filesystem. Looks like leftover debugging.

## Context

Found while removing a deploy-side overlay that patched the whole `mail/clients/` package. That
overlay exists specifically because it posts directly to the configured URL; the three
differences above are the reason we cannot yet drop it and run the shipped image as-is.

aatchison:
Fixes #1216.

`JMAPClient` dropped `STALWART_JMAP_API_URL` after the handshake — `get_session()` set `self.api_url = session.api_url`, and `request()` used only that. So the setting decided where we *asked* for a url, not where we sent requests, and a server could point the client (and its credentials) at any host.

## Change

```python
self.api_url = urljoin(self.base_url, urlparse(session.api_url).path)
```

Path from the session, origin from config. Non-default JMAP paths still work; off-host redirection becomes impossible.

Three tests: off-host advertised url, non-default path, already-matching host.

## Not included

Left out to keep this reviewable, both described in #1216: hardcoded `verify_ssl = False` (ignores `VERIFY_PRIVATE_LINK_SSL`), and the `./d_get_session.json` debug write.

### DIFF
--- a/d_get_session.json
+++ b/d_get_session.json
@@ -1,0 +1,245 @@
+{
+  "capabilities": {
+    "urn:ietf:params:jmap:core": {
+      "maxSizeUpload": 50000000,
+      "maxConcurrentUpload": 4,
+      "maxSizeRequest": 10000000,
+      "maxConcurrentRequests": 4,
+      "maxCallsInRequest": 16,
+      "maxObjectsInGet": 500,
+      "maxObjectsInSet": 500,
+      "collationAlgorithms": [
+        "i;ascii-numeric",
+        "i;ascii-casemap",
+        "i;unicode-casemap"
+      ]
+    },
+    "urn:ietf:params:jmap:sieve": {
+      "implementation": "Stalwart v1.0.0"
+    },
+    "urn:ietf:params:jmap:websocket": {
+      "url": "wss://stalwart.local/jmap/ws",
+      "supportsPush": true
+    },
+    "urn:ietf:params:jmap:mail": {},
+    "urn:ietf:params:jmap:calendars": {},
+    "urn:ietf:params:jmap:calendars:parse": {},
+    "urn:ietf:params:jmap:contacts": {},
+    "urn:ietf:params:jmap:contacts:parse": {},
+    "urn:ietf:params:jmap:filenode": {},
+    "urn:ietf:params:jmap:principals": {},
+    "urn:ietf:params:jmap:principals:availability": {},
+    "urn:ietf:params:jmap:submission": {},
+    "urn:ietf:params:jmap:vacationresponse": {},
+    "urn:ietf:params:jmap:blob": {},
+    "urn:ietf:params:jmap:quota": {},
+    "urn:ietf:params:jmap:emailpush": {}
+  },
+  "accounts": {
+    "d333333": {
+      "name": "admin",
+      "isPersonal": true,
+      "isReadOnly": false,
+      "accountCapabilities": {
+        "urn:ietf:params:jmap:mail": {
+          "maxMailboxesPerEmail": null,
+          "maxMailboxDepth": 10,
+          "maxSizeMailboxName": 255,
+          "maxSizeAttachmentsPerEmail": 50000000,
+          "emailQuerySortOptions": [
+            "receivedAt",
+            "size",
+            "from",
+            "to",
+            "subject",
+            "sentAt",
+            "hasKeyword",
+            "allInThreadHaveKeyword",
+            "someInThreadHaveKeyword"
+          ],
+          "mayCreateTopLevelMailbox": true
+        },
+        "urn:ietf:params:jmap:submission": {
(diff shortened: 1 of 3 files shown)


aatchison:
Closing as superseded by #1211, and because the approach here is unsound.

#1211 already covers everything this PR attempted — it parameterises `verify_ssl`, removes the `d_get_session.json` write, adds timeouts and `allow_redirects=False`, and handles the origin problem by failing closed:

```python
if self._origin(session.api_url) != self._origin(self.base_url):
    raise ValueError('JMAP apiUrl origin does not match configured base URL')
```

That is the better design. Raising surfaces the server-side misconfiguration instead of masking it, and there is no URL construction to subvert — whereas the `urljoin` rewrite here **did not actually pin the origin**:

```
advertised //attacker.tld/jmap  ->  https://attacker.tld/jmap    (host replaced)
```

`urljoin` treats a `//`-prefixed path as a network-path reference. Reachable without an attacker, too: a trailing slash on Stalwart's configured public URL produces `//jmap/` -> `https://jmap/`, which inside EKS resolves to any Service named `jmap` in the pod's namespace.

This PR also accidentally committed `d_get_session.json` — the very artifact its description said it was leaving alone — because its tests are the first to call the real `get_session()`.

#1216 stays open; #1211 addresses most of it. Two items from the review there are worth carrying over, noted on that PR.

PR #1219, opened 2026-08-20
fix(jmap): review follow-ups for the v0.16 JMAP work

aatchison:
Stacked on #1211. Fixes the issues a review of that branch turned up, either by merging this into it or cherry-picking. Rationale for each is in the commit messages.

| # | Fix | Why it blocks |
|---|---|---|
| 1 | `get_session()` follows redirects again | Stalwart v0.16 **always** 307s `/.well-known/jmap` with an empty body, so `.json()` raises an uncaught `JSONDecodeError`. JMAP is dead at client construction. Measured on live v0.16.16. |
| 2 | `JMapOriginMismatchError(StalwartError)` replaces bare `ValueError` | The bare error escapes `fix_archives_folder` (global middleware) and `core.views.home` (`re_path(r"^.*$")`), causing **site-wide 500s** that never self-heal. |
| 3 | `verify_ssl` wired to `VERIFY_PRIVATE_LINK_SSL` | Parameter added but never connected; new `True` default means `SSLError` on every tb-dev call, fixable only by code deploy. |
| 4 | Alias delete scoped to `ALLOWED_EMAIL_DOMAINS` | **Data loss.** The new typed branch activates a dead delete path that removes custom-domain aliases from Stalwart while the local row survives. |
| 5 | `tiny_jmap_client` origin-pinned | Same bug, live user path, carries the **user's** access token. |
| 6 | `AppPassword.description` / `.permissions` optional | One externally-created credential aborts the whole list, locking the user out of setting an app password. |
| 7 | `signature.get("@type") or ""` | Pydantic emits the key as `None`, so `in None` is a `TypeError`, masking the intended error. |
| 8 | Quota `is not None` check | `"quotas": {}` parses truthy, then `null`, then `NaN%` in the UI. |
| 9 | Origin check made exercisable | The test double bypassed the check (~17 tests ran unpinned) and disagreed with its own fixture; dev `.env` pointed at a different origin than Stalwart advertises, so it failed closed locally. |

`request()` keeps `allow_redirects=False`. That's correct and deliberately untouched: a 307/308 there replays the POST body (DKIM keys, app passwords) at the new host.

Not changed: `STALWART_ADMIN_API_USE_JMAP` stays `false`. Flipping the default is a rollout decision.

A follow-up commit trims overly verbose comments per review feedback. No functional change.

Refs #1216.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -100,7 +100,12 @@ STALWART_DKIM_STAGE_MANAGEMENT_ENABLED=false
 
 # Use the v0.16 jmap api instead of the pre-v0.16 rest based api
 STALWART_ADMIN_API_USE_JMAP=false
-STALWART_JMAP_API_URL=http://stalwart_new:8080
+# Must match the origin Stalwart ADVERTISES as apiUrl, or the client fails closed with
+# JMapOriginMismatchError. The dev container advertises https://stalwart_new (port 443,
+# published as 8443 on the host); reaching it on :8080 is a different origin.
+STALWART_JMAP_API_URL=https://stalwart_new
+# Dev Stalwart serves a self-signed cert, and verify_ssl now comes from this setting.
+VERIFY_PRIVATE_LINK_SSL=false
 STALWART_JMAP_API_AUTH_USER=admin
 STALWART_JMAP_API_AUTH_SECRET=admin
 STALWART_JMAP_API_AUTH_METHOD=basic
--- a/.env.test
+++ b/.env.test
@@ -89,7 +89,12 @@ STALWART_API_AUTH_METHOD=basic
 
 # Use the v0.16 jmap api instead of the pre-v0.16 rest based api
 STALWART_ADMIN_API_USE_JMAP=false
-STALWART_JMAP_API_URL=http://stalwart_new:8080
+# Must match the origin Stalwart ADVERTISES as apiUrl, or the client fails closed with
+# JMapOriginMismatchError. The dev container advertises https://stalwart_new (port 443,
+# published as 8443 on the host); reaching it on :8080 is a different origin.
+STALWART_JMAP_API_URL=https://stalwart_new
+# Dev Stalwart serves a self-signed cert, and verify_ssl now comes from this setting.
+VERIFY_PRIVATE_LINK_SSL=false
 STALWART_JMAP_API_AUTH_USER=admin
 STALWART_JMAP_API_AUTH_SECRET=admin
 STALWART_JMAP_API_AUTH_METHOD=basic
--- a/src/thunderbird_accounts/mail/clients/jmap_client.py
+++ b/src/thunderbird_accounts/mail/clients/jmap_client.py
@@ -7,7 +7,7 @@ from urllib.parse import urlsplit
 import requests
 from pydantic import ValidationError
 
-from thunderbird_accounts.mail.exceptions import InvalidJMapResponseError
+from thunderbird_accounts.mail.exceptions import InvalidJMapResponseError, JMapOriginMismatchError
 from thunderbird_accounts.mail.types.jmap import Invocation, JMapRequest, JMapResponse, SessionResource
 
 
@@ -68,7 +68,16 @@ class JMAPClient:
                 'Content-Type': 'application/json',
                 'Authorization': self._authorization_value(),
             },
-            allow_redirects=False,
+            # RFC 8620 2.2 defines the session resource as fetched "following any redirects",
+            # and Stalwart v0.16 ALWAYS 307s /.well-known/jmap -> /jmap/session with an empty
+            # body. With redirects disabled, raise_for_status() is a no-op on 3xx and the empty
+            # body reaches .json(), raising an untyped JSONDecodeError before any of this class's
+            # error handling runs. Redirects stay enabled here; the destination is still pinned,
+            # because the origin check below runs against the resulting session.
+            #
+            # This does NOT relax request(), where allow_redirects=False is correct: a 307/308
+            # replays the POST body (DKIM private keys, app passwords) at the new host.
+            allow_redirects=True,
             verify=self.verify_ssl,
             timeout=self.timeout,
         )
@@ -78,7 +87,7 @@ class JMAPClient:
         except ValidationError as ex:
             raise InvalidJMapResponseError(ex) from ex
         if self._origin(session.api_url) != self._origin(self.base_url):
-            raise ValueError('JMAP apiUrl origin does not match configured base URL')
+            raise JMapOriginMismatchError(session.api_url, self.base_url)
         self.session = session
         if not self.session:
             raise RuntimeError('Failed to get session')
--- a/src/thunderbird_accounts/mail/clients/jmap_client.py
+++ b/src/thunderbird_accounts/mail/clients/jmap_client.py
@@ -86,11 +86,18 @@ class JMAPClient:
             session = SessionResource.model_validate(r.json())
         except ValidationError as ex:
             raise InvalidJMapResponseError(ex) from ex
+        return self._accept_session(session)
+
+    def _accept_session(self, session: SessionResource) -> SessionResource:
+        """Validate and store a fetched session.
+
+        Split out of get_session() so the origin check has exactly one home: any test double or
+        future cache that populates a session must come through here, rather than reimplementing
+        the assignment and silently bypassing the pin.
+        """
         if self._origin(session.api_url) != self._origin(self.base_url):
             raise JMapOriginMismatchError(session.api_url, self.base_url)
         self.session = session
-        if not self.session:
-            raise RuntimeError('Failed to get session')
         self.api_url = session.api_url
         return session
 
--- a/src/thunderbird_accounts/mail/clients/mail_client_jmap.py
+++ b/src/thunderbird_accounts/mail/clients/mail_client_jmap.py
@@ -58,7 +58,16 @@ class BaseJMAP(ABC):
     def _get_user_client(
         self, username, user_token, auth_type: JMAPClient.AUTH_TYPES = JMAPClient.AUTH_TYPES.BEARER
     ) -> JMAPClient:
-        client = JMAPClient(settings.STALWART_JMAP_API_URL, username, user_token, auth_type)
+        client = JMAPClient(
+            settings.STALWART_JMAP_API_URL,
+            username,
+            user_token,
+            auth_type,
+            # Same knob the legacy client uses for this backend. tb-dev reaches Stalwart at an
+            # in-cluster Service DNS name while Stalwart serves the PUBLIC mail-host cert, so
+            # verification must be switchable per environment rather than hardcoded either way.
+            verify_ssl=settings.VERIFY_PRIVATE_LINK_SSL,
+        )
         # Make sure we retrieve the session
         client.get_session()
         return client
--- a/src/thunderbird_accounts/mail/dkim.py
+++ b/src/thunderbird_accounts/mail/dkim.py
@@ -143,7 +143,9 @@ def _dkim_key_type_and_public_key_value(signature: dict) -> tuple[str, str]:
             try:
                 public_key_obj = serialization.load_der_public_key(public_key_bytes)
             except (ValueError, UnsupportedAlgorithm) as ex:
-                signature_type = signature.get('@type', '')
+                # Pydantic emits the key with a None value when unset, so a bare default is
+                # not enough -- `'Ed25519' in None` is a TypeError, not the intended RuntimeError.
+                signature_type = signature.get('@type') or ''
                 if 'Ed25519' in signature_type and len(public_key_bytes) == 32:
                     public_key_obj = ed25519.Ed25519PublicKey.from_public_bytes(public_key_bytes)
                 else:
--- a/src/thunderbird_accounts/mail/exceptions.py
+++ b/src/thunderbird_accounts/mail/exceptions.py
@@ -181,6 +181,22 @@ class EmailNotValidError(RuntimeError):
         return f'EmailNotValidError: {self.email}, {self.error_message}'
 
 
+class JMapOriginMismatchError(StalwartError):
+    """The JMAP session advertised an apiUrl on a different origin than the configured base URL.
+
+    Subclasses StalwartError (and so RuntimeError) deliberately: callers that already degrade
+    gracefully on Stalwart being unreachable should treat this the same way, rather than 500.
+    """
+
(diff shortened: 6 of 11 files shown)


src/thunderbird_accounts/mail/clients/jmap_client.py
```diff
@@ -68,7 +68,16 @@ def get_session(self) -> SessionResource:
...
-            allow_redirects=False,
+            # RFC 8620 2.2 defines the session resource as fetched "following any redirects",
+            # and Stalwart v0.16 ALWAYS 307s /.well-known/jmap -> /jmap/session with an empty
+            # body. With redirects disabled, raise_for_status() is a no-op on 3xx and the empty
+            # body reaches .json(), raising an untyped JSONDecodeError before any of this class's
+            # error handling runs. Redirects stay enabled here; the destination is still pinned,
+            # because the origin check below runs against the resulting session.
+            #
+            # This does NOT relax request(), where allow_redirects=False is correct: a 307/308
+            # replays the POST body (DKIM private keys, app passwords) at the new host.
```
markstos, line 79: This commentary seems useful in a PR, but feels overly verbose to explain in this much detail why we've set a boolean in the code.

src/thunderbird_accounts/mail/clients/jmap_client.py
```diff
@@ -77,11 +86,18 @@ def get_session(self) -> SessionResource:
...
             raise InvalidJMapResponseError(ex) from ex
+        return self._accept_session(session)
+
+    def _accept_session(self, session: SessionResource) -> SessionResource:
+        """Validate and store a fetched session.
+
+        Split out of get_session() so the origin check has exactly one home: any test double or
+        future cache that populates a session must come through here, rather than reimplementing
+        the assignment and silently bypassing the pin.
+        """
```
markstos, line 97: It seems sufficient to leave the comment at "Validate and store a fetched session".

src/thunderbird_accounts/mail/clients/mail_client_jmap.py
```diff
@@ -58,7 +58,16 @@ class BaseJMAP(ABC):
...
-        client = JMAPClient(settings.STALWART_JMAP_API_URL, username, user_token, auth_type)
+        client = JMAPClient(
+            settings.STALWART_JMAP_API_URL,
+            username,
+            user_token,
+            auth_type,
+            # Same knob the legacy client uses for this backend. tb-dev reaches Stalwart at an
+            # in-cluster Service DNS name while Stalwart serves the PUBLIC mail-host cert, so
+            # verification must be switchable per environment rather than hardcoded either way.
+            verify_ssl=settings.VERIFY_PRIVATE_LINK_SSL,
```
markstos, line 69: This comment can be dropped. We reasonably expect that an option is tunable-per-env if we set it via "settings" and don't need to spell that out in a comment.

src/thunderbird_accounts/mail/tests/test_clients/test_jmap.py
```diff
@@ -148,12 +152,10 @@ def get_session(self) -> SessionResource:
...
-        session = SessionResource(**fixture_data)
-        self.session = session
-        if not self.session:
-            raise RuntimeError('Failed to get session')
-        self.api_url = session.api_url
-        return session
+        # Stub only the HTTP fetch; everything after it, including the origin check, must run
+        # exactly as it does in production. Previously this reimplemented the assignment, so the
+        # ~17 tests built on this double exercised UNPINNED behaviour and CI carried no signal.
+        return self._accept_session(SessionResource(**fixture_data))
```
markstos, line 158: Our comments rarely need to explain what the code *previously* did, which adds unnecessary layer of complexity. This comment can be dropped.

src/thunderbird_accounts/mail/tests/test_clients/test_jmap.py
```diff
@@ -901,3 +903,22 @@ def test_get_dkim_dns_records_returns_legacy_dictionary_shape(self):
             ],
         )
         self.mail_client._get_dkim_dns_records.assert_called_once_with('example.com')
+
+
+class TestOriginMismatchIsRecoverable(SimpleTestCase):
+    """The origin mismatch must be catchable by the handlers that already degrade gracefully.
```
markstos, line 909: The top-line comment here is sufficient.

src/thunderbird_accounts/mail/types/stalwart.py
```diff
@@ -218,11 +218,15 @@ class SecondaryCredential(Credential):
...
+    # Optional on purpose: these are required only by OUR convention, not by Stalwart. A
+    # credential created outside Accounts (admin CLI, another client) can omit either, and a
+    # required field aborts the whole get_app_passwords list -- locking the user out of setting
+    # a new app password because of one unrelated credential.
+    description: Optional[str] = None
     secret: Optional[str] = None  # Server set
     created_at: Optional[datetime] = None
     expires_at: Optional[datetime] = None
-    permissions: StalwartType
+    permissions: Optional[StalwartType] = None
```
markstos, line 229: This is the code for *accounts*. Why should we be loosening validation to support other CLI tools or other clients?

src/thunderbird_accounts/mail/dkim.py
```diff
@@ -143,7 +143,9 @@ def _dkim_key_type_and_public_key_value(signature: dict) -> tuple[str, str]:
             try:
                 public_key_obj = serialization.load_der_public_key(public_key_bytes)
             except (ValueError, UnsupportedAlgorithm) as ex:
-                signature_type = signature.get('@type', '')
+                # Pydantic emits the key with a None value when unset, so a bare default is
+                # not enough -- `'Ed25519' in None` is a TypeError, not the intended RuntimeError.
+                signature_type = signature.get('@type') or ''
```
markstos, line 148: No need for a comment here. If anyone tried to change the syntax back-- they would quickly find the pydantic  error

src/thunderbird_accounts/mail/exceptions.py
```diff
@@ -181,6 +181,22 @@ def __str__(self):
         return f'EmailNotValidError: {self.email}, {self.error_message}'
 
 
+class JMapOriginMismatchError(StalwartError):
+    """The JMAP session advertised an apiUrl on a different origin than the configured base URL.
+
+    Subclasses StalwartError (and so RuntimeError) deliberately: callers that already degrade
+    gracefully on Stalwart being unreachable should treat this the same way, rather than 500.
```
markstos, line 188: Top-line comment is sufficient.

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -367,8 +367,17 @@ def create_stalwart_account(
...
             # Diff of new emails
             new_emails = set(emails) - set(stalwart_emails)
-            # Diff of the old emails
-            old_emails = set(stalwart_emails) - set(emails)
+            # Diff of the old emails. Scope the DELETE to addresses in ALLOWED_EMAIL_DOMAINS --
+            # the only ones this task owns. `emails` is just the canonical set, so an unscoped
+            # diff also sweeps up custom-domain aliases the user added via the mail UI, removing
+            # them from Stalwart while their local Email row survives. Previously unreachable
+            # (get_account always returned a dict, so stalwart_emails was always empty); the
+            # typed branch above makes it live.
```
markstos, line 375: The original simpler "Diff of the old emails" comment remains sufficient here.

src/thunderbird_accounts/mail/tiny_jmap_client.py
```diff
@@ -43,6 +46,10 @@ def get_session(self):
         )
         r.raise_for_status()
         self.session = session = r.json()
+        # Same pin as JMAPClient: this client carries the USER's access token, and apiUrl is
+        # server-supplied, so an advertised URL on another origin would send that token off-host.
```
markstos, line 50: Simplify comment like "Validate API URL matches our hostname".

src/thunderbird_accounts/subscription/views.py
```diff
@@ -277,7 +277,10 @@ def get_subscription_plan_info(request: Request, paddle: Client):
         stalwart_client = MailClient()
         account = stalwart_client.get_account(request.user.stalwart_primary_email)
         if isinstance(account, stalwart.Account):
-            quota = account.quotas.max_disk_quota if account.quotas else 0
+            # Stalwart sends "quotas": {} for an unlimited account, which parses to a TRUTHY
+            # StorageQuota() whose max_disk_quota is None -> mailStorage: null -> NaN% in the UI.
+            # Mirror the `is not None` check used for used_disk_quota below.
```
markstos, line 282: Simplify comment:  " # Treat Stalwart's empty quotas object as unlimited to avoid NaN% in the UI."

.env.example
```diff
@@ -100,7 +100,12 @@ STALWART_DKIM_STAGE_MANAGEMENT_ENABLED=false
 
 # Use the v0.16 jmap api instead of the pre-v0.16 rest based api
 STALWART_ADMIN_API_USE_JMAP=false
-STALWART_JMAP_API_URL=http://stalwart_new:8080
+# Must match the origin Stalwart ADVERTISES as apiUrl, or the client fails closed with
+# JMapOriginMismatchError. The dev container advertises https://stalwart_new (port 443,
+# published as 8443 on the host); reaching it on :8080 is a different origin.
+STALWART_JMAP_API_URL=https://stalwart_new
```
markstos, line 106: Simplify comment as "Must match the origin Stalwart advertises as apiUrl,"

.env.test
```diff
@@ -89,7 +89,12 @@ STALWART_API_AUTH_METHOD=basic
 
 # Use the v0.16 jmap api instead of the pre-v0.16 rest based api
 STALWART_ADMIN_API_USE_JMAP=false
-STALWART_JMAP_API_URL=http://stalwart_new:8080
+# Must match the origin Stalwart ADVERTISES as apiUrl, or the client fails closed with
+# JMapOriginMismatchError. The dev container advertises https://stalwart_new (port 443,
+# published as 8443 on the host); reaching it on :8080 is a different origin.
+STALWART_JMAP_API_URL=https://stalwart_new
```
markstos, line 95: Simplify comment as "Must match the origin Stalwart advertises as apiUrl,"

src/thunderbird_accounts/mail/types/stalwart.py
markstos: The PR description says "One externally-created credential aborts the whole list → user locked out of setting an app password."  But no evidence is offered of any case where credentials are externally created. Is that even possible? Should it be? Unless there's evidence we actually have this case, I recommend removing this weaker validation from the scope.

markstos: Regarding this one, AI analysis found that by Stalwart 0.16.19, these two fields have defaults set, so that even if /they/ were optional, it would still not be a problem.  It also recommended to retain them unless we have a demonstrated reason not to.

src/thunderbird_accounts/mail/clients/jmap_client.py
```diff
@@ -77,11 +86,18 @@ def get_session(self) -> SessionResource:
...
         if self._origin(session.api_url) != self._origin(self.base_url):
-            raise ValueError('JMAP apiUrl origin does not match configured base URL')
+            raise JMapOriginMismatchError(session.api_url, self.base_url)
```
markstos, line 99: This is fix is incomplete.

` JMapOriginMismatchError` is a `RuntimeError`, but `core.views.home`
     (https://github.com/thunderbird/thunderbird-accounts/blob/357bfba77d21f9c9de4e6f45d82096df5417a223/src/thunderbird_accounts/core/views.py#L80-L117)
     catches only `AccountNotFoundError` and `requests.ConnectionError`. So there's still a gap for problems when `MailClient()`  performs eager session discovery.

The health check has the same problem: it catches only `HTTPError`, so other error types can turn `/health` into a 500 rather than reporting Stalwart unhealthy.

Recommended fix: catch the mail-client failure family at these boundaries—at least StalwartError,
InvalidJMapResponseError, and requests.RequestException—and add request-level tests proving the home page degrades and health returns an unhealthy result.

markstos (changes requested):
To review this, I used AI to first review the base PR, then to check if this PR addressed all the concerns there, also cross referencing the lastest Stalwart release of v0.16.19 to see if maybe fixes on our side have already been fixed upstream.  I also took a pass to read through the diff here, mostly noting that's lots of comments are overly verbose.

## Other recommended fixes within scope of 0.16 regressions.

Besides my in-line comments, some more feedback here:

1. Handle app-password replacement at quota.

   Replacement creates the new credential before deleting the old one. Stalwart checks `maxAppPasswords` before creation, so replacement fails when the account is already at quota. Either guarantee spare quota operationally or explicitly handle `overQuota`; the current flow should not be described as universally safe.

2. Finish the v0.16 admin-list mapping.

   [`list_principals()`](https://github.com/thunderbird/thunderbird-accounts/blob/357bfba77d21f9c9de4e6f45d82096df5417a223/src/thunderbird_accounts/mail/clients/mail_client_jmap.py#L686-L718) returns at most 100 accounts, while the view still reads legacy fields such as `type`, `emails`, `quota`, and `usedQuota`. It no longer crashes, but much of the displayed data remains empty under v0.16.


## Related issues discoveded in flight. Can be follow-up PRs:

1. invalid origins still raise the original bare `ValueError`.

   [`_origin()`](https://github.com/thunderbird/thunderbird-accounts/blob/357bfba77d21f9c9de4e6f45d82096df5417a223/src/thunderbird_accounts/mail/clients/jmap_client.py#L53-L59) raises `ValueError` for a missing/invalid scheme or hostname, and accessing `parsed.port` can also raise `ValueError`. Only a valid-but-different origin becomes `JMapOriginMismatchError`.

   Recommended fix: normalize all advertised-origin parsing failures into a typed JMAP/Stalwart client error.

2. Normalize syntactically invalid JSON.

   Both `get_session()` and `request()` catch only Pydantic `ValidationError`; `requests.exceptions.JSONDecodeError` still escapes. Add tests where `response.json()` raises, not merely returns a structurally invalid array.

3. Remove the hosted-DKIM publish/delete race.

   Custom-domain creation queues publication before the local `Domain` write. If persistence fails, a deletion task is queued afterward, but Celery does not guarantee those tasks execute in that order. Persist first and enqueue publication with `transaction.on_commit()` so failed persistence never schedules publication.

4. If `TinyJMAPClient` remains, complete its transport hardening.

   #1219 pins its discovered `apiUrl`, but `make_jmap_call()` still follows redirects and has no timeout. Either retire it in favor of `JMAPClient` or give its authenticated POSTs the same redirect and timeout behavior.

PR #1223, opened 2026-08-24
Fix intermittent BrowserStack E2E races (Fixes #1222)

### LINKED ISSUE #1222: Intermittent Browserstack E2E race conditions
The browserstack tests intermittently fail when:

1. Invalid username validation disables submit while the test attempts to click it.
2. Dashboard tests assert password or service-popup destinations before OIDC navigation completes.

We should assert the disabled validation state and poll final navigation destinations instead of using fixed waits or immediate checks.

Sancus:
5 subsequent runs on this were successful so seems like it resolves the issue.

### DIFF
--- a/test/e2e/pages/dashboard-page.ts
+++ b/test/e2e/pages/dashboard-page.ts
@@ -115,9 +115,8 @@ export class DashboardPage {
 }
 
   async verifyPasswordChangeNavigation() {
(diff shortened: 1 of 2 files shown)


rwood-moz (approved):
Good catch, thanks for this!

PR #1226, opened 2026-08-25
Update Keycloak to 26.7.2

markstos:
## QA Log

- Docker image build passes

### DIFF
--- a/Dockerfile.keycloak
+++ b/Dockerfile.keycloak
@@ -58,7 +58,7 @@ COPY --chown=keycloak:keycloak keycloak/data/import /opt/keycloak/data/import
 COPY --from=build /app/keycloak/themes/tbpro /opt/keycloak/themes/tbpro
 
 # keycloak-config-cli: declarative reconciler for the managed tbpro realm configuration.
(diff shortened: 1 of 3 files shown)


Dockerfile.keycloak
```diff
@@ -58,7 +58,8 @@ COPY --chown=keycloak:keycloak keycloak/data/import /opt/keycloak/data/import
 COPY --from=build /app/keycloak/themes/tbpro /opt/keycloak/themes/tbpro
 
 # keycloak-config-cli: declarative reconciler for the managed tbpro realm configuration.
-# The jar is version-matched to Keycloak 26.5 and runs on the image's bundled Java 21;
+# The latest config-cli jar is built against Keycloak 26.5.5 and runs on the image's
```
ryanjjung: Wrong version in the comment here.

MelissaAutumn:
```suggestion
# The latest config-cli jar is built against Keycloak 26.7 and runs on the image's
```

Sancus: we aren't updating config-cli jar so the comment is correct (there isn't a new version of it and it doesn't need to track the keycloak version, but still useful to have this in as a reminder I think)

PR #1230, opened 2026-08-26
fix(keycloak-theme): guard the OTP, recovery-code and reset forms against double submit (#1133)

aatchison:
Pressing Enter on the OTP, recovery-code, or password-reset form submits it twice. The second POST reuses a spent `session_code`, so Keycloak restarts the flow and the user lands back at the start of sign-in. On the recovery-code form the first POST also spends a single-use code, so the retry costs another one.

All three views bind `onSubmit` to both `@submit.prevent` and `@keyup.enter`, and `onSubmit` calls the native `form.submit()`. Enter reaches it once through implicit submission and once through the bubbling keyup.

Each form now submits once and disables its button while a submission is in flight. Native `submit()` skips constraint validation, so the code forms check it explicitly; previously an empty code posted anyway.

Six other views share the dual binding. Enter on a focused button double-fires there at any field count, so the follow-up should cover all six. Tracked in #1231.

Refs #1133. Does not fix the separate `/oidc/callback/` state-replay errors.

### DIFF
--- a/keycloak/themes/tbpro/assets/vue/views/ForgotPasswordView/index.vue
+++ b/keycloak/themes/tbpro/assets/vue/views/ForgotPasswordView/index.vue
@@ -8,9 +8,6 @@ const loginUrl = window._page.currentView?.loginUrl;
 const username = window._page.currentView?.attemptedUserName;
 const resetPasswordForm = useTemplateRef('reset-password-form');
 
-// Enter can reach onSubmit twice: the form's implicit submission fires @submit.prevent and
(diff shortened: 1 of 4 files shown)


MelissaAutumn (approved):
This should work, and fixes a small oversight (form and button submission occuring at the same time) that would affect a11y. Thanks!

I'm not sure if we need the comment context though, as we use it throughout the accounts app.

aatchison:
Comments removed, ty!

PR #1234, opened 2026-08-26
fix(keycloak-theme): guard remaining forms against double submit (#1133)

### LINKED ISSUE #1231: Guard the remaining dual-binding login-theme forms against double-submit
Nine views in `keycloak/themes/tbpro/assets/vue/views/` bind `onSubmit` to both `@submit.prevent` and `@keyup.enter`, and `onSubmit` calls the native `form.submit()` with no re-entrancy guard. #1230 guards three of them (`LoginTotpView`, `ForgotPasswordView`, `LoginRecoveryAuthnCodeView`). These six still do not:

| View | Blocking fields | Has `checkValidity()` |
|---|---|---|
| `ConfigToptView` | 2 | no |
| `LoginView` | 2 | yes |
| `LogoutView` | 0 | no |
| `RegisterView` | 7 | yes |
| `UpdatePasswordView` | 2 | no |
| `VerifyEmailView` | 0 | no |

## The double-fire

The submit buttons render as `type="button"` with `@click="onSubmit"` and sit inside the form, while `@keyup.enter` is bound on the form element. Pressing Enter while the button has focus fires `click` and then a bubbling `keyup`, so `onSubmit` runs twice and two identical POSTs go out. This happens regardless of field count, including the two views with no text fields at all.

The second POST reuses a spent `session_code`, so Keycloak restarts the flow (`error="expired_code"`, `restart_after_timeout="true"`) and drops the user at the start of sign-in.

## Fix

Follow the pattern in #1230: an `isSubmitting` ref guarding `onSubmit`, `:disabled="isSubmitting"` on the submit button, and a `checkValidity()` gate for the views that lack one, since native `submit()` skips constraint validation.

## Severity

Lower than the three already guarded. Those have a single blocking field, so Enter pressed inside the field also triggers HTML implicit submission and double-fires, which is the path users actually hit. These six need Enter on a focused button, which is less common. Worth doing for consistency and to close out the pattern.

Refs #1133.

WilliamK112:
## What changed?

- Added an `isSubmitting` re-entrancy guard to `ConfigToptView`, `LoginView`, `LogoutView`, `RegisterView`, `UpdatePasswordView`, and `VerifyEmailView`.
- Disabled each submit button after the first valid submission and added explicit `checkValidity()` gates before native `form.submit()`.
- Kept `ConfigToptView` on step 2 when validation fails instead of advancing to an empty step.
- Changed the shared MFA sign-in helper to focus the login submit button and press Enter, exercising the exact click-plus-keyup path that previously submitted twice.

AI disclosure: this patch and its validation were performed by Codex. The implementation was constrained to the six views and regression path specified in #1231 and follows the pattern established in #1230.

### Verification

- `npm run build-theme`
- `npm run build`
- `npm run test:frontend` (3 files, 12 tests passed)
- `npx eslint` on the changed files: `LoginView` passes; the remaining diagnostics are pre-existing in the unmodified baseline (legacy Keycloak views lack script `lang` attributes, and the E2E package resolves Playwright from its nested install).
- The repository-hosted E2E workflow will exercise the full Keycloak/Django stack with its required secrets.

## Why?

These forms bind `onSubmit` to both `@submit.prevent` and `@keyup.enter`. Pressing Enter while the submit button is focused fires its click and then a bubbling keyup, issuing two identical POSTs. The second POST reuses a spent Keycloak `session_code` and can restart the sign-in flow.

## Limitations and Notes

#1230 is not required for this PR: it covers three different views, while this PR closes the remaining six. The changes are intentionally disjoint and can merge in either order.

No screenshots are included because this is a submission-control behavior change with no intended visual difference beyond the button disabling while navigation is in flight.

## Applicable Issues

Closes #1231
Refs #1133

## Screenshots

Not applicable; no persistent visual change.

### DIFF
--- a/keycloak/themes/tbpro/assets/vue/views/ConfigToptView/index.vue
+++ b/keycloak/themes/tbpro/assets/vue/views/ConfigToptView/index.vue
@@ -10,6 +10,8 @@ import { computed, ref, useTemplateRef } from "vue";
 import { i18n } from '@kc/composables/i18n.js';
 import CancelForm from '@kc/vue/components/CancelForm.vue';
 const isManualMode = ref(false);
+// @submit and @keyup.enter can both fire for one Enter keypress; keep the POST single-shot (#1231).
+const isSubmitting = ref(false);
 const formAction = window._page.currentView?.formAction;
 const settingsForm = useTemplateRef('settings-form');
 const cancelForm = useTemplateRef('cancel-form');
@@ -67,7 +69,10 @@ const otpManualUrl = computed(() => {
 });
 
 const onSubmit = () => {
-  settingsForm?.value?.submit();
+  if (isSubmitting.value) return;
+  if (!settingsForm.value?.checkValidity()) return;
+  isSubmitting.value = true;
+  settingsForm.value.submit();
 };
 const onCancel = () => {
   cancelForm?.value?.cancel();
@@ -94,10 +99,11 @@ const copyLink = async () => {
 };
 
 const onNext = () => {
-  totpStep.value += 1;
-  if (totpStep.value >= 3) {
+  if (totpStep.value >= 2) {
     onSubmit();
+    return;
   }
+  totpStep.value += 1;
 };
 const onPrevious = () => {
   totpStep.value -= 1;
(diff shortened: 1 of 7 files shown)


WilliamK112:
Understood — sorry for adding to the noise here. I've closed this PR and won't pursue the remaining edge case further.

PR #1235, opened 2026-08-26
Pin Pulumi CLI version in CI

markstos:
## What's changing

This PR improves reproducibility and consistency of CI runs by using a known-good version
of the Pulumi CLI instead of the "latest" version.

The version chosen was recently successfully used for both staging and production.

We'll need to periodically bump the version to new known-good versions.

## Why?

This change is recommended on principle, not due to a problem observed so far with using latest.

## QA Log

 - Analyzed recent deploys to find a known good version.
 - The syntax to install a specific version is documented here: https://www.pulumi.com/docs/install/

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -150,7 +150,7 @@ jobs:
         run: |
           cd $PULUMI_DIR
           source ./venv/bin/activate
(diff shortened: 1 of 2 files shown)


markstos:
According to @aatchison this approach of downloading Pulumi during CI is going away soon, so I'll close this prevent churn.

Issue #1229, opened 2026-08-25
keycloak requires client using the token to also be in its `aud` claim


MelissaAutumn:
New keycloak requires the client using the token to also be in its `aud` claim in order to introspect it. Makes sense but it requires some work...

## Services that need `stalwart` mapped
* thunderbird accounts (to id token)
* desktop
* mobile
* stormbox
* fairemail

## Other instances
* appointment frontend needs appointment backend mapped (as the backend uses the token)

Sancus:
I think the complete list is:

- desktop : stalwart
- desktop : thunderbird-send-backend
- ext-fairemail : stalwart
- mobile-android-k9mail : stalwart
- mobile-android-thunderbird : stalwart
- thunderbird-accounts : stalwart
- thunderbird-accounts : thunderbird-accounts
- thunderbird-appointment-frontend : thunderbird-appointment-backend
- thunderbird-stormbox : stalwart
- mobile-ios-thunderbird : stalwart

stalwart : stalwart  (already exists)

I thought there must be one for Send but it seems like Send doesn't actually use the `thunderbird-send-frontend` client? It just uses desktop.

kewisch:
Yeah, there was some back and forth on Send OAuth client usage, we eventually just picked `desktop` so we could use one token and not log in multiple times. This was before the OAuth browser changes.

MelissaAutumn:
I'm starting with stage, I'll create some client scopes labelled `include_$client_aud` and just include them as defaults as mentioned above. Once that's done I'll remove the work-around env in the container and deploy it to stage.

PR #1236, opened 2026-08-26
Disable the aud check work-around for keycloak stage (#1229)

MelissaAutumn:
Part of #1229 

I've made some changes to how we apply keycloak cli configs, and while it's not perfect (they race to see which one runs first lol, thankfully we have that db lock lol) it does work on my local config and is no longer click-ops.

Since keycloak doesn't have a PATCH update route, we must declare every single value in a field we update. This is why these are distinct scopes rather scopes linked via yaml merge keys. We normally don't use the dedicated scope for anything so it's okay if they're entirely replaced by the ones we declare here.

## Important Note

This will require a dev db wipe as I renamed tb-accounts to thunderbird-accounts to match stage/prod. Any clients not in your keycloak realm will be created as empty, so that's good at least. 

## Related Docs

 * [keycloak-config-cli](https://adorsys.github.io/keycloak-config-cli/) Reconciles YAML config files against the database. It is idempotent.


MelissaAutumn:
Ah we probably want this in the mfa setup yml eh.

keycloak/config-cli/tbpro-mfa-stepup.yaml
```diff
@@ -24,22 +24,6 @@ attributes:
 # Bind the realm's browser login to the step-up flow defined below.
 browserFlow: browser-mfa-stepup
 
-# Keycloak 26.6.2+ requires the client performing token introspection to be in the
```
Sancus, line 27: You remove this but I don't think you restored it in the other file? It's causing one of the failing e2e tests

Sancus (changes requested):

Sancus:
The second failing test is the one that tests the aud check is disabled and it should just be removed.

keycloak/config-cli/tbpro-mfa-stepup.yaml
MelissaAutumn: Stalwart -> stalwart should resolve via the audience resolver mapper on roles scope...i will do some more digging.

But maybe its not being mapped correctly by default.

MelissaAutumn: Yeah we shouldn't need this as the roles default scope (built-in to keycloak) includes their own client id as the aud.

Here's a sample of the access token for stalwart on my local machine.
```
{
  "exp": 1788200101,
  "iat": 1788199801,
  "jti": "onrtna:8e16b5f9-10c9-7067-2a6a-81752e99e791",
  "iss": "http://keycloak:8999/realms/tbpro",
  "aud": "account",
  "sub": "5f75218f-1cb0-49a5-bd1c-e38c3b32dbd2",
  "typ": "Bearer",
  "azp": "stalwart",
  "sid": "pKK3roc16on_tI8q6mJ7NcF3",
  "acr": "1",
  "allowed-origins": [
    "*"
  ],
  "realm_access": {
    "roles": [
      "offline_access",
      "uma_authorization",
      "default-roles-stalwart"
    ]
  },
  "resource_access": {
    "account": {
      "roles": [
        "manage-account",
        "manage-account-links",
        "view-profile"
      ]
    }
  },
  "scope": "openid profile email",
  "zoneinfo": "America/Vancouver",
  "email_verified": true,
  "is_services_admin": "yes",
  "name": "Admin Example",
  "preferred_username": "admin@example.org",
  "given_name": "Admin",
  "locale": "en",
  "family_name": "Example",
  "email": "admin@example.com"
}
```

MelissaAutumn: Okay you're right, but it's seemingly a bug on keycloaks end. (Don't at me, I haven't looked at the oauth spec here lol.) 

The aud returns as `account` (keycloak's account management client) because we're logging in as a user for this request. I think Keycloak should be also checking the azp for this aud check but that's not happening. So I'll re-add the stalwart->stalwart mapping. I think that would be the only time we use user sso -> client.

MelissaAutumn:
...ignore that last commit.

scripts/apply-cli-config.sh
markstos, line 1: After this rename, several comments need to be updated to match. Search the code base for `apply-mfa-config.sh` to find them all.

keycloak/config-cli/tbpro-aud-mapper.yaml
markstos, line 1: How did you determine this was the complete list? Did export all the things needed from keycloak in production to make sure this action doesn't cause anything we care about to be deleted?

keycloak/config-cli/tbpro-aud-mapper.yaml
```diff
@@ -0,0 +1,81 @@
...
+# A rough map looks like:
+# - stalwart : stalwart
+# - desktop : stalwart
+# - desktop : thunderbird-send-backend
+# - ext-fairemail : stalwart
+# - mobile-android-k9mail : stalwart
+# - mobile-android-thunderbird : stalwart
+# - mobile-ios-thunderbird : stalwart
+# - thunderbird-accounts : stalwart
+# - thunderbird-accounts : thunderbird-accounts
```
markstos, line 17: The comment expresses a mapping of `thunderbird-accounts` to `thunderbird-accounts`, but the actual code doesn't. Shouldn't these agree?

scripts/entry-keycloak.sh
```diff
@@ -22,6 +22,8 @@ KC_PID=$!
 # Forward termination so container stop stays graceful.
 trap 'kill -TERM "$KC_PID" 2>/dev/null' TERM INT
 
-/bin/bash "$SCRIPT_DIR/apply-mfa-config.sh" &
+/bin/bash "$SCRIPT_DIR/apply-cli-config.sh" "tbpro-mfa-stepup" &
+/bin/bash "$SCRIPT_DIR/apply-cli-config.sh" "tbpro-aud-mapper" &
```
markstos: I expect these are running the background so we can start up faster, but:

- Do we still need to continually apply the MFA stepup during keycloak boot, or was it more like a one-time migration that we can delete?
- This design allows the background task to fail but Keycloak still boots. If it just adds a few seconds to the boot time, could we foreground it instead to have clear error detection?

test/e2e/tests/token-introspection.spec.ts
markstos, line 1: Could we replace this with a test for the new behavior instead?  Something that proves that a thunderbird-accounts user token contains the right audience and can be introspected with strict checking enabled.

keycloak/config-cli/tbpro-aud-mapper.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - clientId: desktop
+    protocolMappers:
+      - *scope_stalwart_audience
+      - name: thunderbird send backend audience
+        protocol: openid-connect
+        protocolMapper: oidc-audience-mapper
+        consentRequired: false
+        config:
+          included.client.audience: thunderbird-send-backend
+          id.token.claim: "true"
```
markstos: I believe `id.token.claim` should be false for the Stalwart audience mapper. This adds the the claim to the ID token, which is only useful if Stalwart consumes. An analysis of the Stalwart source code confirms it consumes /access/ tokens not /id/ tokens. This is true across Stalwart 0.15 and 0.16. In 0.16, JWTs are used more, but it's still consuming access tokens, not id tokens.

markstos (changes requested):
Thanks for this work. See my in-line comments and questions. Also: 

- There's a stale reference to apply-mfa-config.sh in `.github/workflows/validate.yml`.

scripts/entry-keycloak.sh
Sancus: It is intentional that all CLI configs run every time, to prevent drift that could break authentication flows. They should be pretty close to no-ops if they're already set correctly, though.

keycloak/config-cli/tbpro-aud-mapper.yaml
MelissaAutumn: Thanks for spotting this. It's seemingly a problem on Keycloak's end. We'll need to map every client id to the audience of their own tokens...

scripts/entry-keycloak.sh
markstos: If they are generally close to no-ops, could we make them sync rather than async to catch errors if they fail?

MelissaAutumn: The config change was mostly a hack in order to keep the mfa script from becoming the size of the sun. I'm okay with sync, but there's currently no relying information between them afaik.

keycloak/config-cli/tbpro-aud-mapper.yaml
MelissaAutumn: Just kidding, aud is only needed for the client communicating the introspect to the rp iirc. So I think tb-accounts was the only missing link here. It works on my machine now.

keycloak/config-cli/tbpro-aud-mapper.yaml
MelissaAutumn: We thankfully don't use the dedicated client scopes, so nothing will be overwritten. Originally I tried creating individual client scopes and applying them. Creating went fine, but applying required the full list of scopes for each client (as it's a PUT not PATCH...)

I think we'll need to move most of keycloak's existing config to be managed by something like this in the future, but I don't have the mental will power to do it all at once :sweat_smile:.

I'm pushing this to stage in order to do some final testing though, so if we're missing any mappings we'll find out soon!

test/e2e/tests/token-introspection.spec.ts
MelissaAutumn: Probably a good plan, will get that fixed.

keycloak/config-cli/tbpro-aud-mapper.yaml
MelissaAutumn: Yeah it turns otu we don't need id token claims here. So I've flipped them to all false.

scripts/entry-keycloak.sh
Sancus: yeah running sync is fine. They should be independent so I don't think they need to relay information. But if one errors that is definitely a problem.

Sancus (approved):

keycloak/config-cli/tbpro-aud-mapper.yaml
markstos: From what I read about this config CLI tool, it's not only additive, but destructive.

It should be safe if your testing on stage includes reviewing the before/after snapshot states of Keycloak's data store to make sure that we only changed what we intended to.

Dockerfile.keycloak
```diff
@@ -53,22 +53,19 @@ COPY --chown=keycloak:keycloak keycloak/providers/keycloak-mfa-rest.jar /opt/key
...
 COPY --from=build /app/keycloak/themes/tbpro /opt/keycloak/themes/tbpro
 
 # keycloak-config-cli: declarative reconciler for the managed tbpro realm configuration.
 # config-cli 6.5.1 is built against Keycloak 26.5.5 and runs on the image's bundled Java 21.
-# apply-mfa-config.sh invokes it after startup.
+# apply-cli-config.sh invokes it after startup.
 COPY --from=adorsys/keycloak-config-cli:6.5.1-26.5.5 /app/keycloak-config-cli.jar /opt/keycloak/keycloak-config-cli.jar
 COPY --chown=keycloak:keycloak keycloak/config-cli /opt/keycloak/config-cli
 
 COPY scripts/entry-keycloak.sh scripts/entry.sh
```
markstos, line 66: It would be simpler and clearer if these two lines:

```
COPY scripts/entry-keycloak.sh scripts/entry.sh
ENTRYPOINT ["./scripts/entry.sh"]
```

became...

```
ENTRYPOINT ["./scripts/entry-keycloak.sh"]
```

In my review I was briefly confused about why `entry.sh` was being used in the keycloak container instead of `entry-keycloak.sh`.

keycloak/config-cli/tbpro-aud-mapper.yaml
MelissaAutumn: Yes, hence the PUT instead of PATCH comment. Keycloak requires the entire object if you want to change a property. 

The cli tool is a little better though, it'll at least handle single top-level properties iirc, but you still need a full list of like client scopes if you want to add or remove one. I think we have the cli tool set to no-delete as well.

Dockerfile.keycloak
MelissaAutumn: COPY is from outside of container to inside container, so we'd still need it. It's also a nit outside the PR :smile:

scripts/entry-keycloak.sh
```diff
@@ -22,8 +22,6 @@ KC_PID=$!
 # Forward termination so container stop stays graceful.
 trap 'kill -TERM "$KC_PID" 2>/dev/null' TERM INT
 
-/bin/bash "$SCRIPT_DIR/apply-cli-config.sh" "tbpro-mfa-stepup" &
-/bin/bash "$SCRIPT_DIR/apply-cli-config.sh" "tbpro-aud-mapper" &
-
+/bin/bash "$SCRIPT_DIR/apply-cli-config.sh" "tbpro-mfa-stepup" && "$SCRIPT_DIR/apply-cli-config.sh" "tbpro-aud-mapper" &
```
markstos, line 25: This serializes the two steps, but the still run async in the background and it appears there no mechanism to notify us or failure or mark the container unhealthy if there's a problem.

We can accept that risk if we are happy with a test deployment to staging with a manual before/after check of desired state.

Otherwise, we could send a stronger signal that something is wrong by making the container crash if the import fails— because it's not in a state that we intend to have in production— Or we could keep running and notify ourselves through Sentry or another mechanism that something unexpected and potentially serious has gone wrong.

MelissaAutumn: Give on how long it takes to up a keycloak container sometimes I think this is OK. I'll cc @aatchison for visibility but this should be good to merge I think?

markstos: I'm OK with merging this in that considering better failure handling can be considered a separate scope. The issue that one of these startup scripts could fail without us being notified pre-dates this addition of a second start-up script.

PR #1237, opened 2026-08-26
Promote Keycloak 99bdead3 image to production

Sancus:
## What changed?

Pin the production Keycloak image to `keycloak-99bdead307c8c58ee9c8990041fe1b6db60de211`, the stage-verified build from main that includes the double-submit fix from #1230.


Sancus:
I already pushed this yesterday, to be clear.

Issue #1240, opened 2026-08-27
Add a monitoring script that diffs between stalwart / keycloak / accounts


MelissaAutumn:
We have and probably will again run into the situation where the stalwart email list (name=thundermail address) differs from keycloak's user list (username=thundermail address) and/or differs from tb account's user's list (username=thundermail address). 

If we could create a daily cron that monitors these three services and diffs the lists accordingly, we could do things like page if there's manual action required (like we need to manually remove an orphaned stalwart account.)

Issue #1238, opened 2026-08-27
User deletion fails to ensure related records in other apps are removed


Sancus:
In https://github.com/thunderbird/thunderbird-accounts/blob/99bdead307c8c58ee9c8990041fe1b6db60de211/src/thunderbird_accounts/authentication/utils.py#L128-L159 we catch exceptions when we delete the Keycloak identity and Stalwart principal, log the error, and then continue.

This is not safe because it can result in left over accounts in Stalwart. And due to the fact that accounts does not enforce a strict fail-closed identity relationship between each app, any failure to delete Stalwart accounts can cause a left over account to be re-associated with a different identity (if it happens to have the same username) in the future.

PR #1243, opened 2026-08-27
Dont link orphaned stalwart accounts to new users (Fixes #1239)

### LINKED ISSUE #1239: Ensure that orphaned/existing Stalwart principals cannot be linked to new user creations
Linked to Zendesk issue 

gz#7985<br><br><i>(related to <a href='https://tbpro.zendesk.com/agent/tickets/7985'>Zendesk ticket #7985</a>)</i>

MelissaAutumn:
Fixes [Ensure that orphaned/existing Stalwart principals cannot be linked to new user creations #1239](https://github.com/thunderbird/thunderbird-accounts/issues/1239) and also related to [User deletion fails to ensure related records in other apps are removed  #1238](https://github.com/thunderbird/thunderbird-accounts/issues/1238)

I don't think it fully completes #1238, but it's a good step in the right direction. I'm still thinking over some good actions to do if Stalwart (or Keycloak) fail to delete besides notify. (See [Add a monitoring script that diffs between stalwart / keycloak / accounts #1240](https://github.com/thunderbird/thunderbird-accounts/issues/1240) ).

This is mostly for Mark and Davi to review, I've added sancus for sanity check here. 

### Adds `is-address-taken-lookup-stalwart` waffle switch

As documented when enabled it'll add a Stalwart get_account lookup on is_address_taken. This function is used in the sign-up debounced username check, before the actual sign-up occurs, and during alias creation.

### create_stalwart_account no longer succeeds after account lookup

This code should have never left the initial prototype, and is directly the cause of this issue. Before if there was a Stalwart account we purposefully tied the new user with the account. This is the opposite of what we want, so now it'll error out. 

This could get tricky since this occurs during the Paddle webhook processing, but it'll error to Sentry and we take action as needed (like remove the old Stalwart account if needed, and re-run the 'activate subscription features' admin action.)

### Unindent delete stalwart account during delete_user_data

This made it look like it was reliant on delete keycloak to succeed. It used to be but not anymore, so make it more readable.

### Add is_address_taken tests

This should have been tested, and now it is.

### DIFF
--- a/src/thunderbird_accounts/mail/tasks.py
+++ b/src/thunderbird_accounts/mail/tasks.py
@@ -24,6 +24,7 @@ from thunderbird_accounts.mail.exceptions import (
     DomainNotFoundError,
     HostedDkimDeleteRetry,
     HostedDkimPublishRetry,
+    InvalidJMapResponseError,
 )
 from thunderbird_accounts.mail.models import Account, Email
 from thunderbird_accounts.core.types import TaskReturnStatus
@@ -317,6 +318,9 @@ def create_stalwart_account(
     but is still required. App Passwords can be set now, or later.
 
     Note: Email should be Thundermail address. Stalwart does not need your recovery email."""
+
+    user = User.objects.get(oidc_id=oidc_id)
+
     stalwart = MailClient()
     domain = email.split('@')[1]
 
@@ -326,11 +330,7 @@ def create_stalwart_account(
 
         raise TaskFailed(
             str(error),
-            {
-                'oidc_id': oidc_id,
-                'username': username,
-                'email': email,
-            },
+            {'oidc_id': oidc_id, 'user_uuid': user.uuid},
         )
 
     emails = [
@@ -347,33 +347,35 @@ def create_stalwart_account(
         _domain = alias.split('@')[1]
         _stalwart_check_or_create_domain_entry(stalwart, _domain)
 
-    # Lookup the account first, this shouldn't normally happen but if it does we shouldn't explode.
+    # Lookup the account first, this shouldn't happen but if we have a left-over account then it's a problem.
     try:
         stalwart_account = stalwart.get_account(username)
-        stalwart_emails = stalwart_account.get('emails', [])
-
-        # link the stalwart account
-        pkid = stalwart_account.get('id')
-
-        # Check the aliases
-        if emails != stalwart_emails:
-            # Diff of new emails
-            new_emails = set(emails) - set(stalwart_emails)
-            # Diff of the old emails
-            old_emails = set(stalwart_emails) - set(emails)
-
-            stalwart.save_email_addresses(username, list(new_emails))
-            stalwart.delete_email_addresses(username, list(old_emails))
+        logging.error(f'[create_stalwart_account] Account [{user.uuid}] already exists in Stalwart!')
+        raise TaskFailed(
+            str('Account already exists in Stalwart'),
+            {
+                'oidc_id': oidc_id,
+                'user_uuid': user.uuid,
+                'stalwart_pkid': stalwart_account.get('id'),
+            },
+        )
+    except InvalidJMapResponseError as ex:
+        # Cover any jmap response errors, these also require manual fixing (re-running activate sub features)
+        sentry_sdk.capture_exception(ex)
+        raise TaskFailed(
+            str(ex.validation_error),
+            {'oidc_id': oidc_id, 'user_uuid': user.uuid},
+        )
     except AccountNotFoundError:
-        # We need to create this after dkim and domain records exist
-        pkid = stalwart.create_account(emails, username, full_name, app_password, quota)
+        pass
 
-    user = User.objects.get(oidc_id=oidc_id)
+    # We need to create this after dkim and domain records exist
+    pkid = stalwart.create_account(emails, username, full_name, app_password, quota)
     now = datetime.datetime.now(datetime.UTC)
 
-    # Don't create the account if we already have it
+    # Don't create the account if we already have it (this is safe as this object contains no info by itself)
     # Also create their account objects
-    account, _created = Account.objects.update_or_create(
+    account, account_was_created = Account.objects.update_or_create(
         name=user.username,
         defaults={
             'active': True,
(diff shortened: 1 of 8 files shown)


src/thunderbird_accounts/mail/tests/test_tasks.py
```diff
@@ -263,7 +263,7 @@ def test_retries_with_context_when_cloudflare_delete_fails(self, cloudflare_clie
         self.assertEqual('RuntimeError', cm.exception.context['error_type'])
         set_sentry_context_mock.assert_called_once_with('hosted_dkim_delete_retry', cm.exception.context)
 
-
+@override_settings(USE_MAILCHIMP=False)
```
MelissaAutumn, line 282: These were silently failing with "user is not subscribed" errors.

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -350,33 +350,35 @@ def create_stalwart_account(
...
-        pkid = stalwart.create_account(emails, username, full_name, app_password, quota)
+        pass
 
-    user = User.objects.get(oidc_id=oidc_id)
+    # We need to create this after dkim and domain records exist
+    pkid = stalwart.create_account(emails, username, full_name, app_password, quota)
     now = datetime.datetime.now(datetime.UTC)
 
-    # Don't create the account if we already have it
+    # Don't create the account if we already have it (this is safe as this object contains no info by itself)
```
MelissaAutumn, line 388: This is the Account reference model, it's not great to have it exist but there's no personal info here. I can make it hard-fail if it exists though.

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -350,33 +350,35 @@ def create_stalwart_account(
...
+        logging.error(f'[create_stalwart_account] Account [{user.uuid}] already exists in Stalwart!')
+        raise TaskFailed(
+            str('Account already exists in Stalwart'),
+            {
+                'oidc_id': oidc_id,
+                'user_uuid': user.uuid,
+                'stalwart_pkid': stalwart_account.get('id'),
+            },
+        )
+    except InvalidJMapResponseError as ex:
```
davinotdavid: Just checking but we are only catching `InvalidJmapResponseError` here but in case there's other types of errors, we would continue the execution. Should we also add a more generic catch here to fail the task instead of passing through?

davinotdavid: Also, the same exception is being swallowed in `is_address_taken` [here](https://github.com/thunderbird/thunderbird-accounts/pull/1243/changes#diff-399b61475c644ea8a64f0105045f0db251678600fc1f7dd35424a036efa4933bR264).

Wouldn't this mean that sign up says it's ok -> user registers -> `create_stalwart_account` fails the tasks and then we end up having an orphaned-account?

src/thunderbird_accounts/mail/utils.py
```diff
@@ -251,6 +254,16 @@ def is_address_taken(email_address: str) -> bool:
     if aliases:
         return True
 
+    # If this switch is enabled then check stalwart if this address is taken as well
+    if waffle.switch_is_active(WAFFLE_FLAG_IS_ADDRESS_TAKEN_LOOKUP_STALWART):
+        stalwart = MailClient()
```
davinotdavid, line 268: `is_address_taken` is being used in `can_register_with_username` ([here](https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/authentication/utils.py#L94-L98)) in `is_username_available`'s API endpoint which is debounced in the frontend.

Perhaps it is fine / can be punted into a separate ticket and unsure if building the MailClient every time would incur into network issues but just wanted to raise this.

davinotdavid (commented):
I've created a new user and went through the normal sign up steps and deleted the account through Django Admin's action "Delete selected users" in `Authentication` -> `Users` but it doesn't delete the main Account object in Stalwart.

Just wanted to confirm that this is expected behaviour + some comments:

MelissaAutumn:
> I've created a new user and went through the normal sign up steps and deleted the account through Django Admin's action "Delete selected users" in `Authentication` -> `Users` but it doesn't delete the main Account object in Stalwart.


I think that might be due to that account not having a stalwart_id. I'll actually loosen that restriction tomorrow, and bring it back once we move to v0.16 (as we'll need it for querying, it's actually useless in v0.15 lol.) It's possible that was the real cause for this mess. 

But thanks, will give this a double check!

src/thunderbird_accounts/mail/utils.py
MelissaAutumn: Yeah good spot, I'll just shove the remote call into a bool param (like check_remote or something.) I'll keep the switch ofc.

src/thunderbird_accounts/mail/utils.py
```diff
@@ -251,6 +254,16 @@ def is_address_taken(email_address: str) -> bool:
...
+        try:
+            stalwart.get_account(email_address)
+            # It exists? That's a problem.
+            return True
+        except (AccountNotFoundError, InvalidJMapResponseError):
+            pass  # all okay
```
MelissaAutumn, line 274: I'll split off invalid jmap to return True.

src/thunderbird_accounts/mail/tasks.py
MelissaAutumn: Commented in is_address_taken. But yeah, I'll add a general catch here as well.

davinotdavid (approved):
Core fix lgtm and addressed the latest comments!

MelissaAutumn (commented):
LGTM now. We should maybe get one more unbiased review tohugh.

markstos (commented):
This looks good to me, but I want to understand more the intended us of the flag, as it seems more robust if the flag is enabled all the time. I see a couple potential intended uses:

- Maybe there's a concern about resource impact on Stalwart due to extra queries, so the flag allows a fast rollback if the load is too high. If the resource consumption is reasonable after a monitoring period, then making the flag is removed with the always-check-Stalwart behavior to simplify things.

- Or maybe there's some expectation that we may want to quickly switch off checking Stalwart indefinitely so it's intended as a permanent flag

The change here seems like it's doing some valuable work to prevent serious problems, so I'm in favor of the making the new behavior permanent after a monitoring period.

MelissaAutumn:
The change is intended to be permanent, but it's there in case of performance or stability issues on Stalwart's end. I don't have a monitoring period determined, but I think we can take a look at the end of the year and see if there's feature flags (as-in the entire project's usage) we can reasonably trim.

markstos (approved):
Approving now with clarity about the feature flag plan.

PR #1245, opened 2026-08-31
Bump Keycloak to 26.7.3

markstos:
## QA Log

 - Audited that "accounts" does not require further changes for
   the [documented breaking changes](https://www.keycloak.org/docs/latest/upgrading/#migration-changes):

 - I have NOT audited other Keycloak clients (Stormbox, Stalwart) to see if they might be impacted by the breaking changes.

 - Tested local full Docker image build locally.

### DIFF
--- a/Dockerfile.keycloak
+++ b/Dockerfile.keycloak
@@ -34,7 +34,7 @@ RUN rm -rf /app/keycloak/assets/*
 
 # release source — download a pinned, checksummed jar from GitHub Releases.
 # NOTE: requires the keycloak-mfa-rest v* release to exist before this image can build.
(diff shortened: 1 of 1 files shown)


MelissaAutumn:
Tagging @aatchison for any infra and @Sancus for stormbox review.  

Thanks for standing the patch up @markstos !

Sancus:
There are no problems with their breaking change for any of our connecting clients that I can find. And shouldn't be, reusing that prefix or any of those query strings would be silly.

Also, not it for doing pulumi up.

Sancus (approved):

PR #1249, opened 2026-08-31
Mirror stalwart legacy's hostname with the docker container name.

MelissaAutumn:
For v0.16 upgrade I've renamed stalwart to stalwart_legacy in the docker config. But I forgot tothe hostname so jmap calls will fail locally.


markstos:
Makes sense to me. Any concerns from infra perspective @aatchison ?

aatchison:
This is just for local dev I think, so gtg on my end. right ? cc @markstos @MelissaAutumn

MelissaAutumn:
yep just local dev here!

MelissaAutumn:
Oh I've done this PR wrong. Sorry this file shouldn't be included by default.

MelissaAutumn:
Okay sorry, config.toml gets copied over from config.toml.example via bootstrap.py. So had to update that and untrack config.toml.

markstos (approved):
I confirmed this was a local-only change, but it created a path mismatch with E2E logic.

I pushed a commit to resolve the mismatch.

PR #1250, opened 2026-08-31
Fix email alias validation for custom domains

### LINKED ISSUE #1244: Fix address char limit for custom domains with catch alls
#761 regressed the fix from #465 that allows fewer than 3 character addresses on custom domains. It brought the constraint back for domains that have a catch-all set.

The character limit should be the same regardless of whether a catch all is set.

gz#8004<br><br><i>(related to <a href='https://tbpro.zendesk.com/agent/tickets/8004'>Zendesk ticket #8004</a>)</i>

davinotdavid:
## What changed?

A small frontend-only change to validate min length only for shared domains (e.g. thundermail.com / tb.pro) and not for custom-domains regardless of the existence of a catch-all. Also, split the validations a little bit for better readability.

## Why?

We were preventing a user from having an alias for a custom domain that was less than the min length (e.g. me@mydomain.com).

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/1244

## Screenshots

https://github.com/user-attachments/assets/465a06e0-9c47-473d-a9dc-d75eb88c15eb

### DIFF
--- a/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/emailAliasValidation.test.mjs
+++ b/assets/app/vue/views/MailView/sections/EmailSettingsSection/components/emailAliasValidation.test.mjs
@@ -29,3 +29,27 @@ test('email alias validation blocks catch-all aliases on domains with an existin
     })
   ).toBe(EMAIL_ALIAS_VALIDATION_MESSAGES.MIN_LENGTH);
 });
+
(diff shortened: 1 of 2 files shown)


markstos (approved):
## QA Log

- Reviewed related issues
- Reviewed diff
- Ran tests locally
- Watched QA video

PR #1251, opened 2026-08-31
Forward display_name from sign up to keycloak & stalwart

### LINKED ISSUE #1242: Connect the sign-ups display name with account's user display name
I have not looked into this but I think it only goes to Keycloak and not on the account user model for some reason...my bad. 

gz#8010<br><br><i>(related to <a href='https://tbpro.zendesk.com/agent/tickets/8010'>Zendesk ticket #8010</a>)</i>

davinotdavid:
## What changed?

First, we are now actually getting the value input from the sign-up process and passing first into keycloak so it can be injected and retrieved as `display_name` afterwards when the Celery task calls Stalwart's `create_stalwart_account` to finally use it as `full_name` there.

To test this, you can onboard a new user by adding an email (for example added@mydomain.com) to the allow list in Django Admin, then navigating to `/sign-up?email=added@mydomain.com` and going through the steps. After being onboarded, you should see the display name added in the second step in the Mail dashboard instead of a blank field.

Disclaimer: used Claude Code for testing and PR review checks

## Why?

There were several points where we ignored the display name passed in the sign up. Going through the initial setup / FTUE, would always have an empty display name.

## Applicable Issues

Fixes https://github.com/thunderbird/thunderbird-accounts/issues/1242

### DIFF
--- a/src/thunderbird_accounts/mail/utils.py
+++ b/src/thunderbird_accounts/mail/utils.py
@@ -89,15 +89,11 @@ def create_stalwart_account(user, app_password: Optional[str] = None) -> bool:
     if user.plan_id:
         mail_quota = user.plan.mail_storage_bytes
 
-    full_name = (user.display_name or '').strip()
-    if not full_name or full_name == user.username:
-        full_name = user.get_full_name() or None
-
     tasks.create_stalwart_account.delay(
         oidc_id=user.oidc_id,
         username=user.username,
         email=user.username,
-        full_name=full_name or None,
+        full_name=user.display_name or None,
         app_password=app_password,
         quota=mail_quota,
     )
--- a/src/thunderbird_accounts/mail/utils.py
+++ b/src/thunderbird_accounts/mail/utils.py
@@ -89,8 +89,17 @@ def create_stalwart_account(user, app_password: Optional[str] = None) -> bool:
     if user.plan_id:
         mail_quota = user.plan.mail_storage_bytes
 
+    full_name = (user.display_name or '').strip()
+    if not full_name or full_name == user.username:
+        full_name = user.get_full_name() or None
+
     tasks.create_stalwart_account.delay(
-        oidc_id=user.oidc_id, username=user.username, email=user.username, app_password=app_password, quota=mail_quota
+        oidc_id=user.oidc_id,
+        username=user.username,
+        email=user.username,
+        full_name=full_name or None,
+        app_password=app_password,
+        quota=mail_quota,
     )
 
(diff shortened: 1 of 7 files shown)


MelissaAutumn (approved):
This looks like it should work, I'll give it a test tomorrow morning. 

I don't think we actually use first/last name for anything, we're at least not setting it afaik. They're django defaults, but I think they're fairly harmless here.

src/thunderbird_accounts/mail/utils.py
```diff
@@ -89,8 +89,17 @@ def create_stalwart_account(user, app_password: Optional[str] = None) -> bool:
     if user.plan_id:
         mail_quota = user.plan.mail_storage_bytes
 
+    full_name = (user.display_name or '').strip()
+    if not full_name or full_name == user.username:
+        full_name = user.get_full_name() or None
```
MelissaAutumn: Let's just focus on display_name.

src/thunderbird_accounts/mail/utils.py
```diff
@@ -89,8 +89,17 @@ def create_stalwart_account(user, app_password: Optional[str] = None) -> bool:
...
+
     tasks.create_stalwart_account.delay(
-        oidc_id=user.oidc_id, username=user.username, email=user.username, app_password=app_password, quota=mail_quota
+        oidc_id=user.oidc_id,
+        username=user.username,
+        email=user.username,
+        full_name=full_name or None,
```
MelissaAutumn: `user.display_name or None` will work here. If they want an empty space as a display name, why should we deny that?

MelissaAutumn (changes requested):
For some reason my user had a display_name set, but the Stalwart account was not created with it. 

Can you double check the flow and confirm that a new user from sign-up to subscription is set correctly? This could be a dev only paddle weirdness.

davinotdavid:
@MelissaAutumn changed the code to focus on `display_name` in the latest commit. Also, after deleting all my containers and volumes I was also experiencing the same problem of not having the Stalwart account created which was weird since it was definitely working before.

Turns out that after this change https://github.com/thunderbird/thunderbird-accounts/pull/1249 I had to run this:

```bash
cp config.toml.example mail/stalwart_legacy/etc/config.toml
docker compose restart stalwart_legacy
```

Here's a video demo:

https://github.com/user-attachments/assets/2f136089-b214-4175-bb08-47d230ff7a11

MelissaAutumn (approved):
Thanks for the quick change!

PR #1252, opened 2026-09-01
Scale accounts

ryanjjung:
We need to scale up Celery to help process a long queue. This demands at least 10 and allows up to 20 containers.


MelissaAutumn (approved):
I love it when nubmers go up!

PR #1256, opened 2026-09-01
Slow down the payment processing page refresh

### LINKED ISSUE #1253: Increase page refresh time for the "Payment processing" page
Normally it's a few seconds, but with increase demand it's a bit slower and the page flashes quite harshly. Increase it to 15 seconds, and maybe mention on the page it will refresh in 15 seconds?

aaspinwall:
## What this changes

The "Payment processing" page automatically refreshes itself while it waits for a payment to go through. Right now it refreshes every 5 seconds. When things are busy and the page takes a little longer to load, that quick refresh makes the screen flash harshly, which isn't a great experience.

This updates the page to refresh every 15 seconds instead, giving it more breathing room. It also adds a short line on the page letting people know that the page will refresh automatically in 15 seconds, so the refresh doesn't feel unexpected.

## What you'll see

- The waiting page now refreshes every 15 seconds rather than every 5.
- A new message on the page: "This page will refresh automatically in 15 seconds."

Fixes #1253

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -556,6 +556,7 @@
       "verification": {
         "title": "Your payment is being processed.",
         "text": "This step can take up to 48 hours to complete.",
+        "refreshNote": "This page will refresh automatically in 15 seconds.",
         "help": "Need help? Contact {0}.",
         "support": "support"
       }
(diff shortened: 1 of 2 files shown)


MelissaAutumn (approved):
lgtm, thanks!

Issue #1246, opened 2026-08-31
Add automatic retry for celery tasks


MelissaAutumn:
We can setup a base celery task class with some [auto retry logic](https://docs.celeryq.dev/en/stable/userguide/tasks.html#task-autoretry) for `OperationalError` (and maybe some additionals) which would help us guard against random db failures or blips.

We should include some exponential retry logic so we're not immediately hitting a downed db. 

Related sentry error
https://thunderbird.sentry.io/issues/7702684438/?environment=prod&project=4508965616680960&query=is%3Aunresolved&referrer=issue-stream

PR #1260, opened 2026-09-01
Add scheduled task base classes for easy automatic retries (#1246)

markstos:
## What changed?

Adds two new base classes for Celery tasks. Both have automatic retries
for recoverable conditions enabled by default.

 * **DatabaseTask**: Typically for database access that we expect to complete within five minutes. Retries on OperationalError
 * **PatientExternalServiceTask**: For connections with third parties where we can tolerate more downtime or flakines son their end. Automatically retries a number of recoverable conditions
  up to an hour.

Tasks are updated to use one or the other base class as appropriate.

## Why?

The base class pattern is used to make retrying on recoverable errors easy and consistent.

As to why some of the classification choices were made:

The tasks moved to `DatabaseTask` are local event-processing jobs:

    - paddle_transaction_event
    - paddle_subscription_event
    - paddle_product_event
    - update_thundermail_quota

The Paddle event tasks consume webhook data that has already arrived and primarily update local models. `update_thundermail_quota` is also somewhat misleadingly named: it updates local Account.quota fields; it doesn’t call Stalwart directly. For these, the relevant transient failure is the local database, so the "database" policy was applied.

## Limitations and Notes

Some tasks that /intended/ to have automatic retries before weren't completely configured because they didn't have `autoretry_for` set. So this work /actually/ enables some retry logic that we intended to work before.

## Applicable Issues

 * #1246

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -76,6 +76,11 @@ SENTRY_TRACES_SAMPLE_RATE=0.0
 CELERY_BROKER=redis://redis:6379
 CELERY_BACKEND=redis://redis:6379
 CELERY_EAGER=False
+# Maximum delay in seconds between automatic database task retries.
+CELERY_DATABASE_TASK_RETRY_BACKOFF_MAX=300
+# Patient background tasks tolerate external-service outages for much longer.
+CELERY_PATIENT_TASK_RETRY_BACKOFF_MAX=3600
+CELERY_PATIENT_TASK_MAX_RETRIES=24
 
 # Zendesk (contact support form)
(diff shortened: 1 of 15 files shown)


MelissaAutumn (approved):
Took me a minute to walk through the code, looks good though. Could you make and follow-up with a PR for a docs/tasks.rst update? It doesn't have to be paragraphs just something short and to the point on how to use these.

markstos:
> Could you make and follow-up with a PR for a docs/tasks.rst update? 

Will do.

PR #1266, opened 2026-09-02
is_active=False now clears stalwart app password (+ display_name fix) (Fixes #1265)

### LINKED ISSUE #1265: Clear app passwords in Stalwart when User is deactivated

MelissaAutumn:
Fixes #1265

Now when the admin form is submitted with is_active=False it will reach out and clear Stalwart's app passwords (including the appointment one.) 

This also includes a small fix I noticed while I was here (sorry). We don't actually use first/last name anywhere, they're django defaults. But we do use display_name, and that's what a stalwart account is created with. So if there's an update, use display_name. I don't think this will actually do anything since Stalwart is the source of truth here and display_name is a read-only field for that reason. 

I've also guarded against some errors that might occur.

### DIFF
--- a/src/thunderbird_accounts/authentication/admin/forms.py
+++ b/src/thunderbird_accounts/authentication/admin/forms.py
@@ -1,3 +1,5 @@
+from typing import Optional
+from thunderbird_accounts.mail.exceptions import InvalidJMapResponseError, AccountNotFoundError
 from thunderbird_accounts.authentication.utils import KeycloakRequiredAction
 import zoneinfo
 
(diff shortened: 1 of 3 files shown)


src/thunderbird_accounts/authentication/admin/forms.py
```diff
@@ -161,21 +162,25 @@ def clean(self):
...
         username = self.cleaned_data.get('username')
-        full_name = f'{self.cleaned_data.get("first_name")} {self.cleaned_data.get("last_name")}'.strip()
+        display_name = (self.cleaned_data.get('display_name') or '').strip()
+        is_active = self.cleaned_data.get('is_active')
 
         # Update the user on keycloak's end
         keycloak = KeycloakClient()
         try:
+            if not oidc_id:
+                raise UpdateUserError('[Accounts Error] No oidc_id found, cannot update Keycloak.', username)
```
davinotdavid, line 174: Nit: This could be an earlier gate right after we assign the username variable instead

davinotdavid (approved):
lgtm!

## QA Log

- Code diff
- Manually tested through marking a user that has an app password in Stalwart as inactive and app passwords are deleted
- Manually tested that a previous inactive user with an app password (created after the user was inactive, manually), when flipped over to active, doesn't delete the app passwords but flipping it again deletes.

src/thunderbird_accounts/authentication/admin/forms.py
MelissaAutumn: Technically it only blocks Keycloak updates not stalwart updates.

PR #1291, opened 2026-09-08
Recover from user going back to /oidc/callback after a successful login (#1271)

markstos:
## What changed and why?

If a user navigated back in history to /oidc/callback URL, this could
result in a generic "Bad Request (400)" response as well as a Sentry issue.

Now we recover from this state by navigating forward back to the successful
login state result.

Sentry Issue: THUNDERBIRD-ACCOUNTS-2J

## Limitations and Notes

Malicious visits to this URL will still fail. This changes only the failure
response and not the authentication decision.

## Applicable Issues

 * [SuspiciousOperation: OIDC callback state not found in session `oidc_states`! · Issue #1271 · thunderbird/thunderbird-accounts](https://github.com/thunderbird/thunderbird-accounts/issues/1271)

### DIFF
--- a/src/thunderbird_accounts/authentication/views.py
+++ b/src/thunderbird_accounts/authentication/views.py
@@ -1,4 +1,5 @@
 import csv
+import logging
 import re
 
 from django.contrib import messages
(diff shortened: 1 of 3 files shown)


src/thunderbird_accounts/authentication/views.py
```diff
@@ -17,13 +18,31 @@
...
+    def get(self, request):
+        state = request.GET.get('state')
+        if (
+            request.GET.get('code')
+            and state
+            and 'oidc_states' in request.session
+            and state not in request.session['oidc_states']
+        ):
+            logger.warning('OIDC callback state was not found in the session; starting a fresh login flow')
+            return HttpResponseRedirect(reverse(settings.LOGIN_URL))
```
MelissaAutumn, line 43: We might want to log them out of django if they're logged in. 

So like
```python
if request.user.is_authenticated:
  logout(request)
```

Will need to test that though, we had a case a while back where one of the auths were desynced from the other, causing infinite page reloads as one said they were auth, but the other could not auth.

MelissaAutumn (changes requested):
Looks good, just one change request and maybe a test scenario?

src/thunderbird_accounts/authentication/views.py
MelissaAutumn: Hmm on second thought, the case I mentioned should be fixed with the oidc introspection changes we finished, and this PR is for folks who accidentally hit back after a login.

MelissaAutumn (approved):
Looks good on my end! It might be worth using javascript's / vue router's history api to remove the login page from the history so they can't go back. But that sounds complicated, and this works.

PR #1295, opened 2026-09-08
Keep /oidc/callback out of browser history (#1271)

markstos:
## What changed and why

Here we have /oidc/callback redirect via the frontend so that we can call
location.replace(), which will edit /oidc/callback out of the browser history.

This provides a cleaner browser history for users and prevents them from going
"back" to a URL that doesn't make sense to revisit.

This is a follup up to [Recover from user going back to /oidc/callback after a successful login (#1271) by markstos · Pull Request #1291 · thunderbird/thunderbird-accounts](https://github.com/thunderbird/thunderbird-accounts/pull/1291) where Mel suggested keeping /oidc/callback out of the browser history.

## QA Log

 * Because this is part of an auth flow, particular car was given to consider security issues. It's safe from XSS, it's not an open redirect, and we don't forward on the referrer header. This is a safe redirection method
 * Checked that other places are using this method during OIDC flows. Okta's official SDK is one example that uses window.location.replace() the same way: https://github.com/okta/okta-auth-js#handleredirectoriginaluri
 * After coding, manually tested the entry appears gone from the browser history.

## Limitations

 * Like much of our app, this requires JavaScript to be enabled.

## Applicable Issues

 * [SuspiciousOperation: OIDC callback state not found in session `oidc_states`! · Issue #1271 · thunderbird/thunderbird-accounts](https://github.com/thunderbird/thunderbird-accounts/issues/1271)

->

### DIFF
--- a/src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
+++ b/src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
@@ -1,0 +1,14 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Completing login</title>
+  </head>
+  <body>
+    {{ redirect_url|json_script:"oidc-callback-redirect-url" }}
+    <script>
+      const redirectUrl = JSON.parse(document.getElementById('oidc-callback-redirect-url').textContent);
+      window.location.replace(redirectUrl);
+    </script>
+  </body>
+</html>
--- a/src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
+++ b/src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
@@ -2,9 +2,15 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <title>Completing login</title>
+    <noscript>
+      <meta http-equiv="refresh" content="0; url={{ redirect_url }}">
+    </noscript>
+    <title>Logging you in</title>
   </head>
   <body>
+    <noscript>
+      <p><a href="{{ redirect_url }}">Continue to complete your login</a>.</p>
+    </noscript>
     {{ redirect_url|json_script:"oidc-callback-redirect-url" }}
     <script>
       const redirectUrl = JSON.parse(document.getElementById('oidc-callback-redirect-url').textContent);
--- a/src/thunderbird_accounts/authentication/tests/test_views.py
+++ b/src/thunderbird_accounts/authentication/tests/test_views.py
@@ -1,3 +1,5 @@
+from unittest.mock import patch
+
 from django.conf import settings
 from django.contrib.auth.models import Permission
 from django.contrib.messages import get_messages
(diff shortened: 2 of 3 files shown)


src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
```diff
@@ -0,0 +1,14 @@
...
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Completing login</title>
+  </head>
+  <body>
+    {{ redirect_url|json_script:"oidc-callback-redirect-url" }}
+    <script>
+      const redirectUrl = JSON.parse(document.getElementById('oidc-callback-redirect-url').textContent);
+      window.location.replace(redirectUrl);
```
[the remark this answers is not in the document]
markstos: This is a false-positive that adds unnecessary security theater. Our backend has several protections before the value reaches the template:

  1. The normal OIDC request stores oidc_login_next only after url_has_allowed_host_and_scheme() validates its scheme and  allowed host.
  2. Other code paths store values generated by reverse() or request.get_full_path(), which are relative application paths.
  3. The session is server-managed, so an attacker cannot simply edit oidc_login_next in their cookie.
  4. Crucially, our override calls: `redirect_response = super().login_success()
      The parent constructs a Django HttpResponseRedirect. Django rejects dangerous schemes such as       javascript: and data:   at that point. We only read .url after that validation succeeds.

  5. json_script escapes HTML-significant characters such as <, >, and &, preventing </script> breakout.

src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Completing login</title>
+  </head>
+  <body>
+    {{ redirect_url|json_script:"oidc-callback-redirect-url" }}
```
MelissaAutumn, line 14: Maybe we should add the noscript core/temapltes/index.html. (After https://github.com/thunderbird/thunderbird-accounts/pull/1296 is merged ofc.) 

I know you can't actually get here without noscript, but given the state of browser plugins or things like piholes that can block specific urls, we should try and guard against weird behaviour outside of our control.

src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Completing login</title>
```
MelissaAutumn: Maaaybe `Logging you in`?

src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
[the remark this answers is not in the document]
MelissaAutumn: Agreed a false positive. I hate these guys. If we wanted to be super duper careful we could do an origin check in views.py.

MelissaAutumn (approved):
Works on my machine, I have some suggestions you can finish up here or follow up in another PR.

src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
markstos: In my testing, the page loading and redirect happen so immediate that this page is not visible at all. That's why there's no real design to the page. If we are going to be considered about it being visible enough for wordsmitthing, then the whole page design should be considered. An irony there is that a more extensive design would slow down the page load, possibly making it visible when it wouldn't otherwise be. My take is that this not a level of user experience we need to get into.

MelissaAutumn: Yep no worries there, the phrase "Completing login" just seemed weird to me. But I am myself, pretty weird. So it's up to you!

src/thunderbird_accounts/authentication/templates/authentication/oidc_callback_success.html
markstos: I've pushed a fixup commit with layered noscript support:

- First, we try a meta refresh tag, which works without JS.
- Then if JavaScript is blocked and auto-refresh is blocked too... we offer an explicit HTML link to click.

PR #1296, opened 2026-09-08
Fix incorrect django templating for index.html in the noscript element.

MelissaAutumn:
A minor thing I noticed while reviewing another PR. The set tag doesn't seem to exist, so I just inlined it. 

image


markstos (approved):
## QA Log

Avoiding `set` is a nice simplification here. It was unnecessary.

- Reviewed diff
- Confirmed `set` exists for Jinja2, but not for Django templates
- Confirmed Django templates support both `translate` and `blocktranslate`.

Issue #1270, opened 2026-09-02
delete_dkim clears unrelated signers via Stalwart substring filter


Sancus:
`delete_dkim` queries Stalwart `/settings/group` with `filter=<domain>` and treats every returned signer `_id` as belonging to that domain. It clears `signature.<id>.` for all matches with no exact domain or signer-ID validation.

**Confirmed example:** Removing custom domain `thundermail.com.au` also deleted `rsa-thundermail.com` and `ed25519-thundermail.com` because their signer IDs are prefixes of the `.au` domain IDs returned by the substring filter.

Buggy code: https://github.com/thunderbird/thunderbird-accounts/blob/4bc2dc238828ac9c788585784d442e51772d80e8/src/thunderbird_accounts/mail/clients/mail_client_legacy.py#L378-L404

**Fix:** Only clear signers after verifying an exact domain match and the expected signer ID for that domain.

Sancus:
This may be resolved by the 0.16 patch.

https://github.com/thunderbird/thunderbird-accounts/blob/4bc2dc238828ac9c788585784d442e51772d80e8/src/thunderbird_accounts/mail/clients/mail_client_jmap.py#L1027

Sancus:
We should also double check consistency of all other DKIM records in the Stalwart database, there were reports that some users' custom domains are also missing them.

rtanglao:
@markstos no DKIM from my custom domain, glamrocnamecheap.com:

<details>
<summary>click for headers without DKIM</summary>

X-Mozilla-Status: 0001
X-Mozilla-Status2: 00000000
Delivered-To: nemo@thunderbird.net
Received: by 2002:a05:6038:c53a:b0:79b:8c51:59a7 with SMTP id nd10csp7000781nnb;
        Tue, 8 Sep 2026 12:14:42 -0700 (PDT)
X-Received: by 2002:a05:6000:2008:b0:485:8226:c69d with SMTP id ffacd0b85a97d-4858226c991mr39177094f8f.28.1788894882654;
        Tue, 08 Sep 2026 12:14:42 -0700 (PDT)
ARC-Seal: i=1; a=rsa-sha256; t=1788894882; cv=none;
        d=google.com; s=arc-20260327;
        b=m7/Wt1cf9X3jeGvCP8FP3KT4jZexUSS9XEHAT6/t6B05EUw5SIbd7hiNgrzXMcYg60
[...]

PR #1297, opened 2026-09-08
Workaround Stalwart v0.15 domain deletion bug (#1270)

markstos:
## What changed and why?

As explained in the linked issue, Stalwart could return more domains
then it should. We were trusting Stalwarts list and ending up deleting
too many domains.

Here we guard against that behavior by explicitly building a list of
domain records to delete instead of trusting Stalwart.

Stalwart 0.16+ doesn't have this bug.

## Limitations and Notes

- Although v0.15 is on the chopping block, this matters to customers impacted now and was an easy enough fix to justify a short-term fix.

- The scope here does not address restoring records which were previously deleted by this bug. That
scope can be addressed seperately.

## Related Issues

 * [delete_dkim clears unrelated signers via Stalwart substring filter · Issue #1270 · thunderbird/thunderbird-accounts](https://github.com/thunderbird/thunderbird-accounts/issues/1270#issuecomment-5590521193)
 * Follow-up issue for repairing impacted accounts: [Audit DKIM records for consistency after fixing DKIM deletion bug · Issue #1298 ](https://github.com/thunderbird/thunderbird-accounts/issues/1298)

## QA Log

 - With AI-assist, reviewed related Stalwart v0.15 and v0.16 source code to confirm bug in v0.15
   and fix in v0.16.

### DIFF
--- a/src/thunderbird_accounts/mail/tests/test_clients/test_legacy.py
+++ b/src/thunderbird_accounts/mail/tests/test_clients/test_legacy.py
@@ -478,7 +478,8 @@ class TestMailClientDeleteDkim(TestCase):
         success_get_response = requests.Response()
         success_get_response.status_code = 200
         success_get_response._content = bytes(
-            json.dumps({'data': {'total': 1, 'items': [{'_id': f'rsa-{self.domain}'}]}}), 'utf-8'
+            json.dumps({'data': {'total': 1, 'items': [{'_id': f'rsa-{self.domain}', 'domain': self.domain}]}}),
+            'utf-8',
         )
 
         requests_get_mock.return_value = success_get_response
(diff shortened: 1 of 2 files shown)


src/thunderbird_accounts/mail/tests/test_clients/test_legacy.py
```diff
@@ -496,6 +497,68 @@ def test_success(self, requests_post_mock: MagicMock, requests_get_mock: MagicMo
...
+
+        response = self.mail_client.delete_dkim(domain)
+
+        self.assertIs(response, success_post_response)
+        self.assertEqual(
+            requests_post_mock.call_args.kwargs['json'],
+            [
+                {'type': 'clear', 'prefix': f'signature.ed25519-{domain}.'},
+                {'type': 'clear', 'prefix': f'signature.rsa-{domain}.'},
+            ],
```
MelissaAutumn, line 546: This test is unfortunately flawed as it's a Stalwart v0.15 bug not an our bug.

Now the opposite can occur, since we're using prefix. If you have `example.com.au`, and a separate `example.com` as selector ids, removing `example.com` will also remove `example.com.au`.

I'm not sure if we can fix this in v0.15, except for maybe not using dots in the selector ids, and it's too late to change that. There is a filter param that would have been useful if it allowed for exact matching instead of contains -.-

https://github.com/stalwartlabs/stalwart/blob/v0.15.5/crates/http/src/management/settings.rs#L289-L303

https://github.com/user-attachments/assets/b4372da6-1a72-4f73-9648-6f3faa148744

MelissaAutumn (changes requested):
I do hope I'm wrong, but I think we can't really fix this here? Or is my finding a separate bug?

Sancus:
You can clear specific keys only by doing this in 0.15:
```
data = [
    {
        'type': 'delete',
        'keys': [
            f'signature.{dkim_id}.{key}'
            for dkim_id in sorted(dkim_ids)
            for key in STALWART_DKIM_SIGNER_KEYS
        ],
    }
]
response = requests.post(
    f'{self.api_url}/settings',
    json=data,
    headers=self.authorized_headers,
    verify=settings.VERIFY_PRIVATE_LINK_SSL,
)
response.raise_for_status()

return response
```
See Stalwart webadmin code: https://github.com/stalwartlabs/webadmin/blob/main/src/pages/config/list.rs#L156-L167

To do this you need to enumerate all the key types though, and there are a bunch

```
# Keys written by Stalwart v0.15 `create_dkim_key` for every signer.
STALWART_DKIM_SIGNER_KEYS = (
    'private-key',
    'domain',
    'selector',
    'algorithm',
    'canonicalization',
    'headers.0',
    'headers.1',
    'headers.2',
    'headers.3',
    'headers.4',
    'report',
)
```

markstos:
Thanks for the reviews. I've pushed a fixup commit which use the exact-domain deletion method that works in v0.15.

While v0.16 cutover is growing closer by the day, this might still help deliverability on v0.15 for a few folks.

Once the upgrade is complete, we can rip out the legacy v0.15 code and this hack.

PR #1301, opened 2026-09-11
Update quota lookup to handle missing accounts

### LINKED ISSUE #1200: Handle AccountNotFoundError case specially when quota lookup fails
Currently when we fail to lookup quotas via /api/v1/subscription/plan/info, we treat all lookups failure the same in thunderbird_accounts/subscription/views.py around line 280: We log at the error level (which notifies Sentry) and we return a generic failure message to the user.

There's one known case that we should handle differently which would address a common Sentry error and provide a clearer error message to users:

The app may throw the specific `AccountNotFoundError`. First, it would be useful to research how this state arises. Maybe a user logged in but in they are offboarding, so the Stalwart account is gone but their session still there? 

Assuming we find that this is a state we expect to happen sometimes, then we should handle it different:

- Log at the warning level instead of error
- Update the JSON response with a more specific error: "User not found".  Consider status=404 instead of 500 in this case, since it's not a server error. 

Sentry Issue: [THUNDERBIRD-ACCOUNTS-7J](https://thunderbird.sentry.io/issues/7459187242/?referrer=github_integration)

```
Error getting used quota: AccountNotFoundError: [email]
```

MooseTheRebel:
_I'll preface this PR with an apology, I'm sorry I didn't get to this sooner!_ 🙂

## What changed?

Catch `AccountNotFoundError` specifically in `subscription/views.py`'s `get_subscription_plan_info` function, and handle it per the suggested behavior in #1200 . Also added a test.

*AI disclosure: this change was developed with gemma4:26b-a4b-it-q4_K_M (Google) running locally via ollama. I reviewed the GitHub issue with it, discussed and picked the fix scope, and reviewed/edited the diff and tests it wrote (including trimming a test I felt was unnecessary regression coverage) before opening this PR.*

## Why?

`AccountNotFoundError` appears to be an expected/recoverable state, potentially during offboarding, though its cause is still unconfirmed (see [my open question](https://github.com/thunderbird/thunderbird-accounts/issues/1200#issuecomment-5297810695)).

This avoids treating that case as an unexpected server error while giving the client a clearer response.

## Limitations and Notes

- Left `mail/views.py`'s `app_password_set` alone, but it also triggers a `AccountNotFoundError`. This might be something to massage as well.
- This PR doesn't dig into *why* the account goes missing while the session persists (e.g. confirming the offboarding theory) — that's still open per the question linked above.

## Applicable Issues

Closes #1200

## Screenshots

N/A — backend-only change, no UI impact.

cc/ @markstos


markstos:
Adding to the context that is our top unresolved Sentry issue for the last 7 days. (More feedback coming later).

src/thunderbird_accounts/subscription/views.py
```diff
@@ -277,6 +278,10 @@ def get_subscription_plan_info(request: Request, paddle: Client):
         account = stalwart_client.get_account(request.user.stalwart_primary_email)
         quota = account.get('quota', 0)
         used_quota = account.get('usedQuota', 0)
+    except AccountNotFoundError:
+        # Expected when the Stalwart account is gone but the user's session is still active (e.g. offboarding)
+        logging.warning(f'Quota lookup failed, account not found: {request.user.stalwart_primary_email}')
+        return JsonResponse({'success': False, 'error': 'User not found'}, status=404)
```
MelissaAutumn:
```suggestion
        return JsonResponse({'success': False, 'error': 'Subscription plan information is not available at this moment.'}, status=404)
```

Something like that, this only happens during a blip before the user is attached to a stalwart account.

MelissaAutumn (changes requested):

MooseTheRebel:
@MelissaAutumn Thank you for the suggested improvement! Committed 🙂

MooseTheRebel:
@MelissaAutumn Sorry to ping again, I just wanted to note that I've rebased this branch against the current `main` and fixed two minor items (formatting and a failing test).

Tests are passing now, minus the failures we've seen before in `manage.py` calls (due to me being an outsider).

markstos (approved):
I researched the root cause some more. This state could be entered from this sequence:

1. Staff member removes account. 
2. Account is now in a state where Keycloak and Stalwart may reflect the removal, but the session is still active here in the "accounts" service.
3. User still in their stale session triggers this "quota" endpoint and gets the error.

While I agree the change here is good, the better long term fix is to implement #1158, backchannel logout, which may narrow the window that this might happen in to be so small to possibly eliminate this case in practice. With backchannel logout, keycloak would immediately notify accounts of session termination, which should end the session before this is triggered.

## Related tickets

I don't like the verbosity it takes to setup the fixtures, but that's an issue that's specific to the work here, it's a general test suite problem. I used this case as inspiration to open #1319 to propose that we add some factory methods to make it more pleasant to setup test fixtures. 

## QA Log

- Reviewed diff
- Researched root cause further
- Considered slimming up test coverage since this is an edge case. Decided to keep it but minimized the checks a bit. If #1319 leads to refactoring this, the test coverage would look more proposal to the change.

MelissaAutumn (approved):

PR #1305, opened 2026-09-11
Add aria-hidden to decorative icons and labels to icon-only buttons

### LINKED ISSUE #1304: a11y: Add aria-hidden="true" to decorative icons
## Describe the bug

On the fronted, we should add `aria-hidden="true"` attribute to decorative icons that screen readers should ignore.

Spec: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden

In the PR, Including a "QA Log" section that notes how you checked the icons where this added were decorative and not semantically important. Screenshots of impacted locations may be helpful.

rishabhiskawai:
Fixes #1304.

Hides 46 decorative Phosphor icons from screen readers and gives 10 icon-only buttons accessible names.

What changed
- `aria-hidden="true"` on every decorative `ph-*` icon: status/info icons next to visible text, card/section header icons beside titles, copy-button icons (buttons already labeled), badge icons beside status text, and link/button trailing icons.
- `aria-label` on 10 icon-only buttons that previously exposed just an unlabeled icon: modal/notice dismiss buttons, both kebab menus, and the allow-list remove button. 4 new locale keys added (en is the only app locale); 2 cases reuse existing keys.

## QA Log
- Enumerated all 55 `ph-*` usages in `assets/app/vue` and checked each one's context: every tagged icon sits next to visible text, inside a titled card/section, or inside a button that now has (or already had) an accessible name.
- Skipped `DetailsSummary.vue`'s chevron on purpose: its parent span already carries `aria-hidden="true"", so it was covered.
- Verified no icon-only button is left without a name: searched all buttons whose only child is a `ph-*` icon and confirmed each has `aria-label` or `title`.
- Existing `aria-hidden` usages (warning icons in domain forms, accordion chevron, tab icons) left untouched.
- Gates: frontend vitest 17/17 pass, eslint clean on all touched files.

### DIFF
--- a/assets/app/vue/locales/en.json
+++ b/assets/app/vue/locales/en.json
@@ -37,6 +37,9 @@
     },
     "trafficBanner": {
       "message": "We're currently experiencing high demand. Pages may load slowly, and some features may not work. Thank you for your patience."
+    },
+    "genericModal": {
+      "close": "Close dialog"
     }
   },
   "views": {
@@ -223,7 +226,8 @@
           "displayNameRequired": "Display name is required",
           "emailAliasesArticleTitle": "Setting Up Email Aliases",
           "emailAliasesCatchAllArticleTitle": "Setting Up a Catch-All Alias",
-          "emailAliasSupportText": "For help with aliases, see {emailAliasesSupportLink}. For catch-all aliases, see {emailAliasesCatchAllSupportLink}."
+          "emailAliasSupportText": "For help with aliases, see {emailAliasesSupportLink}. For catch-all aliases, see {emailAliasesCatchAllSupportLink}.",
+          "aliasActions": "Alias actions"
         },
         "customDomains": {
           "customDomains": "Custom Domains",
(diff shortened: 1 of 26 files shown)


assets/app/vue/locales/en.json
```diff
@@ -37,6 +37,9 @@
     },
     "trafficBanner": {
       "message": "We're currently experiencing high demand. Pages may load slowly, and some features may not work. Thank you for your patience."
+    },
+    "genericModal": {
+      "close": "Close dialog"
```
markstos:
```suggestion
      "close": "Close"
```

markstos (changes requested):
This looks great to me me with one small wording change.

While  tags that load SVGs could use `alt=""` instead, these icons are rendered as `<svg>` tags where `aria-hidden` is a better approach

markstos (approved):

markstos:
If you could rebase once more and resolve the conflicts, this looks ready to merge, thanks!

Issue #253, opened 2025-09-03
Ensure accounts are cleaned up correctly on removal


MelissaAutumn:
The error we're recieving on stage (resource already managed) is basically a record already exists. Currently it's due to the messy nature of stage, but we'll need to ensure we can handle any such errors in the future by detecting this issue and hitting the correct stalwart api.

MelissaAutumn:
We are hitting stalwart and keycloak on deleting a user in account's admin panel. I have not confirmed if deleting a principal object is enough to clear all mail data though.

PR #1308, opened 2026-09-15
Gate Keycloak logins on Stalwart provisioning (thundermail-access role)

Sancus:
## What changed?

- **Keycloak** (`keycloak/config-cli/tbpro-mfa-stepup.yaml`): `browser-mfa-stepup` wraps its alternatives in a REQUIRED `authenticate` subflow followed by a CONDITIONAL `mail-access-gate` that denies any client lacking the `pre-provisioning-login` client scope when the user lacks the `thundermail-access` realm role. `TBPRO_MAIL_GATE` (`CONDITIONAL`/`DISABLED`) sets the gate's requirement per environment; `apply-mfa-config.sh` falls back to `DISABLED` on unknown values. The tbpro error page shows the deny reason. The local realm import provides the role, scope and `tb-accounts` exemption.
- **Accounts**: `grant_mail_access_role` task, enqueued after `create_stalwart_account` succeeds. `KeycloakClient` realm-role and client-scope helpers plus a one-shot token refresh on 401. `backfill_mail_access_role` management command for existing users.
- Tests for all of the above.

Written by an AI agent (Cursor) under review of the author; verified on a local stack (config-cli update path from the current flow, deny/allow/exempt/SSO-cookie logins, gate on/off toggling) and the full Django suite.

## Why?

Stalwart's OIDC directory creates a principal on the first successful bearer authentication. A user who opens a mail client before the Celery provisioning task runs gets a principal Stalwart created, and `create_stalwart_account` then fails closed on "already exists" (the recent case; same mechanism as the old mailbox reuse). Gating token issuance in Keycloak means Stalwart never sees a token for an unprovisioned user, for any current or future mail client, with only the Accounts web client exempt.

## Limitations and Notes

- Rollout, in order: create the `pre-provisioning-login` scope and add it as a **default** scope of the Accounts client, create the `thundermail-access` role (Admin API; deliberately not declared to config-cli, whose managed mode would delete undeclared roles/scopes), deploy Accounts, run `backfill_mail_access_role --dry-run` then for real, then deploy Keycloak. A missing role or scope disables the gate; a present role and scope without the exemption would deny Accounts sign-in, which the backfill refuses to run into.
- Browser flow only: refresh/offline tokens issued earlier and direct-grant/device flows are not gated. Brokered logins would skip the gate; the realm has no identity providers.
- Existing orphaned Stalwart principals are out of scope (#1238, #253); role revocation on subscription cancellation is not implemented (#348).

## Applicable Issues

No dedicated issue yet; related to #1238 and #253.

## Screenshots

n/a


src/thunderbird_accounts/authentication/clients.py
```diff
@@ -177,6 +188,70 @@ def delete_user(self, oidc_id: str):
...
+        :raises RoleMappingError: If the request fails."""
+        try:
+            return self.request(endpoint, method, **kwargs)
+        except RequestException as exc:
+            sentry_sdk.capture_exception(exc)
+            if exc.response is not None:
+                error = f'Error<{exc.response.status_code}>: {exc.response.content.decode()}'
+            else:
+                error = f'Error<{exc}>: No response!'
+            raise RoleMappingError(error=error, role_name=role_name, oidc_id=oidc_id) from exc
```
MelissaAutumn, line 212: warning log? In case we miss a sentry capture (quota, or otherwise.)

src/thunderbird_accounts/mail/management/commands/backfill_mail_access_role.py
```diff
@@ -0,0 +1,97 @@
...
+        .exclude(account__stalwart_id='')
+        .distinct()
+        .order_by('username')
+    )
+
+
+def unprovisioned_subscribers():
+    """Active subscribers without a linked mailbox."""
+    return (
+        User.objects.filter(subscription__status=Subscription.StatusValues.ACTIVE)
```
MelissaAutumn, line 37: I'm dumb today, would this filter out users without subscription objects at all? (aka the real bulk of unprovisioned users.)

src/thunderbird_accounts/mail/tasks.py
```diff
@@ -433,6 +471,8 @@ def create_stalwart_account(
             },
         )
 
+    grant_mail_access_role.delay(oidc_id)
```
MelissaAutumn, line 474: The other place this could live is with activate_subscription_features, this can be manually called from the admin panel too so it could be handy to have this live there.

But either should be good.

MelissaAutumn (approved):
PR works on my local environment. Thanks!

I could only really test it with accounts itself, removing the default scope does indeed block my login for users without the role. Adding it back in lets them login again, and removing it and logging in with a user with the role works.

src/thunderbird_accounts/mail/management/commands/backfill_mail_access_role.py
MelissaAutumn: I am dumber than dumb today and misread this function. Ignore this :)

src/thunderbird_accounts/mail/tasks.py
Sancus: I added it there so the admin action can do it(in the block that requires a stalwart account to exist), but it needs to happen after we're sure the Stalwart account has been created so I think it needs to stay here too.

Issue #1197, opened 2026-08-13
Unify Main Navigation for Clearer Service Discovery


kewisch:
# 🎯 The objective

Improve the main navigation paradigm to offer a much clearer discovery path to users for all the available services. As we shift away from Thunderbird Pro as the overarching umbrella, and more towards Thundermail as the main name/product, and as Webmail gains more importance, we need to provide a cohesive experience in the header navigation.

### 🎒 Resources

- [Design's SoW](https://docs.google.com/document/d/1gLh_IsRlYvpQZPUD6sUySs5arYG-nPdUYwhmJZ0oKOE/edit)
- [Zeplin file for the Thunderbird Accounts changes](https://zpl.io/lrN1X3D)
- [Zeplin file for the Settings](https://zpl.io/WQ1BD8j)
- [Figjam for the IA phases](https://www.figma.com/board/1pcZ509x4nFhXeF48d7Qs1/tmail_web_2026-Q2_informationArchitecture_uxFlows?node-id=2651-498&t=H0anQ3DdpTMSwvy5-1)
- [Figma file for the High Fidelity designs](https://www.figma.com/design/TVEIcXtMqLTbIRyIYf4Tzt/tMail_web_2026-Q3_nav-IA_v003?node-id=1-6&p=f&t=612lcusEghrMuCpv-0)

### 🟢 In Scope

- Unification of the “Thundermail” logo usage.
- Dashboard pages update according to guidance from design
- Unification of navigation bar of all services.
- Unify the domain names

### 🔴 Out of Scope

- Settings pages update (this is in scope in the Design SoW, however it may turn out to be its own milestone for the engineering implementation)

### 🚧 Technical and/or Business Requirements

- This will require quite a few changes on both Accounts and Webmail / Stormbox since will be porting over most of the current Mail Dashboard into a Settings page inside Webmail. ([Zeplin reference](https://zpl.io/WQ1BD8j))
- All the services will need to have their main navigation updated to reflect the new designs.
- All the services will need to update their favicon + meta title tag.
- The new "App Switcher" component should probably be done in services-ui so it can be easily consumed across services.

PR #1311, opened 2026-09-15
Move AppPassword and DisplayName to Accounts dashboard

### LINKED ISSUE #1286: Move App Settings & Display Name into Account Settings section
Currently, Accounts dashboard have a "My Account" section that should be named to "Account Settings". Also, this section should host the Display Name (alongside it's in-place text input interaction for changing it) and App Password (alongside it's interactions as well).

This is how the new section should look like in the Accounts dashboard:

Image

Zeplin: https://zpl.io/lrN1X3D

davinotdavid:
## What changed?

- Moved `DisplayName` and `AppPasswords` from Mail Dashboard to Accounts Dashboard, including their in-place form functionality.
- Updated `useTour` / Tour Cards FTUE in Mail Dashboard to not have the AppPassword step anymore since it is not on the same page.
  - Also updated the link in the "GetStartedCard" that pointed to AppPassword to now point to the Accounts dashboard.
- Fixed E2E test that is current breaking on `main` (a label change from 'Your Current Subscription' to 'Current Subscription'). 

Zeplin: https://zpl.io/lrN1X3D

## Why?

- This is part of the Thundermail IA efforts!

## Limitations and Notes

- Badge styles need to be updates in services-ui https://github.com/thunderbird/services-ui/issues/319 so that's not matching the Zeplin just yet. Trivial change once that's done though!
- AppPassword tour step has been removed since it is not on the same page anymore. The re-introduction of this step has not yet being discussed in the Design team.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/1286

## Screenshots

Before
image

After
image

Tour / FTUE in Mail Dashboard with 1 less step

https://github.com/user-attachments/assets/4ff96d6d-2774-4bce-b6f3-d47a35406d85


markstos (changes requested):
One bug found:

- After the password is changed, I'm redirected back to /mail. Due to this move, I believe I should be redirected back to /dashboard.

## QA Log

 - Scanned diff
 - Compared with Zeplin spec:
 - Manually tested:
    - Changing display name
    - Changing password
    - Changing app password

davinotdavid:
@markstos Thanks for the review! 🙏 

> After the password is changed, I'm redirected back to /mail. Due to this move, I believe I should be redirected back to /dashboard.

This is the same behavior in prod and out-of-scope for this issue. By the end of the https://github.com/thunderbird/thunderbird-accounts/issues/1197 the idea is to not have `/mail` at all (it will become the /settings) and the redirects after Keycloak operations will then need to be changed.

Good catch though, I've just cut a ticket about it so we can remember to do this last.

markstos (approved):
Approving since my only feedback was out scope.

PR #1317, opened 2026-09-15
Update "Account deletion" section of Accounts dashboard

### LINKED ISSUE #1287: Update "Privacy & Data" section to be "Account Deletion"
The current "Privacy & Data" section should be renamed as "Account Deletion".

Also notice that the section copy + button has been also updated!

Image

Zeplin: https://zpl.io/lrN1X3D

davinotdavid:
## What changed?

- Renamed `PrivacyAndDataCard` component to `AccountDeletionCard` and updated its styles + copy.
- Updated the link button to be a `router-link` instead of a `link-button` wrapped in an `<a>` tag so that we make use of Vue Router effectively + fixed styles.

Zeplin: https://zpl.io/lrN1X3D

## Why?

- Part of the Thundermail IA effort!

## Limitations and Notes

- The colour of the "Contact support" link is `--colour-ti-critical` in Zeplin but until https://github.com/thunderbird/services-ui/issues/317 is done and the services-ui version is updated, it will be slightly wrong.

## Applicable Issues

Closes https://github.com/thunderbird/thunderbird-accounts/issues/1287

## Screenshots

Before
image

After
image


markstos (approved):
## QA Log

- Scanned diff
- Confirmed "After" screenshot matches design spec
- Using <router-link> is consistent with what we do in similar cases.

Issue #1280, opened 2026-09-03
E2E stability and reporting improvements


rwood-moz:
Investigate and look into code changes to fix some of the E2E test stability issues raised by the attached analysis provided by @Sancus.

[browserstack-flakiness-report.html](https://github.com/user-attachments/files/31810217/browserstack-flakiness-report.html)

rwood-moz:
I originally had a huge PR to tackle this (#1281) but ultimately decided it's too large (and I'm not 💯 confident in that mega PR), so will tackle this in much smaller steps instead.

rwood-moz:
`P2 Make dependencies deterministic` was taken care of in #1314.

rwood-moz:
`P0 Retain decisive evidence` is being taken care of in #1325 with better reporting including logging a BrowserStack session link so we can easily navigate from the GHA E2E test job logs to the BrowserStack session to see the corresponding videos etc.

rwood-moz:
Related to #1328.

rwood-moz:
Related to #1331 and #1332.

rwood-moz:
`P1 	Restore a precise popup signal` recommended changes being implemented in #1337.

rwood-moz:
`P1 	Stabilize sign-in and signup` is being addressed in #1338.

PR #1325, opened 2026-09-16
Improve E2E test reporting

rwood-moz:
## What Changed

- Added a lightweight Playwright reporter that writes a pass/fail/skip test table to the GitHub Actions run Summary
- Preserved the existing Playwright output in the expandable E2E workflow step, including console logs, progress, and failure details.
- Kept the detailed HTML report available for local test runs
- Updated the E2E documentation to describe the new reporting and debugging workflow.

## Why

The lightweight Summary provides a quick overview of all test results without archiving detailed Playwright reports that may contain sensitive test-step values.

## Limitations and Notes

- The GitHub Actions Summary contains only test names and passed, failed, or skipped statuses.
- Existing Playwright logging inside the E2E step remains unchanged.
- The Plawywright HTML report is only archived when the tests run against the local stack (validate GHA) but not archived when the E2E tests run in CI against stage or prod because we don't want to expose secrets.
- Test titles are included in the Summary, so secrets should never be used in test titles.
- Tests that pass after a retry are reported as passed.

I worked with ChatGPT's Codex to make this PR.

## Applicable Issues
Related to #1280.

## QA Log
- You'll see the new E2E Test Results report on the Validate GHA that was run as part of [this actual PR's](https://github.com/thunderbird/thunderbird-accounts/actions/runs/35359450736) checks
- I manually triggered the nightly E2E tests on desktop Firefox GHA job (against this branch) and you can see the new [E2E Test Results summary](https://github.com/thunderbird/thunderbird-accounts/actions/runs/35359775951), which also includes a link to the BrowserStack session.


davinotdavid (commented):
This looks so beautiful! Really like the output that it generates at the moment, nice! 🎉 

However, [on your second example](https://github.com/thunderbird/thunderbird-accounts/actions/runs/35359775951), there's a failure that doesn't seem to fail the CI. Is that intentional?

Image

Shouldn't the failure / exit code 1 be a failure in the job itself?

rwood-moz:
> Shouldn't the failure / exit code 1 be a failure in the job itself?

Thanks for the review and that's a great question! For the nightly jobs we don't fail the GHA workflow/CI job itself, we set `continue-on-error: true`:

https://github.com/thunderbird/thunderbird-accounts/blob/main/.github/workflows/e2e-desktop-nightly-chrome.yml#L53

Only because IMO the GHA workflow itself did it's job fine, but yes there might be E2E test failures. In regular CI where we run tests on PRs etc we fail the job because we want to flag it in the PR. For the nightly GHA jobs I check the results each day in BrowesrStack regardless. Let me think on that, if we should fail the nightly GHA job itself if there are test failures, maybe we should...

rwood-moz:
> > Shouldn't the failure / exit code 1 be a failure in the job itself?

Ok I thought more about it haha yes I don't see the harm in failing the actual nightly GHA workflows themselves if there are E2E test failures, might help ensure any new issues are caught faster. I'm going to update this PR to take the `continue-on-error: true` out of all of the nightly E2E test workflows. Thanks for that suggestion! Note: this will cause some emails to be sent when the GHA fails, but I'm not sure who receives them... we can always put back the `continue-on-error: true` if they're a hassle.

davinotdavid (approved):
LGTM, thanks!

I think being noisy can be good since E2E tests shouldn't be broken by folks (I am guilty of doing this often heh)

